PR #2, opened 2023-07-28
main dependencies installed required for future development

DYA13:


.DS_Store
akosasante, line 1: You may want to remove this following the steps you've used in the past for ensuring that `.DS_Store` files are handled by gitignore (https://codethedream.slack.com/archives/D04TDKZGYNT/p1679608645838939)

src/app.js
```diff
@@ -1,20 +1,64 @@
...
 const cors = require('cors')
-const favicon = require('express-favicon');
-const logger = require('morgan');
+const favicon = require('express-favicon')
+const mainRouter = require('./routes/mainRouter.js')
+const helmet = require('helmet')
+const session = require('express-session')
+const passport = require('passport')
+const mongoose = require('mongoose')
+const asyncErrors = require('express-async-errors')
```
akosasante: The docs for this library indicate that only need to require it, in order for it to work (https://www.npmjs.com/package/express-async-errors)

You can see in the Github code for that library, that it doesn't export anything: https://github.com/davidbanham/express-async-errors/blob/master/index.js

If you were to do a `console.log(asyncErrors)` you'll see that it just is logging an empty object `{}`.

So all we need on this line is:
```suggestion
require('express-async-errors')
```

and we should remove line 42-43.

This is similar to this exercise in the Node course, where we call `require('./07-mind-grenade')` to run the code in that file without saving anything to a variable: https://github.com/Code-the-Dream-School/node-express-course/blob/main/01-node-tutorial/03-modules.js

src/app.js
```diff
@@ -1,20 +1,64 @@
...
 
-const mainRouter = require('./routes/mainRouter.js');
-
-// middleware
-app.use(cors());
-app.use(express.json());
-app.use(express.urlencoded({ extended: false }));
-app.use(logger('dev'));
+// middleware setup
+app.use(cors())
```
akosasante: For now, leaving this empty will work fine with the frontend since it's not trying to pass any cookies in the request yet. But once they do add cookies (ie: making calls with `withCredentials: true` set in their axios configuration), this CORS middleware will block it because we're not currently telling the middleware which websites it should allow cookies from.

It would be good to have that set up ahead of time at least for local development:

```suggestion
app.use(cors({ origin: ["http://localhost:3000", credentials: true }));
```

or with regex:
```suggestion
app.use(cors({ origin: [/localhost:3000$/], credentials: true }));
```

Later once we deploy these repos to Render.com, then we'll also need to add the deployed URL but we can do that when the time comes.

src/app.js
```diff
@@ -1,20 +1,64 @@
...
+app.use(xss())
+app.use(limiter)
+app.use(mongoSanitize())
+app.use(express.json())
+app.use(helmet())
+app.use(express.urlencoded({ extended: false }))
+app.use(morgan('dev'))
 app.use(express.static('public'))
-app.use(favicon(__dirname + '/public/favicon.ico'));
+app.use(favicon(__dirname + '/public/favicon.ico'))
```
akosasante, line 42: Currently, `app.js` lives inside the `src` folder. That means when `app.js` gets run, `__dirname` will evaluate to `<path_to_repo>/src`. So this will look for a folder called `public` _inside_ the `src` folder. 

Currently, you have your `public` folder outside in the root of the repository. I would advise you move that into the `src` folder, or make sure that this line is pointing to the correct path if you choose not to move it. 

Eg: `app.use(favicon(path.resolve(__dirname, '../public/favicon.ico')));` (requires `const path = require("path");` at top of file to use the [path module](https://nodejs.org/api/path.html#pathresolvepaths))

src/app.js
```diff
@@ -1,20 +1,64 @@
...
+app.use(bodyParser.json())
+app.use(passport.initialize())
+app.use(passport.session())
+app.use(cookieParser())
+app.use(
+  session({ secret: 'your-secret-key', resave: false, saveUninitialized: true })
```
akosasante: We should go ahead and set this in the `.env` file and make sure all team members are using the same value

src/app.js
```diff
@@ -1,20 +1,64 @@
...
+)
```
akosasante: Based on the docs, these two Passport lines should be defined _after_ you define the Express Session middleware, since Passport will be adding data to that: https://www.npmjs.com/package/passport#:~:text=%7D)%3B-,Middleware,-To%20use%20Passport

```suggestion
app.use(cookieParser())
app.use(
  session({ secret: 'your-secret-key', resave: false, saveUninitialized: true })
)
app.use(passport.initialize())
app.use(passport.session())
```

src/app.js
```diff
@@ -1,20 +1,64 @@
...
+const helmet = require('helmet')
+const session = require('express-session')
+const passport = require('passport')
+const mongoose = require('mongoose')
+const asyncErrors = require('express-async-errors')
+const cookieParser = require('cookie-parser')
+const rateLimit = require('express-rate-limit')
+const xss = require('xss')
+const mongoSanitize = require('mongo-sanitize')
+const limiter = rateLimit({
```
akosasante, line 17: Currently, this isn't being used anywhere. Based on the docs you need to call `app.use(limiter)` at some point in order for the rate-limiter library to use these values: https://www.npmjs.com/package/express-rate-limit

package.json
```diff
@@ -9,11 +9,21 @@
...
-    "mongoose": "^7.0.1",
-    "morgan": "^1.10.0"
+    "express-rate-limit": "^6.8.1",
+    "express-session": "^1.17.3",
+    "helmet": "^7.0.0",
+    "mongo-sanitize": "^1.1.0",
+    "mongoose": "^7.4.1",
+    "morgan": "^1.10.0",
+    "passport": "^0.6.0",
+    "xss": "^1.0.14"
```
akosasante: This library seems to be meant to be used more manually, rather than as an Express middleware: https://www.npmjs.com/package/xss

I would suggest instead using https://www.npmjs.com/package/express-xss-sanitizer

package.json
```diff
@@ -4,18 +4,42 @@
...
+    "helmet": "^7.0.0",
+    "jsonwebtoken": "^9.0.1",
+    "mongo-sanitize": "^1.1.0",
+    "mongoose": "^7.4.1",
+    "morgan": "^1.10.0",
+    "passport": "^0.6.0",
+    "validator": "^13.9.0",
+    "winston": "^3.10.0",
+    "xss": "^1.0.14",
+    "xss-clean": "^0.1.4"
```
akosasante: This package seems to have been retired: https://www.npmjs.com/package/xss-clean

I would recommend this one instead to prevent cross-site-scripting using Express middleware: https://www.npmjs.com/package/express-xss-sanitizer

akosasante (commented):
Review notes for just the first [commit](https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/2/commits/449d909ab973bc4216bc489ecdd89aa4ee43de20) (main dependencies installed):

src/app.js
```diff
@@ -1,80 +1,86 @@
...
+const asyncErrors = require('express-async-errors')
+const cookieParser = require('cookie-parser')
+const rateLimit = require('express-rate-limit')
 const limiter = rateLimit({
   windowMs: 15 * 60 * 1000, // 15 minutes
-  max: 60, // limit each IP to 100 requests per windowMs
-});
+  max: 60 // limit each IP to 100 requests per windowMs
+})
+const logger = require('./logger')
```
akosasante: Where is this file or folder? I don't see it checked into this PR

akosasante: For a simple Winston logger that prints to the console/stdout, you could have a file like `src/logger/index.js`:

```javascript
const { createLogger, format, transports } = require('winston');

const logger = createLogger({
  level: 'info',
  format: format.combine(
    // Sets up the padding and coloring of the logs for the console
    format.cli(),
    // Allows us to log errors directly, eg: logger.error(new Error('Error Message')) will log the stack trace for that error
    format.errors({ stack: true }),
    // generates a timestamp that we can add to each log
    format.timestamp({
      format: 'YYYY-MM-DD HH:mm:ss'
    }),
    format.printf(({ level, message, timestamp, stack }) => { 
      // Formatting the output of the logging function.
      // Stack may be undefined if there was no error passed to this log.
      return stack ? `[${timestamp} ${level}]: ${message} - ${stack}` : `[${timestamp}] ${level}: ${message}`;
    })
  ),
  transports: [
    new transports.Console({'timestamp':true}),
    // If we wanted to write logs to a file, we could uncomment below:
    // new transports.File({ filename: 'quick-start-combined.log' })
  ]
});

module.exports = logger;
```

src/app.js
```diff
@@ -1,80 +1,86 @@
...
   try {
-    await connectDB(process.env.MONGO_URL);
-    app.listen(port, console.log(`Server is listening on port ${port}`));
+    await connectDB(process.env.MONGO_URL)
+    app.listen(port, () => {
+      logger.info(`Server running on http://localhost:${port}`)
+    })
   } catch (error) {
-    console.log(error);
+    console.log(error)
```
akosasante:
```suggestion
    logger.error(error)
```

src/app.js
```diff
@@ -1,80 +1,86 @@
...
-});
+  max: 60 // limit each IP to 100 requests per windowMs
+})
+const logger = require('./logger')
 
 // middleware setup
 // Configure express-session middleware
 app.use(
   session({
     secret: 'your-secret-key', // Replace with a secret key for session data encryption
```
akosasante: We should go ahead and set this up in the .env file

akosasante (commented):
Review comments for logging+security [commit](https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/2/commits/f3713f10a48996135640065a577bf446aacd4ea1)

akosasante (commented):
Changes in error handling + middleware [commit](https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/2/commits/48e11bb413c8ea3910860d7d8fc901c3d8236f29) all look good to me ✅

db/connect.js
akosasante, line 1: This file exists but is not being used 😕 

I would have expected that in `app.js` connecting to the db would call this `connectDB` method exported by this file.

.prettierrc
akosasante, line 1: When I ran `npm run prettier` I got the following console warning:

```
[warn] Ignored unknown option { "prettier/prettier": ["error", { singleQuote: true, parser: "flow" }] }.
```

We should be using the format described in the Prettier docs: https://prettier.io/docs/en/configuration.html

If we want to just stick to the defaults, we can even just make this an empty config object: `{}`

Otherwise, if you have strong opinions on things like single-quotes vs double-quotes, you can see all the available options here in this JSON schema: http://json.schemastore.org/prettierrc

akosasante: If you're interested, you can also install and setup `husky` : [https://prettier.io/docs/en/install.html#git-hooks](https://prettier.io/docs/en/install.html#git-hooks)

The frontend team is doing this here: https://github.com/Code-the-Dream-School/dd-prac-team4-front/pull/1#discussion_r1278290160

akosasante (commented):
Review for formatting + linting [commits](https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/2/files/48e11bb413c8ea3910860d7d8fc901c3d8236f29..627ddb2bdd67bc9adcc66ee63d984a588024f784)

.DS_Store
DYA13: I have  added `.DS_Store` in gitignore. At the moment it seems there are no problems with cache of DS Store

src/app.js
DYA13: Corrected commit- 46b69103aad60abd5f75962aba94b9fce32e31a3

src/app.js
DYA13: it is done commit-f51ae19a3da8ebd320d5f74026470b7877262ac0

src/app.js
DYA13: it is corrected , commit- 53547538447ea06d0d06645414122fc910ef604e

src/app.js
DYA13: should I delete these dependencies from lines 15-18 ?

package.json
DYA13: should I remove "xss" and install "express-xss-sanitizer"?

package.json
DYA13: xss package removed and  express-xss-sanitizer installed  commit -305bc2f6fa189d62cb587dd271f4b31d1ae4b1b6

package.json
DYA13: xss removed

src/app.js
DYA13: for winston I have created folder logs with file logger.js 
`// logger.js
const winston = require("winston");

// Define the log levels and colors (optional)
const logLevels = {
  error: "red",
  warn: "yellow",
  info: "green",
  debug: "blue",
};

winston.addColors(logLevels);

const logger = winston.createLogger({
  level: "debug",
  levels: logLevels,
  format: winston.format.combine(
    winston.format.colorize(),
    winston.format.timestamp({ format: "YYYY-MM-DD HH:mm:ss" }),
    winston.format.printf(({ timestamp, level, message }) => {
      return `${timestamp} [${level.toUpperCase()}]: ${message}`;
    }),
  ),
  transports: [
    new winston.transports.Console(),
    new winston.transports.File({ filename: "logs/app.log" }),
  ],
});

module.exports = logger;
`

DYA13: Should I change for your code ?

src/app.js
DYA13: done 0b7b09c3af94cc4e4cd2867ea4c682053153f5c4

db/connect.js
DYA13: In this case I will delete this file as far as we want to make mongoDB setup in app.js , is it correct?

.prettierrc
DYA13: .prettierrc changed according to https://prettier.io/docs/en/configuration.html

src/app.js
```diff
@@ -1,20 +1,87 @@
...
-app.use(cors());
 app.use(express.json());
-app.use(express.urlencoded({ extended: false }));
-app.use(logger('dev'));
-app.use(express.static('public'))
-app.use(favicon(__dirname + '/public/favicon.ico'));
+//Security middleware
+app.use(helmet());
+
+app.use(cors({ origin: [/localhost:8000$/], credentials: true }));
```
akosasante: Is `8000` the port that the frontend will run on? We want to tell our backend, which urls from the front to allow through and not block with CORS

src/app.js
akosasante: I don't think you should delete, you just need to add `app.use(limiter)` somewhere below this in the file so that the limiting will actually be applied for requests going through the Express server 🙂

package.json
akosasante: Looks like removed `xss` package :white\_check\_mark: 

But I'm also still seeing the retired `xss-clean` here, can we remove that one as well please ?

src/app.js
akosasante: Ah what you have seems fine as well; I am not seeing it in this PR however?

db/connect.js
akosasante: It's up to you all where you want to keep it. It's pretty small amount of code, so I think it's fine to keep it in-line in `app.js`. In that case yes delete the file and folder. Otherwise, if you want to keep it separate, then make sure the code in `app.js` is correctly connecting using the file the db folder.

src/app.js
DYA13: added  commit 78f5c188050a31f6e0827daf03e12d01c88e4e74
Thank you very much!

.prettierrc
akosasante: Currently if I try to run `npm run format` I get warnings:

`[warn] Ignored unknown option { "prettier/prettier": ["error", { trailingComma: "es5", tabWidth: 4, semi: false, singleQuote: true }] }.`

Looking at the link you sent, I don't believe we need this part:

```
"prettier/prettier": [
    "error",

  ]
```

I think that's more the format that `ESLint` uses for their configuration or something.

Rather you _just_ want a JSON object with the rules that you want to use for formatting with Prettier:

```json
{
    "trailingComma": "es5",
    "tabWidth": 4,
    "semi": false,
    "singleQuote": true
}

```

package.json
```diff
@@ -4,18 +4,42 @@
...
+    "passport": "^0.6.0",
+    "validator": "^13.9.0",
+    "winston": "^3.10.0",
+    "xss-clean": "^0.1.4"
   },
   "devDependencies": {
-    "nodemon": "^2.0.21"
+    "eslint": "^8.46.0",
+    "eslint-config-prettier": "^8.9.0",
+    "eslint-plugin-prettier": "^5.0.0",
```
akosasante: I don't believe you need `eslint-plugin-prettier`. Just `eslint-config-prettier` should be enough to have ESLint and Prettier work well together

src/app.js
akosasante: @DYA13  Not sure if you saw this comment since I think I reviewed it on an older commit

package.json
DYA13: The package removed commit 3bd2bed91c76f18184818a3142ab0a35eb927b59

src/app.js
akosasante: @DYA13  Not sure if you saw this comment since I think I reviewed it on an older commit

src/app.js
DYA13: It is strange, I have this folder in root directory in my branch. Should I try to change the location of the folder logs possibly ?

src/app.js
DYA13: the  frontend repo runs on 3000 port as mentioned in README. Thus should be changed for 3000?

package.json
DYA13: eslint-plugin-prettier removed , commit bb4a51e79cbcd88ffdaa4e080689113269222f95

src/app.js
akosasante: Yes please change it to 3000 :+1:

src/app.js
akosasante: You try doing `git add ./logger` and see if it shows up in "Changes staged for commit" when you do `git status` ? 

I'm wondering if this `.gitignore` file is causing it to be ignored, though that would surprise me since it doesn't match up to what you said you've called it.

akosasante:
@DYA13 latest commits look great! There's still a few things left to resolve (just listing them here since in a longer PR like this comments can get difficult to find fter a while lol):

*   change cors middleware whitelisted localhost port from `8000` to `3000` (to allow requests originating from the frontend running locally)
*   update  prettierrc.json to correct format
*   decide whether to keep the db connection code in-line or in a separate folder
*   set the session secret in the `.env` file rather than hardcode a placeholder
*   don't store the result of `express-async-errors` in a variable
*   get `logger` folder/file pushed to git

db/connect.js
DYA13: Hello Akos, I have corrected this code, mongoDB setup in app.js commit - c5221b54f74c5f66e1aee801da9d83d900bd96aa

src/app.js
DYA13: Sorry Akos, I didn't see it , now I have changed commit- d777f3cb45232dcd987b186f5c6544f67714967f

src/app.js
DYA13: yes this one I have done in older commit

src/app.js
DYA13: Akos you are completely right , this folder logs with logger.js  was ignored by .gitignore. Now I pushed it to my branch. Could you please check if you see it now ? commit - cdbceb8eb06c70fb5ecd5f5e18cc2d25cf496bf8

.prettierrc
DYA13: Done commit-497b002af376ea48b56bd00e3bd5de9441751573

src/app.js
DYA13: Changed to 3000 , commit 4682f9f48cd7941d2c850db797dcede94158b5de

src/app.js
```diff
@@ -60,11 +60,14 @@ app.use(passport.initialize());
 app.use(passport.session());
 
 // Database setup (using Mongoose)
+mongoose.set("strictQuery", true);
 
 const connectDB = (url) => {
   return mongoose.connect(url);
 };
 
+module.exports = connectDB;
```
akosasante: Why do we need to  export this? Will other files/modules be needing this `connectDB` function? I think we can just remove it for now if it's not being used yet.

src/app.js
akosasante: Yes, I can see the file now :tada: 

This line should be updated though to match the path the file is at otherwise Node won't be able to find/load it correctly:

```suggestion
const logger = require('./logs/logger')
```

.eslintrc.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": ["eslint:recommended", "plugin:prettier/recommended"],
```
akosasante, line 2: Since we're using the `eslint-config-prettier` rather than the `eslint-plugin-prettier` we need to update the `extends` here:

```suggestion
  "extends": ["eslint:recommended", "prettier"],
```

src/app.js
```diff
@@ -1,20 +1,89 @@
...
+app.use(passport.session())
+
+// Database setup (using Mongoose)
+mongoose.set('strictQuery', true)
+
+const connectDB = (url) => {
+    return mongoose.connect(url)
+}
+
+module.exports = connectDB
```
akosasante: Why do we need to export this? Will other files/modules be needing this `connectDB` function? I think we can just remove it for now if it's not being used yet.

src/app.js
```diff
@@ -1,20 +1,89 @@
...
+
+app.use('/api/v1/auth', authRouter)
+
+// Error handling middleware (must be defined after all other routes and middleware)
+//add later
 
-// routes
-app.use('/api/v1', mainRouter);
+// Start the server
+const port = process.env.PORT || 3000
```
akosasante, line 74: I'm sorry I think I miscommunicated here. We can keep the **backend** server's port as `8000`, but the **frontend** repo will run locally on `3000`.

Because of that we need to make our CORS middleware, allow requests _from_ port 3000, so we need to update line 33 to allow that:

```
app.use(cors({ origin: [/localhost:3000$/], credentials: true }))
```

and revert this line back to how it was before:

```suggestion
const port = process.env.PORT || 8000
```

.eslintrc.json
DYA13: done

src/app.js
DYA13: removed

src/app.js
DYA13: done

DYA13:
> @DYA13 latest commits look great! There's still a few things left to resolve (just listing them here since in a longer PR like this comments can get difficult to find fter a while lol):
> 
> * change cors middleware whitelisted localhost port from `8000` to `3000` (to allow requests originating from the frontend running locally)
> * update  prettierrc.json to correct format
> * decide whether to keep the db connection code in-line or in a separate folder
> * set the session secret in the `.env` file rather than hardcode a placeholder
> * don't store the result of `express-async-errors` in a variable
> * get `logger` folder/file pushed to git

Akos ,  I have edited all these points. Could you please check?

src/app.js
```diff
@@ -1,20 +1,89 @@
...
+mongoose.set('strictQuery', true)
+const cookieParser = require('cookie-parser')
+const rateLimit = require('express-rate-limit')
+const limiter = rateLimit({
+    windowMs: 15 * 60 * 1000, // 15 minutes
+    max: 60, // limit each IP to 100 requests per windowMs
+})
 
-const mainRouter = require('./routes/mainRouter.js');
+const logger = require('./logs/logger')
```
akosasante: @DYA13 , hmm, I just realized that your `logs` folder is outside of the `src` folder where the `app.js` lives. So you'll either need to update this to: `../logs/logger` (the two dots mean "the folder above/outside of this one); or move the `logs` folder to be inside the `src` folder together with `app.js`

logs/logger.js
```diff
@@ -0,0 +1,30 @@
...
+  error: "red",
+  warn: "yellow",
+  info: "green",
+  debug: "blue",
+};
+
+winston.addColors(logLevels);
+
+const logger = winston.createLogger({
+  level: "debug",
```
akosasante: We can remove this. By having this here, it means we can _only_ use `logger` to log DEBUG level logs. So `logger.info` and `logger.error` won't print anything, only `logger.debug` will. We _could_ do it that way and create separate loggers for the info and error levels but that feels too repetitive for this case.

Rather, we can leave out the `level` in this `createLogger` function; and then that will allow us to choose what level we want to use each time when we log things.

logs/logger.js
```diff
@@ -0,0 +1,30 @@
...
+  levels: logLevels,
+  format: winston.format.combine(
+    winston.format.colorize(),
+    winston.format.timestamp({ format: "YYYY-MM-DD HH:mm:ss" }),
+    winston.format.printf(({ timestamp, level, message }) => {
+      return `${timestamp} [${level.toUpperCase()}]: ${message}`;
```
akosasante: The `level.uppercase` is kinda showing up with weird blanks, because `colorize` is changing the value of `level` to include some extra text that tells the console how to color the word: winstonjs/winston#1345

what we can do if you want this to be capitalized is to set up the capitalization to happen earlier:

```javascript
const capitalizeLevel = winston.format((info) => {
  info.level = info.level.toUpperCase();
  return info;
});

winston.addColors(logLevels);

const logger = winston.createLogger({
  // level: "debug",
  levels: logLevels,
  format: winston.format.combine(
    capitalizeLevel(),
    winston.format.colorize(),
    winston.format.timestamp({ format: "YYYY-MM-DD HH:mm:ss" }),
    winston.format.printf(({ timestamp, level, message }) => {
      return `${timestamp} [${level}]: ${message}`;
    }),
  ),
  transports: [
    new winston.transports.Console(),
    new winston.transports.File({ filename: "logs/app.log" }),
  ],
});
```

src/app.js
```diff
@@ -1,20 +1,89 @@
...
 app.use(express.static('public'))
-app.use(favicon(__dirname + '/public/favicon.ico'));
+app.use(favicon(__dirname + '/public/favicon.ico'))
+app.use(bodyParser.json())
+
+app.use(cookieParser())
+// Configure express-session middleware
+app.use(
+    session({
+        secret: 'your-secret-key', // Replace with a secret key for session data encryption
```
akosasante: This is still being hardcoded rather than set in `.env` and then this can be `process.env.SESSION_SECRET` or something like that

src/app.js
akosasante: I still see it here; have you pushed that commit yet?

src/app.js
```diff
@@ -1,20 +1,89 @@
...
+app.use(limiter)
+// middleware setup
+
+app.use(xss())
+
+app.use(express.json())
+//Security middleware
+app.use(helmet())
+
+app.use(cors({ origin: [/localhost:8000$/], credentials: true }))
```
akosasante, line 33: Still looking for this to be updated to `3000` please :slightly\_smiling\_face:

akosasante (approved):
I'm giving the :white\_check\_mark: :smile: 

There's a few small things to clean up but we can follow up in a separate PR :+1: 

*   localhost:3000 for CORS, 
*   session secret in env, 
*   remove the \`module.exports = connectDB\`

PR #3, opened 2023-07-29
[PS-26] created User model in the backend with validation

NatachaKey:


akosasante (commented):
I know this one was already merged into the feature branch; but all looks good to me 👍 ✅

PR #5, opened 2023-07-29
[PS-28]  [PS-30] Created a /get_user endpoint, works in postman

NatachaKey:
Update user to be done

### DIFF
--- a/src/controllers/userController.js
+++ b/src/controllers/userController.js
@@ -9,12 +9,21 @@ const createJWT = ({ payload }) => {
   return token;
 };
 
+//returning a regular javascript object that is a subset of the fields of the User model.
+const formatUserForFrontend = (user) => {
+  return { name: user.name, userId: user._id, role: user.role };
+};
+
+
 const getSingleUser = async (req, res) => {
-  const user = await User.findOne({ _id: req.params.id }).select('-password');
+  const user = await User.findOne({ _id: req.params.id });
   if (!user) {
     res.status(400).json({ msg: `No user with id : ${req.params.id}` });
   }
-  res.status(200).json({ user });
+  // If the user is found, it calls the formatUserForFrontend function to format the user data before sending it in the response
+  const tokenUser = formatUserForFrontend(user);
+
+  res.status(200).json({ user : tokenUser });
 };
 
 const updateUser = async (req, res) => {
@@ -33,12 +42,7 @@ const updateUser = async (req, res) => {
   user.name = name || user.name
 
   user = await user.save();
-  
-  //returning a regular javascript object that is a subset of the fields of the User model.
-  const formatUserForFrontend = (user) => {
-    return { name: user.name, userId: user._id, role: user.role };
-  };
-
+  //After updating the user, it calls the formatUserForFrontend function to format the updated user data before sending it in the response
   const tokenUser = formatUserForFrontend(user);
   res.status(200).json({ user: tokenUser });
 };
--- a/src/controllers/userController.js
+++ b/src/controllers/userController.js
@@ -34,12 +34,12 @@ const updateUser = async (req, res) => {
 
   user = await user.save();
   
-  //create tokenUser with data updated
-  const createJWT = (user) => {
+  //returning a regular javascript object that is a subset of the fields of the User model.
+  const formatUserForFrontend = (user) => {
     return { name: user.name, userId: user._id, role: user.role };
   };
 
-  const tokenUser = createJWT(user);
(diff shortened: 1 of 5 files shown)


src/models/User.js
```diff
@@ -57,5 +58,13 @@ const UserSchema = new mongoose.Schema({
   creditCardInfo: creditCardSchema,
 });
 
+//before we save the document we hash password; this points to the user (this= user)
+UserSchema.pre('save', async function () {
+  console.log(this.modifiedPaths()); //to see which properties were modified
+   console.log(this.isModified('name')); // in console we get false if we are not modifying name, and true if we do
```
akosasante, line 64: Do we need these logs here or was it just for debugging?

src/models/User.js
```diff
@@ -57,5 +58,13 @@ const UserSchema = new mongoose.Schema({
...
+  
+  if (!this.isModified('password')) return; //if the password is not modified then stop
+  });
```
akosasante, line 67: Is there meant to be an else branch here of what to do if password was modified (or does that change live in another branch)?

src/controllers/userController.js
```diff
@@ -0,0 +1,48 @@
...
+// create token (we use it here after updating user's data )
+const createJWT = ({ payload }) => {
+  const token = jwt.sign(payload, process.env.JWT_SECRET, {
+    expiresIn: process.env.JWT_LIFETIME,
+  });
+  return token;
+};
+
+const getSingleUser = async (req, res) => {
+  const user = await User.findOne({ _id: req.params.id }).select('-password');
```
akosasante: are there any other fields that we don't need to return the frontend? I'm thinking maybe `role` and `creditCardInfo` don't need to be returned in most cases? If you end up implementing a `formatUserForFrontend` function ([PR 1](https://github.com/Code-the-Dream-School/dd-prac-team4-back/commit/b48b2d231e4183b4ca3b283bc055a0eafb657021#r123402026) / [PR2](https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/6#discussion_r1282047069)), maybe you can just reuse that here?

src/controllers/userController.js
```diff
@@ -0,0 +1,48 @@
...
+// Update name if request included a non-null value, otherwise keep the name as-is
+user.name = name || user.name
+
+await user.save(); // using pre save hook (see user model) to save updated user
+  //create tokenUser with data updated
+  const createJWT = (user) => {
+    return { name: user.name, userId: user._id, role: user.role };
+  };
+
+  const tokenUser = createJWT(user);
```
akosasante: Same comment as: b48b2d2, this is not really a JWT, it's just a regular json object.

src/controllers/userController.js
```diff
@@ -0,0 +1,48 @@
...
+// Update email if request included a non-null value, otherwise keep the email as-is
+user.email = email || user.email
+
+// Update name if request included a non-null value, otherwise keep the name as-is
+user.name = name || user.name
+
+await user.save(); // using pre save hook (see user model) to save updated user
+  //create tokenUser with data updated
+  const createJWT = (user) => {
+    return { name: user.name, userId: user._id, role: user.role };
```
akosasante: Have you double checked, if this is actually returning the updated version of the user? I'm not sure if `.save()` will edit the value inside the `user` variable after it updates the database. We may need to actually grab the returned value. 

Something like:  
 

```javascript
  let user = await User.findOne({ _id: req.params.id }); //get user
  
  // Update email if request included a non-null value, otherwise keep the email as-is
  user.email = email || user.email

  // Update name if request included a non-null value, otherwise keep the name as-is
  user.name = name || user.name

  user = await user.save();
```

src/models/User.js
NatachaKey: It's for debugging only, should I delete them? as well as 
comments?

src/models/User.js
NatachaKey: we will add it later, i just was recycling code from the previous project🫣

src/controllers/userController.js
NatachaKey: @akosasante yes,  it updated the data correctly in both cases- -the hashed password and id and role stood the same - i will commit your code to the project

src/controllers/userController.js
NatachaKey: done!

src/controllers/userController.js
NatachaKey: @akosasante done

PR #6, opened 2023-07-31
Ps4 login endpoint

mvstoyan:


src/controllers/authController.js
```diff
@@ -45,6 +50,30 @@ const register = async (req, res) => {
...
+  }
+
+  // Verify the entered password
+  const isPasswordValid = await argon2.verify(user.password, password);
+  if (!isPasswordValid) {
+    return res.status(401).json({ msg: 'Invalid password' });
+  }
+
+  // Create and send a JWT token
+  const tokenUser = createJWT(user);
```
akosasante: I left some similar comments on Natalia's register endpoint, which you can refer to here: https://github.com/Code-the-Dream-School/dd-prac-team4-back/commit/b48b2d231e4183b4ca3b283bc055a0eafb657021#r123402117 

I believe it would be best to use a cookie-based authentication scheme for this project. We can still generate and store a JWT encrypted token in the cookie. But that way the requests are more secure and the frontend doesn't need to manually store the information from the cookie itself.

We can have a function to create and attach the cookie to the response, that way both the login and the register endpoints can use this function. It can look something like:

```js
// these functions can be kept in this file or if you prefer, moved to a separate file and imported/required here:

// function to attach a signed cookie to the response, that expires after 1 day (user will have to log in again), and contains information about the user as a JSON Web Token
const attachCookiesToResponse = (res, user) => {
  const token = createJWT(user);
  // calculate 1 day in milliseconds = 1000 milliseconds/s * 60 seconds/minute * 60 minutes/hour * 24 hours/day
  const oneDay = 1000 * 60 * 60 * 24;
  res.cookie('token', token, {
    httpOnly: true,
    expires: new Date(Date.now() + oneDay),
    secure: true,
    signed: true,
    sameSite: 'none',
  });
};

const formatUserForFrontend = (user) => ({ name: user.name, userId: user._id, role: user.role });

const login = async (req, res) => {
  ...all the existing code you had, then once you've checked if the user credentials were valid:

  attachCookiesToResponse(res, user);
  res.status(200).json({ user: formatUserForFrontend(user) });
```

akosasante: Please coordinaate with @NatachaKey  when making these changes since it will overlap with her code on the register endpoint :slightly\_smiling\_face:

mvstoyan: Thank you. I made cookies in the next ticket and committed. But I can’t send it to the PR, because I’m stuck on showCurrentUser. You prefer that I do it in this ticket. Or should I delete it there and do it here? 
cookie

akosasante: That's fine, I'll approve this PR so you can merge, and you can add the cookie stuff in a separate PR or your next ticket :+1:

akosasante (approved):

PR #7, opened 2023-08-02
Ps27 show current user

mvstoyan:
Сurrently there are cookies and errors, not finished showCurrentUser. I would like to refactor, and will continue to work on this task, so as not to do double work for everyone later


src/controllers/userController.js
```diff
@@ -3,6 +3,13 @@ const { StatusCodes } = require('http-status-codes');
 const CustomError = require('../errors');
 const { createTokenUser, attachCookiesToResponse, checkPermissions } = require('../utils');
 
+const getAllUsers = async(req,res) => { // Function to get all users
+    console.log(req.user);
```
akosasante: Can be removed before merging

src/controllers/userController.js
```diff
@@ -13,6 +20,11 @@ const getSingleUser = async(req,res) => { // Function to get a single user by ID
     res.status(StatusCodes.OK).json({ user }); // Send a JSON response with the status code 200 OK and the user
 };
 
+const showCurrentUser = async(req,res) => { // Function to get the current user's information
+    // Send a JSON response with the status code 200 OK and the current user's information
+    res.status(StatusCodes.OK).json({ user: req.user });
```
akosasante: One thing that will be a bit odd here is that this will return a tokenized user (with `userId,` `role`, `email`) but not much other info about the current user. If the frontend wants to use this endpoint for something like showing the user profile page, they won't have the user's name or credit card info for example. Instead, lets ensure that all endpoints in the `userController` return a full user (minus the password of course). 

For the `authController` it's fine to return the tokenized user since we shouldn't expect the frontend to use the return value of those endpoints for the same type of purpose. 

Here, this would probably mean that you need to add a `User.findOne` call to this handler

src/controllers/userController.js
```diff
@@ -30,7 +42,37 @@ const updateUser = async(req,res) => {
...
+    if (!oldPassword || !newPassword) { // Check if oldPassword and newPassword are provided
+        throw new CustomError.BadRequestError('Please provide both values');
+    }
+    const user = await User.findOne({ _id: req.user.userId }); // Find the user in the database based on the current user's ID
+    // Compare the old password provided with the user's current password
+    const isPasswordCorrect = await user.comparePassword(oldPassword);
+    if(!isPasswordCorrect) { // Throw an UnauthenticatedError if the old password is incorrect
+        throw new CustomError.UnauthenticatedError('Invalid Credentials');
+    }
+    user.password = newPassword; // Update the user's password
```
akosasante: Is this saving the user's new password in plain text? I think it's not hashing it since I don't see `argon` being used here or in the User model.

akosasante (commented):
For now, I changed the base branch to your refactor branch, just to make the changes here easier to view and review. Looks good, just a couple small comments, then you should be good to go.

src/controllers/userController.js
mvstoyan: 57-58 - Hash the new password

src/controllers/userController.js
mvstoyan: Thank you very much for the review. I may have misunderstood, so I would like to clarify. Does my error handling and data return in the showCurrentUser function look correct? Are there any other improvements you would recommend?

src/controllers/userController.js
mvstoyan: console.log removed

PR #12, opened 2023-08-07
logout endpoint created

DYA13:


src/app.js
```diff
@@ -51,6 +51,12 @@ app.use(
     saveUninitialized: false,
   })
 );
+
+const crypto = require('crypto');
+
+const secretKey = crypto.randomBytes(32).toString('hex');
+console.log('Secret Key:', secretKey);
+
```
akosasante: Curious what this is being used for?

src/controllers/authController.js
```diff
@@ -88,7 +89,18 @@ const login = async (req, res) => {
   }
 };
 
+//logout endpoint
+
+const logout = async (req, res) => {
+  res.cookie('token', 'logout', {
+    httpOnly: true,
+    expires: new Date(Date.now() + 1000),
```
akosasante: We can set this to just `new Date()` (which would return just the current datetime). We don't need the logged out cookie to last for a specific amount of time.

src/controllers/authController.js
```diff
@@ -88,7 +89,18 @@ const login = async (req, res) => {
   }
 };
 
+//logout endpoint
+
+const logout = async (req, res) => {
+  res.cookie('token', 'logout', {
```
akosasante: This works fine, it's setting the value of the `token` variable in the cookie to the string `"logout"` instead of whatever user data was there. But an even better way would be to just set the `token` to an empty string since that's more clear to anyone using the cookie that it's empty now. This can be done with `res.cookie('token', '', {...})` or even more simply with the [clearCookie](http://expressjs.com/en/5x/api.html#res.clearCookie) function: `res.clearCookie('token', {...})`

src/app.js
DYA13: I used this code to see cookies in console, sorry I have forgotten to delete it

src/controllers/authController.js
DYA13: done `const logout = async (req, res) => {
  res.clearCookie('token', {
    httpOnly: true,
    expires: new Date(Date.now() + 1000),
  });

  res.status(StatusCodes.OK).json({ msg: 'user logged out!' });
};`

src/controllers/authController.js
DYA13: commit  d0963dfab2eba3423eacaf31d2876c72b05df7f2

src/controllers/authController.js
DYA13: commit 7d57f4b4c85fdc0bd16394cb73ab7103c40eeeab

src/routes/authRoutes.js
```diff
@@ -1,9 +1,10 @@
 const express = require('express');
 const router = express.Router();
 
-const { register, login } = require('../controllers/authController');
+const { register, login, logout } = require('../controllers/authController');
 
 router.post('/register', register);
 router.post('/login', login);
+router.get('/logout', logout);
```
akosasante: Per our convo in slack:

As for using `POST` vs `GET` for logout. The latest recommendation is to use `POST` because POST means “changing state in the backend”, vs GET is used to “retrieve state in backend”.. We are changing state because we are updating the cookies and changing the user from being logged into not being logged in. Even though we’re not sending any request body, POST is also better because sometimes browsers will try to eager load `GET` requests or cache them; and we don’t want the browser doing that for the logout functionality. We only want to logout when the user has explicitly requested that

akosasante (approved):
Last suggestion to change endpoint to a `POST`

You can keep the logout handler as-is or implement the if/else logic to return 4xx if the user wasn't logged in. Either way is fine. Approving PR early so feel free to merge after changing the endpoint to POST if you don't wish to make the other change for now.

src/routes/authRoutes.js
DYA13: I have changed the logout router for POST commit d57c739ed60c5d19b29d92430eeab9e723381c8a

PR #13, opened 2023-08-07
Ps48 refactor backend

mvstoyan:
Refactor code. 
Utils folder - jwt.js(cookie), createTokenUser.js, checkPermissions.js.
Error folder - bad-request.js, custom-api.js.


src/app.js
```diff
@@ -51,6 +51,11 @@ app.use(
     saveUninitialized: false,
   })
 );
+// SESSION_SECRET=generate a key for env
+//const crypto = require('crypto');
+//const secretKey = crypto.randomBytes(32).toString('hex');
+//console.log('Secret Key:', secretKey);
```
akosasante: Is this for what each member should use to generate this key for their `.env` file? It looks like it belongs in a documentation file rather than here in the code if it's not being used at runtime.

src/controllers/authController.js
```diff
@@ -1,94 +1,57 @@
...
-      return res.status(404).json({ error: 'User not found' });
-    }
-
-    // Verify the entered password
-    const isPasswordValid = await argon2.verify(user.password, password);
-    if (!isPasswordValid) {
-      return res.status(401).json({ error: 'Invalid password' });
-    }
+  if (!email || !password) {
+    throw new CustomError.BadRequestError('Please provide email and password'); // If both fields are not provided, throw an error
```
akosasante:
```suggestion
    throw new CustomError.BadRequestError('Please provide email and password'); // If either field is not provided, throw an error
```

src/controllers/authController.js
```diff
@@ -1,94 +1,57 @@
...
-};
+const { StatusCodes } = require('http-status-codes');
+const CustomError = require('../errors');
+const { attachCookiesToResponse, createTokenUser } = require('../utils');
 
 const register = async (req, res) => {
-  const { name, email, password } = req.body;
-
-  const emailAlreadyExists = await User.findOne({ email });
+  const { email, name, password } = req.body; // Extract data from the request
```
akosasante, line 8: This code should also check if `email`, `name`, and `password` were passed in/not empty from the request, just like you've done on line35 for the login handler. Otherwise, we may end up returning weird errors if the user only remembers to enter for example only their name but not their email/password

src/controllers/authController.js
```diff
@@ -1,94 +1,57 @@
...
 
-  //first registered user is an admin
-  //get all users, and if there are no users assign role to admin
-  const isFirstAccount = (await User.countDocuments({})) === 0;
-  const role = isFirstAccount ? 'admin' : 'user';
+  const isFirstAccount = (await User.countDocuments({})) === 0; // Check if it's the first account
+  const role = isFirstAccount ? 'admin' : 'user'; // Assign a role based on first account or not
 
-  const hashedPassword = await argon2.hash(password);
+  const hashedPassword = await argon2.hash(password); // Hash the password
```
akosasante: In the User model, we currently have a half-finished pre-save hook, where this password hashing could occur: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/2de5af21b18d62bce63c37223da0e5a92378c8de/src/models/User.js#L62-L69](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/2de5af21b18d62bce63c37223da0e5a92378c8de/src/models/User.js#L62-L69)

If you prefer to instead do it in-line here in the controller, that's totally fine, but in that case would you mind using this PR to delete that pre-save hook in the model since it's not doing anything now apart from logging to console?

src/controllers/userController.js
```diff
@@ -1,59 +1,36 @@
...
-  //After updating the user, it calls the formatUserForFrontend function to format the updated user data before sending it in the response
-  const tokenUser = formatUserForFrontend(user);
-  res.status(200).json({ user: tokenUser });
+// update user with user.save()
+const updateUser = async(req,res) => {
+    const { email, name } = req.body;
+    if (!email || !name) { // Check if email and name are provided
+        throw new CustomError.BadRequestError('Please provide all values');
+    }
+    const user = await User.findOne({ _id: req.user.userId }); // Find the user in the database based on the current user's ID
```
akosasante: This means this endpoint is only ever for updating the _requesting user_ info. Is that the intention? Will we ever want for example the ability for an admin to edit user information? If this is the intention, it may be more clear to add a comment up top or name the function something like `updateCurrentUser`

src/controllers/userController.js
```diff
@@ -1,59 +1,36 @@
...
+     // Update the user's email and name
+    user.email = email;
+    user.name = name;
+    await user.save();  // Save the updated user to the database
+    const tokenUser = createTokenUser(user); // Create a token user and attach the user's cookies to the response
+    attachCookiesToResponse({ res, user:tokenUser });
+    // Send a JSON response with the status code 200 OK and the updated user
+    res.status(StatusCodes.OK).json({ user: tokenUser });
```
akosasante: It's a little odd to return the user token, which has diff fields than what would be returned from the other endpoint this controller. It's probably fine for now, but best practice would be to do the same as what you're doing in the `getSingleUser` endpoint and return the `User` model but without the password field. That way the frontend doesn't have to figure out that sometimes a user is returned with fields like `{email, id, etc}` and other times, from the same `api/v1/users` path, it's returned with different fields like `{userId, role, etc}`

src/middleware/error-handler.js
```diff
@@ -21,6 +22,10 @@ const errorHandlerMiddleware = (err, req, res) => {
     customError.msg = `No item found with id : ${err.value}`;
     customError.statusCode = 404;
   }
+  if (err instanceof UnauthenticatedError) {
+    customError.statusCode = err.statusCode;
+    customError.msg = err.message;
+  }
```
akosasante: This is already handled in the default case on lines 5-8; I don't think this additional if statement is needed.

src/routes/userRoutes.js
```diff
@@ -1,13 +1,17 @@
...
+    authenticateUser
+} = require('../middleware/authentication')
+const { 
+    getSingleUser,
+    updateUser,
 } = require('../controllers/userController');
 
-router.route('/:id').patch(updateUser);
-router.route('/:id').get(getSingleUser);
+router.route('/updateUser').patch(authenticateUser, updateUser);
```
akosasante: Per my other comment, maybe rename the route `updateCurrentUser`

src/middleware/full-auth.js
akosasante, line 1: Is this file being used / what is the difference between this one and `authentication.js` ?

akosasante (commented):
Left some comments. Also don't forget to run `npm run format`\` and `npm run lint:fix`

src/app.js
mvstoyan: Yes, I left that temporarily for other team members. I will remove it from the code since they have already seen it.

src/controllers/authController.js
mvstoyan: Replaced

src/controllers/authController.js
mvstoyan: Снимок экрана 2023-08-09 174821

mvstoyan:
added new folder db-connect.js (connect to the MongoDB database)

src/controllers/authController.js
mvstoyan: I've moved password hashing to User.js and removed it from controllers as discussed. I'll handle any conflicts in PS27 in favor of User.js.
Снимок экрана 2023-08-09 190413

src/routes/userRoutes.js
mvstoyan: Renaming the function to updateCurrentUser

src/middleware/error-handler.js
mvstoyan: Fixed

src/controllers/userController.js
mvstoyan: Fixed

src/middleware/full-auth.js
mvstoyan: The main difference between `full-auth.js` and `authentication.js` is in how the token is extracted from the request. I've decided to stick with `authentication.js` for now as it provides this additional layer of security and automatically includes the token in every request.

src/app.js
akosasante: Sounds good; a better place for it would likely be a documentation folder in the repo instead; that way it's always available if someone needs to refer to it again; without being in the codebase

src/middleware/full-auth.js
akosasante: Sounds good; let's either remove the one that is not being used; or include a clear comment at the top of the file so that it's clear to all team members whether they should be using the functionality from one or the other :+1:

package.json
```diff
@@ -36,9 +36,14 @@
...
+    "@babel/preset-env": "^7.22.10",
+    "@babel/preset-typescript": "^7.22.5",
+    "@types/jest": "^29.5.3",
     "eslint": "^8.46.0",
     "eslint-config-prettier": "^8.9.0",
+    "jest": "^29.6.2",
     "nodemon": "^3.0.1",
-    "prettier": "^3.0.0"
+    "prettier": "^3.0.0",
+    "ts-jest": "^29.1.1"
```
akosasante: Why were the `@babel/...` and `ts-jest` libraries added/needed? I'm not seeing them in use in this PR.

src/app.js
```diff
@@ -56,21 +56,24 @@ app.use(
 app.use(passport.initialize());
 app.use(passport.session());
 
-// Database setup (using Mongoose)
-mongoose.set('strictQuery', true);
```
akosasante, line 60: I think we're now missing `mongoose.set('strictQuery', true);`

I think we would like for this to still be set to `true` because it does the following:

> When the `strict` option is set to **true**, Mongoose will ensure that only the fields that are specified in your schema will be saved in the database, and all other fields will not be saved (if some other fields are sent).

We can either include this line in the `app.js` or better yet, move it into the `db` folder or `db/connect.js` connectDB function.

src/controllers/userController.js
```diff
@@ -1,59 +1,97 @@
...
+    const user = await User.findOne({ _id: req.user.userId }).select(
+      '-password'
+    );
+    if (!user) {
+      throw new CustomError.NotFoundError('User not found');
+    }
+    res.status(StatusCodes.OK).json({ user });
+  } catch (error) {
+    next(error);
+  }
```
akosasante:
```suggestion
  const user = await User.findOne({ _id: req.user.userId }).select(
    '-password'
  );
  if (!user) {
    throw new CustomError.NotFoundError('User not found');
  }
  res.status(StatusCodes.OK).json({ user });
```

^ We don't need to wrap this in a `try/catch` statement because we're using `express-async-errors` library and have error handling middleware which will catch and handle the error and pass it along to the `next()` middleware if needed.

package.json
akosasante: Or did these come from one of the merged in branches? If so we can deal with cleaning this up separately no worries.

src/routes/authRoutes.js
```diff
@@ -5,6 +5,6 @@ const { register, login, logout } = require('../controllers/authController');
 
 router.post('/register', register);
 router.post('/login', login);
-router.post('/logout', logout);
+router.get('/logout', logout);
```
akosasante: I think this got lost from a previous PR; do you mind switching this back to a `post` ?

akosasante (approved):
A few small remaining comments; but after those you can feel free to merge when you're ready! :ship:

src/routes/authRoutes.js
mvstoyan: my mistake, switching this back to a post.

package.json
mvstoyan: removed

src/app.js
mvstoyan: Снимок экрана 2023-08-10 165510

mvstoyan: Снимок экрана 2023-08-10 165856

src/controllers/userController.js
mvstoyan: Снимок экрана 2023-08-10 170418

PR #14, opened 2023-08-08
[PS-61] Album mongoose model

NatachaKey:


src/models/Album.js
```diff
@@ -0,0 +1,63 @@
...
+    default: '/uploads/example.jpg',
+  },
+  releaseDate: {
+    type: Date,
+    required: false,
+  },
+  category: {
+    type: String,
+    required: [false, 'Please provide music category'],
+    enum: ['pop', 'rock', 'rap'],
```
akosasante, line 32: Since there can be a huge amount of genres and subgenres (rap-rock, heavy metal, bubblegum pop etc etc); I think it'll be difficult to cover all of them in an enum. My suggestion would be to leave this one as a plain string for now.

akosasante (approved):
One comment about genre, but otherwise everything looks good. Feel free to change in this PR or in the other one, whichever works best for you :+1:

src/models/Album.js
NatachaKey: @akosasante oh , sure,  I added an array here just to try on the frontend to add several options in a form, just to have most possible options there, does the 'plain string' means to delete the whole enum property?

akosasante: Yes correct remove just the `enum` line; since currently (I think) this will validate/only allow entries that are of one of those three genres/categories.

NatachaKey: Got it ! Thank you 😊

akosasante: Ah I reread your last comment and I think I missed or misundersood this part before: "just to try on the frontend to add several options in a form, just to have most possible options there". Sure that's a great idea and fits well with the music product! If that's something y'all want to try, you'll want to change the type like this, to allow an array of string entries: `type:[String]`

PR #15, opened 2023-08-08
[Ps-49]Fetch album from  spotify api on backend

NatachaKey:


fetchAlbumSpotify.js
```diff
@@ -0,0 +1,65 @@
...
+    });
+    console.log(response.body.albums);
+    // Save albums to the database
+    const albumsToSave = response.body.albums.items.map((item) => ({
+      artistName: item.artists.map((artist) => artist.name).join(', '),
+      albumName: item.name,
+      image: item.images?.[0].url,
+      releaseDate: item.release_date,
+      category: item.genres?.[0], // see Album model- enum - should we leave it as is?
+      spotifyUrl: item.href,
```
akosasante: Hmm, I think let's store the `item.external_urls.spotify` instead.

`href` is the API endpoint, eg: `'`[`https://api.spotify.com/v1/albums/18NOKLkZETa4sWwLMIm0UZ`](https://api.spotify.com/v1/albums/18NOKLkZETa4sWwLMIm0UZ)`'`

`external_urls.spotify` gives us the Web endpoint, eg: `'`[`https://open.spotify.com/album/18NOKLkZETa4sWwLMIm0UZ`](https://open.spotify.com/album/18NOKLkZETa4sWwLMIm0UZ)`'`  (which can be used by the frontend to embed an iframe or link to the spotify data)

fetchAlbumSpotify.js
```diff
@@ -0,0 +1,65 @@
...
+//       await Album.deleteMany({});
+//       console.log('Collection deleted successfully.');
+//     } catch (error) {
+//       console.error('Error deleting collection:', error);
+//     } finally {
+//       mongoose.connection.close();
+//     }
+//   }
+
+//   deleteCollection();
```
akosasante, line 66: :star\_struck:

src/app.js
```diff
@@ -51,10 +51,13 @@ app.use(
     saveUninitialized: false,
   })
 );
+<<<<<<< HEAD
 // SESSION_SECRET=generate a key for env
 //const crypto = require('crypto');
 //const secretKey = crypto.randomBytes(32).toString('hex');
 //console.log('Secret Key:', secretKey);
+=======
+>>>>>>> cc6efc9fc083c8429772ce8480c34bb39906f5c0
```
akosasante: Conflict resolve text got carried over :eyes:

src/controllers/authController.js
```diff
@@ -1,8 +1,26 @@
...
+    name: user.name,
+    userId: user._id,
+    role: user.role,
+  };
+  const token = jwt.sign(payload, process.env.JWT_SECRET, {
+    expiresIn: process.env.JWT_LIFETIME,
+  });
+  return token;
+};
+>>>>>>> cc6efc9fc083c8429772ce8480c34bb39906f5c0
```
akosasante: conflict resolve markers got carried over

src/controllers/authController.js
```diff
@@ -51,7 +69,23 @@ const login = async (req, res) => {
...
   register,
   login,
-};
\ No newline at end of file
+<<<<<<< HEAD
+};
+=======
+  logout,
+};
+>>>>>>> cc6efc9fc083c8429772ce8480c34bb39906f5c0
```
akosasante: conflict resolve markers

src/models/Album.js
```diff
@@ -1,61 +1,57 @@
...
-      validator: function (v) {
-        return /^https:\/\/open\.spotify\.com\/album\/[a-zA-Z0-9]+$/.test(v); //ensures that the URL starts with https://open.spotify.com/album/ followed by alphanumeric characters.
-      },
-      message: 'Invalid Spotify URL',
+const AlbumSchema = new mongoose.Schema(
+  {
+    artistName: {
+      type: String,
+      required: [true, 'Please provide artist name'],
+      maxlength: [300, 'Artist name can not be more than 100 characters'],
```
akosasante:
```suggestion
      maxlength: [300, 'Artist name can not be more than 300 characters'],
```

src/models/Album.js
```diff
@@ -1,61 +1,57 @@
...
+      trim: true,
+    },
+    albumName: {
+      type: String,
+      required: [true, 'Please provide album name'],
+      maxlength: [300, 'Album name can not be more than 100 characters'],
```
akosasante:
```suggestion
      maxlength: [300, 'Album name can not be more than 300 characters'],
```

src/models/Album.js
```diff
@@ -1,61 +1,57 @@
...
+      trim: true,
+    },
+    price: {
+      type: Number,
+      required: [false, 'Please provide album price'],
```
akosasante: Should this be `required: [true, <message>]` ? I'm not sure otherwise when the message will be shown if it's not required to be populated. And since you've put a default value it won't really need to be required anyway.

src/models/Album.js
```diff
@@ -1,61 +1,57 @@
...
+      type: String,
+      default: '/uploads/example.jpg',
+    },
+    releaseDate: {
+      type: Date,
+      required: false,
+    },
+    category: {
+      type: String,
+      required: [false, 'Please provide music category'],
```
akosasante: I think we can leave off the message? Or will it still show the message even though the field is not required?

src/models/Album.js
```diff
@@ -1,61 +1,57 @@
...
+      default: 0,
+    },
+    numOfReviews: {
+      type: Number,
+      default: 0,
+    },
+    user: {
+      type: mongoose.Types.ObjectId,
+      ref: 'User', // we are referencing the user model . if we don´t provide user- in postman we'll get  "msg": "Path `user` is required."
+      required: false, //CHANGE TO TRUE LATER!!!!!!
```
akosasante: In this code, we are saying that the album/product belongs to a single person. This _could_ be the case for something like seats in a concert venue, but in most online stores with physical products we likely want to be able to sell multiple copies of the same thing. So we end up with a "many-to-many" [relationship.](https://community.dbdiagram.io/t/tutorial-many-to-many-relationships/412) A user can buy multiple albums, and an album could be purchased by multiple users.

MongoDB is a No-SQL database so it doesn't have the usual way to do "joins" to handle this type of relationship. There's instead a few different ways we could do it: https://stackoverflow.com/a/46020968

1.  We could make this AlbumSchema have a `users` field, that is an array of `User` refs. That would allow us to do something like `album.users` to see all the users that bought a particular album. But then if we want to also be able to easily see all albums a particular user bought, then we need to also add to the `User` schema an `albums` field. And we need to remember to push to both of those whenever a purchase goes through. This is fine as long as we can make sure to remember to do so.
    1.  Here's a video tutorial that seems to do an okay job going over it: [https://www.youtube.com/watch?v=-El-86OzAhA](https://www.youtube.com/watch?v=-El-86OzAhA)
2.  We can create an intermediary schema (in a relational SQL table this would be called a "join table") like this:

```javascript
const PurchasedAlbumSchema = new Schema({
  albumId: { type: Schema.Types.ObjectId, ref: 'Album', required: true },
  userId: { type: Schema.Types.ObjectId, ref: 'User', required: true }
});
```

Then when someone buys an album, you'd record it like:

```javascript
await PurchasedAlbum.create({ albumId: <id>, userId: <id> });
```

For option number 2 there's two ways to handle fetching that info:

1) Virtuals:

```javascript
AlbumSchema.virtual('purchasedByUsers', {
  ref: 'PurchasedAlbum',
  localField: '_id',
  foreignField: 'userId'
});

UserSchema.virtual('purchasedAlbums', {
  ref: 'PurchasedAlbum',
  localField: '_id',
  foreignField: 'albumId'
});

# then in the code can use populate()
// Show user by id with all the albums they've purhcased
    let user = await User.findOne("id").populate({
      path: 'purchasedAlbums',
      populate: { path: 'albumId' }
    });
```

2) Aggregate queries:

```javascript
const userId = 'your_user_id';

try {
 const result = await PurchasedAlbum.aggregate([
  {
    $match: { userId: mongoose.Types.ObjectId(userId) }
  },
  {
    $lookup: {
      from: 'Album',
      localField: 'albumId',
      foreignField: '_id',
      as: 'albums'
    }
  },
  {
    $unwind: '$albums'
  },
  {
    $project: {
      _id: 0,
      userId: 1,
      albums: 1
    }
  }
]).exec();

console.log(result);

} catch (err) {
  logger.error(err);
}
```

akosasante:
FYI, I switched the base branch to `albumSchema` just to be able to easier see the changes specifically in this repo.

src/models/Album.js
NatachaKey: @akosasante actually I didn't want to hardcode and put the same price to every album but i don't know what to do in this case..  So i was thinking_
As we fetch data from spotify api and it doesn't have a 'price' property- let's imagine that as admins we need to put different prices to the albums that we've fetched from spotify api- the only idea i have is to make a patch request to the db for EVERY item- but maybe there is a better way?

src/models/Album.js
NatachaKey: @akosasante deleted it, i was trying to figure out why do we have an array here- did you mean that for using this array and error message we should implement populate method somewhere in the code?

src/models/Album.js
akosasante: That's a fair point. Here's my thought. What if we default it to 0? or make it not required; in which case if it doesn't have a price it'll be `null`

Then when doing a `getAllAlbums` request, we can check the user's role. If they are an `admin` we can return all albums, including ones that don't have a price set yet. If they are just a regular `user` we can have the query only return albums with prices.

On the admin side of things, the admin can update albums to change their properties, including setting a price. If we want to have a bulk endpoint `updateManyAlbums` that could be something where the request looks like:  
 

```javascript
PATCH /api/v1/albums

# request body is an array of albums to update:

[
  { id: <album object id>, price: 25.00 },
  { id: <album object id>, price: 2.25 },
  etc
]
```

And a mongoose query that uses the `bulkWrite` function to do multiple document updates in a single request to the db:

```
const bulkUpdateOps = req.body.map(update => ({
  updateOne: {
    filter: { _id: update.id },
    update: { price: update.price }
  }
}));

try {
  await Album.bulkWrite(bulkUpdateOps);
  logger.info('Albums updated successfully');
} catch (err) {
  logger.error(err);
}
```

In the frontend, there should be some sort of UI that allows for bulk editing. Maybe an ability to select multiple albums, and submit a price for them. 

Resources: 

*   [https://www.geeksforgeeks.org/mongoose-document-model-bulkwrite-api/](https://www.geeksforgeeks.org/mongoose-document-model-bulkwrite-api/)
*   https://kb.objectrocket.com/mongo-db/mongoose-bulkwrite-1213

src/models/Album.js
akosasante:
```suggestion
      required: false,
```

My suggestion is to just make it the value `false` to indicate it is not required.

The array here on \`required\` is just one of the ways Mongoose allows yo uto define the \`required\` option; it doesn't make the category field an array or anything. Overall, Mongoose allows a couple different types for this field: [https://mongoosejs.com/docs/api/schematype.html#SchemaType.prototype.required()](https://mongoosejs.com/docs/api/schematype.html#SchemaType.prototype.required())

It can be a boolean value (`required: true` or `required: false`). It can be an array, with `required: [true, "error message"]` . It can be just the error message (in which case mongoose assumes you want the field to be required) `required: "error message"`, or it can even be a function if you want the field to only be required in some cases.

src/models/Album.js
NatachaKey: @akosasante I've just tried to do it but i don't get the price changed in postman- i checked the fields in schema and set up patch request with admin authentication for albums round- can't find the problem- could you please take a look? https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/15/commits/321237cb8427951d64b61e7c73e1a6062f131ff3 

admin credentials for testing: {
   "email":"ki@gmail.com",
    "password":"secret"
}

NatachaKey: image

src/models/Album.js
NatachaKey: @akosasante sure, will leave it as 'required: false', i guess we could also delete it as it is false by default

src/models/Album.js
NatachaKey: @akosasante thank you! i've just tried to add virtuals- could you please take a look https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/15/commits/45e24d40b1ccf778befa9c8cc609ab0f4d4751be
should we add await PurchasedAlbum.create({ albumId: <id>, userId: <id> }); to ordersControllers.js or maybe i am missing smth and should add it somewhere else?

src/models/Album.js
akosasante: Yep, either way works!

src/models/Album.js
akosasante: Hmm, from what I'm seeing in `res.status(StatusCodes.OK).json({ albums: bulkUpdateOps});` you're returning just the list of `updateOne` objects that you generated from the request.

The bulkWrite function will return something like, https://mongoosejs.com/docs/api/model.html#Model.bulkWrite(): 

```javascript
{
  insertedCount: 0,
  matchedCount: 2,
  modifiedCount: 2,
  deletedCount: 0,
  upsertedCount: 0,
  upsertedIds: {},
  insertedIds: {}
}
```

If I fetch the individual Albums it does look to me like it successfully updated the price.

If you want to return the updated albums, you'll have to make a second call to the db:

```javascript
	const bulkUpdateOps = req.body.map(update => ({
		updateOne: {
			filter: { _id: update.id },
			update: { price: update.price },
		},
	}));
	const ids = req.body.map(update => update.id)

	const bulkWriteResponse = await Album.bulkWrite(bulkUpdateOps);
	logger.info(`${bulkWriteResponse.modifiedCount} Albums updated successfully`);
	const updatedAlbums = await Album.find({_id: { $in: ids}})
	res.status(StatusCodes.OK).json({ albums: updatedAlbums});
```

src/models/PurchasedAlbum.js
```diff
@@ -0,0 +1,10 @@
+const mongoose = require('mongoose');
+
+const PurchasedAlbumSchema = new mongoose.Schema({
+  albumId: { type: Schema.Types.ObjectId, ref: 'Album', required: true },
+  userId: { type: Schema.Types.ObjectId, ref: 'User', required: true },
```
akosasante:
```suggestion
  albumId: { type: mongoose.Schema.Types.ObjectId, ref: 'Album', required: true },
  userId: { type: mongoose.Schema.Types.ObjectId, ref: 'User', required: true },
```

src/models/PurchasedAlbum.js
```diff
@@ -0,0 +1,10 @@
...
+});
+
+const PurchasedAlbum = mongoose.model('AlbumPurchased', PurchasedAlbumSchema);
```
akosasante:
```suggestion
const PurchasedAlbum = mongoose.model('PurchasedAlbum', PurchasedAlbumSchema);
```
(since we're referring to the model as PurchasedAlbum elsewhere, like in the virtuals, it has to match this string.

src/models/PurchasedAlbum.js
akosasante: Otherwise we get the error `Uncaught ReferenceError: Schema is not defined`.

Other option would be to add `const { Schema } = require('mongoose');` to the top of the file to import that variable directly

src/models/PurchasedAlbum.js
akosasante: Otherwise we get the error `Uncaught MissingSchemaError: Schema hasn't been registered for model "PurchasedAlbum".` when we try to use the virtuals, because the virtuals code will look for `PurchasedAlbum` but here we've defined the model to have the name `AlbumPurchased`.

src/models/Album.js
akosasante: 1.  PurchasedAlbum model:

Apart from my other earlier comments, in the example below I've also renamed the properties from userId/albumId to user/album, just to make it clearer when we do the `populate` call that we'll usually be working with these as the full object rather than jsut the id (except maybe when inserting)

```javascript

const PurchasedAlbumSchema = new mongoose.Schema({
  album: { type: mongoose.Schema.Types.ObjectId, ref: 'Album', required: true },
  user: { type: mongoose.Schema.Types.ObjectId, ref: 'User', required: true },
});

const PurchasedAlbum = mongoose.model('PurchasedAlbum', PurchasedAlbumSchema);

module.exports = PurchasedAlbum;
```

2\. Virtual in the User model:

I would rename the virtual to be a bit clearer, that this is returning _all_ the albums this single _user_ has purchased

```javascript
UserSchema.virtual('purchasedAlbums', {
  ref: 'PurchasedAlbum',
  localField: '_id',
  foreignField: 'user'
});
```

3\. Virtual in the Album model

I would rename the virtual to be a bit clearer, that this is returning all of the users that have purchased _this album_

```
AlbumSchema.virtual('purchasedByUsers', {
  ref: 'PurchasedAlbum', //specifies that the virtual field purchasedAlbums is referencing the PurchasedAlbum model.
  localField: '_id', //_id field (from the db= id of the product) of the Album model is used as the local field to establish the relationship.
  foreignField: 'album' // album field in the PurchasedAlbum model is used as the foreign field to establish the relationship.
});
```

Now we can do all sorts of queries, like:

1.  Fetching a user from the database, including all the albums they've purchased
2.  Fetching an album from the database, including all the users that have purchased it
3.  Fetching a list of purchased albums for a user
4.  Fetching a list of users who purchased an album
5.  Saving to the database that a user purchased a particular album

```javascript
// 1 - example A:
const getCurrentUserWithPurchasedAlbums = async (req, res) => {
  // Show current user by id with all the albums they've purchased
  let usersWithAlbums = await User.findById(req.user.userId).populate({
      path: 'purchasedAlbums', // name of the virtual to populate
      populate: { path: 'album' } // nested populate, without this we would just get back a list of PurchasedAlbum models. 
      // But we just want to further populate to get the Album model refferred to in  the PurchasedAlbum.album proprty.
    });
    res.status(StatusCodes.OK).json({ usersWithAlbums, count: usersWithAlbums.length });
  }

// 1 - example B (updating existing getUser endpoint to take query parameters)
const getSingleUser = async (req, res) => {
  // Function to get a single user by ID
  // Find the user in the database based on the provided user ID and exclude the 'password' field
  let query = User.findOne({ _id: req.params.id }).select('-password');
  if (req.query.includeAlbums === "true") {
    query = query.populate({
      path: 'purchasedAlbums', // name of the virtual to populate
      populate: { path: 'album' } // nested populate, without this we would just get back a list of PurchasedAlbum models. 
      // But we just want to further populate to get the Album model refferred to in  the PurchasedAlbum.album proprty.
    });
  }
  const user = await query;
  if (!user) {
    // Throw a NotFoundError if the user is not found
    throw new CustomError.NotFoundError(`No user with id: ${req.params.id}`);
  }
  checkPermissions(req.user, user._id); // Check if the user has permission to access the user's information
  res.status(StatusCodes.OK).json({ user }); // Send a JSON response with the status code 200 OK and the user
};

// 2 - Same as above but you would populate with the object { path: 'purchasedByUsers', populate: { path: 'user' }}

// 3 - await PurchasedAlbum.find({user: user._id}).populate('album')
Returns a list of PurchasedAlbum records, with the album populated with the full model instance:
[
  {
    _id: new ObjectId("64d91dbece2137867e2f40d3"),
    album: {
      _id: new ObjectId("64d2a94c793389a43fc5a8d5"),
      artistName: 'Travis Scott',
      albumName: 'UTOPIA',
      price: 19.99,
      image: 'https://i.scdn.co/image/ab67616d0000b273881d8d8378cd01099babcd44',
      releaseDate: 2023-07-28T00:00:00.000Z,
      spotifyUrl: 'https://api.spotify.com/v1/albums/18NOKLkZETa4sWwLMIm0UZ',
      averageRating: 0,
      numOfReviews: 0,
      __v: 0,
      createdAt: 2023-08-08T20:45:00.942Z,
      updatedAt: 2023-08-13T17:56:20.086Z
    },
    user: new ObjectId("64d44bc8337399ccf9ad7e4d"),
    __v: 0
  }
]

// 4 - same idea, but instead `await PurchasedAlbum.find({album: albumId}).populate('user')

// 5 -  await PurchasedAlbum.create({albumId: album._id, userId: user._id})

```

And yes for (5) we would probably end up putting that somewhere in an `OrdersController`. We would save the order model. Then once the user's payment has gone through successfully and the order is permanent, we can do the code in (5) to store their purchased albums. 

(1) and (3) will probably be in the `UsersController`. (2) and (4) in the `AlbumsController`.

NatachaKey:
@akosasante Akos, could we name here in this example-   let **userWithAlbums** instead of  let usersWithAlbums  - because I suppose we get only one user by id and then add virtually his albums from PurchasedAlbums.js model?      

// 1 - example A:
const getCurrentUserWithPurchasedAlbums = async (req, res) => {
  // Show current user by id with all the albums they've purchased
  let usersWithAlbums = await User.findById(req.user.userId).populate({
      path: 'purchasedAlbums', // name of the virtual to populate
      populate: { path: 'album' } // nested populate, without this we would just get back a list of PurchasedAlbum models. 
      // But we just want to further populate to get the Album model refferred to in  the PurchasedAlbum.album proprty.
    });
    res.status(StatusCodes.OK).json({ usersWithAlbums, count: usersWithAlbums.length });
  }

NatachaKey:
@akosasante Akos could you please take a look, I think I ruined it- I was trying  to add controllers as you showed to fetch user and his albums and album and users that purchased it (maybe I am wrong- but are the options ( 1 and 3)  and ( 2 and 4 )are the same?) 

I had to resolve conflicts and now I am worried if deleted smth important-
also I get an error in terminal could you please take a look?.

image

akosasante:
> Akos, could we name here in this example- let userWithAlbums instead of let usersWithAlbums - because I suppose we get only one user by id and then add virtually his albums from PurchasedAlbums.js model?

Yes I think that makes sense 👍

akosasante:
> (but are the options ( 1 and 3) and ( 2 and 4 )are the same?)

To clarify, you don't necessarily need to implement all 5 of those but yes 1/3 and 2/4 are very similar. Let's say for 1/3. For 1, we'd be returning a user object, that has albums embedded, maybe just from the regular `GET /users/:userId` endpoint maybe with a query parameter like `GET /users/:userId?includeAlbums=true`:

```
{
  name: "John Doe",
  email: "jd@example.com",
  role: "user",
  profileImage: null,
  creditCardInfo: null,
  purchasedAlbums: [
    { artistName: "Taylor Swift", albumName: "evermore", etc},
    { artistName: "Karol G", albumName: "Unstoppable", etc}
  ] 
}
```

Vs. maybe if the front end just wants to fetch the list of albums themselves and doesn't need the user info (let's say if they're showing the albums in the user's profile page), they might want just an endpoint like `/users/:userId/albums` that just returns that users albums:

```
[
  {
    _id: new ObjectId("64d91dbece2137867e2f40d3"),
    album: { _id, artistName: "Taylor Swift", albumName: "evermore", etc},
    user: new ObjectId("64d44bc8337399ccf9ad7e4d"),
    __v: 0
  },
  {
    _id: new ObjectId("64d91dbece2137867e2f40d3"),
    album: { _id, artistName: "Karol G", albumName: "Unstoppable", etc},
    user: new ObjectId("64d44bc8337399ccf9ad7e4d"),
    __v: 0
  }
]
```


In that case, maybe we'd even want to further simplify the return value to something like:

```
const justAlbums = purchasedAlbums.map(purchasedAlbum => purchasedAlbum.album)
resp.status(200).json(justAlbums)

// returns:
[
  { _id, artistName: "Taylor Swift", albumName: "evermore", etc},
  { _id, artistName: "Karol G", albumName: "Unstoppable", etc},
]
```

akosasante:
That error looks like it's complaining about `userRoutes.js` line 22.

In my IDE it's showing that `getUsersPurchasedAlbums` as not being available. 

Aha, yeah in `userController` the name of the handler that we're exporting is: `getCurrentUserWithPurchasedAlbums`. So if we change that in the `userRoutes` file it should work.

src/controllers/authController.js
```diff
@@ -4,11 +4,26 @@ const { StatusCodes } = require('http-status-codes');
...
+    name: user.name,
+    userId: user._id,
+    role: user.role,
+  };
+  const token = jwt.sign(payload, process.env.JWT_SECRET, {
+    expiresIn: process.env.JWT_LIFETIME,
+  });
+  return token;
+};
+
```
akosasante: I think these lines came from resolving merge conflicts but can be removed. The `createJWT` function now lives in the `utils` folder

src/controllers/userController.js
```diff
@@ -9,6 +9,7 @@ const {
 
 const getAllUsers = async (req, res) => {
   // Function to get all users
+  console.log(req.user);
```
akosasante: Probably can be removed

src/controllers/userController.js
```diff
@@ -27,6 +28,13 @@ const getSingleUser = async (req, res) => {
 };
 
 const showCurrentUser = async (req, res) => {
+  // Function to get the current user's information
+  // Send a JSON response with the status code 200 OK and the current user's information
+  res.status(StatusCodes.OK).json({ user: req.user });
+};
```
akosasante: I think this also got caught up when resolving merge conflicts; the version of this function that we want is the one from main: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/controllers/userController.js#L29-L37

akosasante:
```js
const showCurrentUser = async (req, res) => {
  const user = await User.findOne({ _id: req.user.userId }).select('-password');

  if (!user) {
    throw new CustomError.NotFoundError('User not found');
  }

};
```

src/controllers/userController.js
```diff
@@ -27,6 +28,13 @@ const getSingleUser = async (req, res) => {
...
+
+// update user with user.save()
+const updateUser = async (req, res) => {
```
akosasante: I think lines 36 to 42 can be removed; `updateUser` became `updateCurrentUser` in main: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/controllers/userController.js#L39-L55

src/middleware/error-handler.js
```diff
@@ -1,3 +1,4 @@
+const UnauthenticatedError = require('../errors/unauthenticated');
```
akosasante: I think this can be removed, does not seem to be in use

src/middleware/full-auth.js
akosasante, line 1: This file probably came from resolving merge conflicts. It can be removed it is no longer being used in main.

src/models/PurchasedAlbums.js
akosasante, line 1: It looks like we now have both `PurchasedAlbums.js` and `PurchasedAlbum.js` we should only need one. The schema in `PurchasedAlbums` looks more correct to me; though maybe we should rename it to still follow the practice of naming models/schemas in their singular noun form.

src/routes/albumRoutes.js
```diff
@@ -10,14 +10,16 @@ const {
...
   getSingleAlbum,
+  updatePriceOfAlbums,
 } = require('../controllers/albumController');
 
 router
   .route('/')
   .post(authenticateUser, authorizePermissions('admin'), createAlbum) //only admin can create product
-  .get(getAllAlbums); //everyone can access all products- no middleware
+  .get(getAllAlbums) //everyone can access all products- no middleware
+  .patch(getAllAlbums,authenticateUser, authorizePermissions('admin'), updatePriceOfAlbums );
```
akosasante:
```suggestion
  .patch(authenticateUser, authorizePermissions('admin'), updatePriceOfAlbums );
```

src/controllers/albumController.js
```diff
@@ -37,9 +36,39 @@ const updateAlbum = async (req, res) => {
...
+  try {
+    await Album.bulkWrite(bulkUpdateOps);
+    logger.info('Albums updated successfully');
+  } catch (err) {
+    logger.error(err);
+  }
+  res.status(StatusCodes.OK).json({ albums: bulkUpdateOps });
+};
+
+const getAlbumsPurchasedByUser = async (req, res) => {
```
```suggestion
const getUsersWhoPurchasedAlbum = async (req, res) => {
```

src/controllers/albumController.js
```diff
@@ -37,9 +36,39 @@ const updateAlbum = async (req, res) => {
...
+  res.status(StatusCodes.OK).json({ albumsOfUser, count: albumsOfUser.length });
+};
+
 module.exports = {
   updateAlbum,
   createAlbum,
   getAllAlbums,
   getSingleAlbum,
+  updatePriceOfAlbums,
+  getAlbumsPurchasedByUser,
```
```suggestion
  getUsersWhoPurchasedAlbum,
```

src/routes/albumRoutes.js
```diff
@@ -10,14 +10,16 @@ const {
   updateAlbum,
   getAllAlbums,
   getSingleAlbum,
+  updatePriceOfAlbums,
 } = require('../controllers/albumController');
```
akosasante, line 16: Will we also be importing our `getUsersWhoPurchasedAlbum` handler and setting up a route for it?

src/models/Album.js
```diff
@@ -50,15 +53,17 @@ const AlbumSchema = new mongoose.Schema(
...
-      required: false, //CHANGE TO TRUE LATER!!!!!!
-    },
   },
   { timestamps: true } //timestamps provides fields of createdAt , updatedAt... and exact time,
 );
 
+AlbumSchema.virtual('purchasedAlbums', {
+  ref: 'PurchasedAlbum', //specifies that the virtual field purchasedAlbums is referencing the PurchasedAlbum model.
+  localField: '_id', //_id field (from the db= id of the product) of the Album model is used as the local field to establish the relationship.
+  foreignField: 'albumId' // albumId field in the PurchasedAlbum model is used as the foreign field to establish the relationship.
```
akosasante: Keep in mind, this should match the field in the `PurchasedAlbum` schema. If we end up using the new one you wrote; then this foreignField will need to be named `album` in order for the connections to work correctly

src/models/User.js
```diff
@@ -64,6 +67,13 @@ const UserSchema = new mongoose.Schema({
   creditCardInfo: creditCardSchema,
 });
 
+UserSchema.virtual('purchasedByUsers', {
+  ref: 'PurchasedAlbum',
+  localField: '_id',
+  foreignField: 'userId'
```
akosasante: Keep in mind that in the version of `PurchasedAlbum` schema that you just added, this field is called `user` so we'd need to update it here as well so that mongoose knows how to link them up

src/controllers/userController.js
```diff
@@ -83,11 +90,25 @@ const deleteSingleUser = async (req, res) => {
...
 
+//Fetching a user from the database, including all the albums they've purchased
+const getCurrentUserWithPurchasedAlbums = async (req, res) => {
+  // Show current user by id with all the albums they've purchased + see createTokenUser- userId comes from there
+  let usersWithAlbums = await User.findById(req.user.userId).populate({
+      path: 'purchasedAlbums', // name of the virtual to populate
+      populate: { path: 'album' } // nested populate, without this we would just get back a list of PurchasedAlbum models. 
+      // But we just want to further populate to get the Album model refferred to in  the PurchasedAlbum.album proprty.
+    });
+    res.status(StatusCodes.OK).json({ usersWithAlbums, count: usersWithAlbums.length });
```
akosasante: What is the length we're calculating here? `User.findById` will just return a single user object. We could do `userWithAlbums.purchasedAlbums.length` to return a count of albums this user has purchased

src/models/PurchasedAlbums.js
NatachaKey: @akosasante thank you!!!!

src/routes/userRoutes.js
```diff
@@ -12,12 +12,14 @@ const {
...
   deleteSingleUser,
+  getCurrentUserWithPurchasedAlbums,
 } = require('../controllers/userController');
 
 // Define routes for handling user-related operations
 router
   .route('/')
-  .get(authenticateUser, authorizePermissions('admin'), getAllUsers);
+  .get(authenticateUser, authorizePermissions('admin'), getAllUsers)
+  .get(authenticateUser, getCurrentUserWithPurchasedAlbums);
```
akosasante: I think this means both controller handlers are trying to use the path `GET /api/v1/users`.  I think in the end Express will overwrite the first one with the second one.

We probably don't want that so we should have a separate endpoint for getting a user with their albums. We could do something like:

```
router.route('/showMe/withAlbums').get(authenticateUser, getCurrentUserWithPurchasedAlbums)
```

or if we want to keep it as part of the existing `showMe` endpoint and have people do `GET /api/v1/users/showMe?includeAlbums=true`, then we could do:

```
router.route('/showMe').get(authenticateUser, showCurrentUser)

# then inside userController
const showCurrentUser = async (req, res) => {
  if (req.query.includeAlbums === "true") {
    // have the other handler handle this request instead
    return getCurrentUserWithPurchasedAlbums(req, res)
  } else {
    const user = await User.findOne({ _id: req.user.userId }).select('-password');

    if (!user) {
      throw new CustomError.NotFoundError('User not found');
    }

  }
};
```

src/routes/albumRoutes.js
akosasante: Now called; `getAlbumWithAllUsersWhoPurchasedIt` but my comment is still applicable :smile:

NatachaKey: @akosasante I think we should- just imported getUsersWhoPurchasedAlbum  to albumRoutes.js and set up new route .get(authenticateUser, authorizePermissions('admin'), getAlbumWithAllUsersWhoPurchasedIt)  //will commit soon

src/controllers/albumController.js
```diff
@@ -35,37 +34,75 @@ const updateAlbum = async (req, res) => {
...
+    updateOne: {
+      filter: { _id: update.id },
+      update: { price: update.price },
+    },
+  }));
+
+  const ids = req.body.map((update) => update.id); //Creates an array of _id values from the albums to be updated in the request body.
+
+  const bulkWriteResponse = await Album.bulkWrite(bulkUpdateOps); //he response object of bulkWrite () contains information about the number of modified documents.
+  logger.info(`${bulkWriteResponse.modifiedCount} Albums updated successfully`);
```
akosasante, line 51: Do we need to `require` logger for this to work or does it just work already?

src/routes/albumRoutes.js
```diff
@@ -10,19 +10,27 @@ const {
...
+  .patch(authenticateUser, authorizePermissions('admin'), updatePriceOfAlbums);
 
-  router.get('/filter', getFilteredAlbums);
+router.get('/filter', getFilteredAlbums);
 
 router
   .route('/:id')
   .get(getSingleAlbum) //everyone can access all products- no middleware
-  .patch(authenticateUser, authorizePermissions('admin'), updateAlbum); //only admin can update product
+  .get(
```
akosasante, line 34: Same thing here about using the same method (`get`) twice on the same route; Express won't know which one to run; so it'll just run the one that was the most recent to be defined. Either put this on a separate route (maybe `/:id/withUsers`) or query parameter on the existing `getSingleAlbum` endpoint

akosasante (commented):
Sorry just one small thing remaining with the album route

src/controllers/albumController.js
NatachaKey: @akosasante oh yes, when I run npm run lint:fix it gives me  50:3  error  'logger' is not defined  no-undef so I installed 'winston' library to log this messages, will commit it now

src/routes/albumRoutes.js
NatachaKey: @akosasante just did it setting up a separate route  , thank you, I am just curious about how to set up a query parameter on the existing getSingleAlbum endpoint - should we do smth like:

1.  in **albumController.js**

```
const getSingleAlbum = async (req, res) => {
  const { id: albumId } = req.params;
  const listOfUsersWhoPurchasedThisAlbum= req.query.listOfUsersWhoPurchasedThisAlbum=== 'true'; // Check if query parameter is 'true'

  const album = await Album.findOne({ _id: albumId });
  if (!album) {
    throw new CustomError.NotFoundError(`No album with id ${albumId}`);
  }

  let albumResponse = { album };

  if (listOfUsersWhoPurchasedThisAlbum) {
    // Include additional details based on the query parameter
    albumResponse = {
      ...albumResponse,
      additionalDetails: {
        // added additional details here
 let usersThatPurchasedThisAlbum = await Album.findOne({
    _id: req.params.id,
  }).populate({
    path: 'purchasedByUsers', 
    populate: { path: 'user' }, 
  });
        }
    };
  }

  res.status(StatusCodes.OK).json(albumResponse);
};
```

Will it be smth like this? And how would we get from the frontend if user introduced this query parameter? 
With smth like this? -->
 // Construct the URL with query parameters
```
const url = `/albums/${albumId}?listOfUsersWhoPurchasedThisAlbum=${listOfUsersWhoPurchasedThisAlbum}`;

// Make the GET request
fetch(url)
  .then(response => response.json())
  .then(data => {
    // Handle the response data
    console.log(data);
  })
  .catch(error => {
    // Handle errors
    console.error(error);
  });
```

Since only admin can see this info- we need to implement authorizePermissions('admin') middleware anywhere- how would we do it using queries. And what should admin do in browser to get the list of users who bought this album -press some kind of button?

src/controllers/albumController.js
```diff
@@ -1,7 +1,20 @@
...
+  transports: [
+    new winston.transports.Console(), // Output logs to the console
+    new winston.transports.File({ filename: 'app.log' }) // Output logs to a file
+    // You can add more transports as needed
+  ],
+  format: winston.format.combine(
+    winston.format.timestamp(), // Add timestamps to log messages
+    winston.format.printf((info) => `${info.timestamp} ${info.level}: ${info.message}`)
+  )
+});
```
akosasante: Sorry, my earlier comment wasn't clear 😢  We already have `winston` installed and set up in this repo by an earlier PR. It is in the `logs/logger.js` file. So to use that logger in a file we just need to import/require the logger variable from there: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/8541240605826e65cabc2396798f8babee572b07/logs/logger.js

src/routes/albumRoutes.js
akosasante: I think what you have would work;  but when you do `if(listOfUsersWhoPurchasedThisAlbum)` you could also check the requesting user's role to make sure they are an admin at that point. It would return a bit of redundant data in the `additionalDetails` since it would return a list of albums, then in `additionalDetails` a list of albums _again_ but this time with their users. You could do some `.map` to transform that data into a nicer format.

Here is my thought of how to do it in the backend to avoid that and avoid having to execute two separate queries:

```js
const getSingleAlbum = async (req, res) => {
  const {id: albumId} = req.params;
  // Check if query parameter is the string "true". If not or if the query parameter `?includeUsers` was not passed in at all
  // then just set the variable to false.
  const includePurchasingUsers = req.query.includeUsers === "true" ? true : false;

  // Mongoose will only actually make the query once we call `await`, so we can prep our query first, then actually execute it later
  // https://mongoosejs.com/docs/promises.html
  let albumQuery = Album.findOne({_id: albumId});
  // Only allow admins to view all the users who purchased an album:
  if (includePurchasingUsers && req.user.role === 'admin') {
    albumQuery = albumQuery.populate({path: 'purchasedByUsers', populate: {path: 'user'}});
  }
  // Run the query now
  const album = await albumQuery;

  if (!album) {
    throw new CustomError.NotFoundError(`No album with id ${albumId}`);
  }

  res.status(StatusCodes.OK).json({album});
};
```

For the frontend yes that looks correct to me! 👍 

```
const shouldIncludeUsersWhoPurchasedThisAlbum = true;
const url = `/albums/${albumId}?includeUsers=${shouldIncludeUsersWhoPurchasedThisAlbum}`;

// Make the GET request (I'm showing my example with Axios just cause that's what we're using in the frontend repo; but your fetch example works too)
fetch(url)
  .then(response => response.json())
  .then(data => {
    // Handle the response data
    console.log(data);
  })
  .catch(error => {
    // Handle errors
    console.error(error);
  });
```

I would imagine that either we would have an admin page that only admin can log into that shows this info and makes a request to the backend for this info.

akosasante (approved):

src/routes/albumRoutes.js
NatachaKey: @akosasante and using axios will it look like?
 ```
const shouldIncludeUsersWhoPurchasedThisAlbum = true;
const url = `/albums/${albumId}?includeUsers=${shouldIncludeUsersWhoPurchasedThisAlbum}`;

// Make the GET request using Axios
axios.get(url)
  .then(response => {
    // Handle the response data
    const data = response.data;
    console.log(data);
  })
  .catch(error => {
    // Handle errors
    console.error(error);
  });
```

src/controllers/albumController.js
NatachaKey: @akosasante sure! just deleted it and imported logger

PR #16, opened 2023-08-09
Swagger

DYA13:
**Edits added by Akosua**
- Added some additional swagger tags to the get user by id function so that future documentation can use it as an example: https://www.npmjs.com/package/swagger-autogen#response-language
- Fixed merge conflicts caused by `app.js` being split into `expressServer.js` + `app.js` (this was done to support the testing PR)
- Did some changes to make sure that the music store API is what is shown by default on the `/api-docs` route so that it's less confusing how to see it (had to use workaround from this Github issue: https://github.com/swagger-api/swagger-ui/issues/5710)
- renamed the script from `npm run swagger` to `npm run generate-api-docs` (it's longer but more clear about what steps are being done)


src/app.js
```diff
@@ -17,6 +17,21 @@ const logger = require('../logs/logger');
...
 
+// Serve the Swagger JSON file
+// app.get('/swagger.json', (req, res) => {
+//   res.sendFile('swagger-output.json');
+// });
+
+// Serve Swagger UI static files
+app.use(
+  '/swagger-ui',
+  express.static(path.join(__dirname, 'node_modules/swagger-ui-dist'))
```
akosasante: `__dirname` points to the directory of this file, so `./src`. But `node_modules` is in the root folder above. So currently this is point ing to `/path/to/backend/repo/src/node_modules/swagger-ui-dist` which is not an existing file. 

You can either:

1.  Use `..` to point to the folder _above_ this one: `app.use("/swagger-ui", express.static(path.join(__dirname, '../node_modules/swagger-ui-dist')`
2.  The docs for `swagger-ui-dist` show that it exports a function `absolutePath` that we can use here instead: [link](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/installation.md#:~:text=The%20module%27s%20contents%20mirror%20the%20dist%20folder%20you%20see%20in%20the%20Git%20repository.%20The%20most%20useful%20file%20is%20swagger%2Dui%2Dbundle.js%2C%20which%20is%20a%20build%20of%20Swagger%20UI%20that%20includes%20all%20the%20code%20it%20needs%20to%20run%20in%20one%20file.%20The%20folder%20also%20has%20an%20index.html%20asset%2C%20to%20make%20it%20easy%20to%20serve%20Swagger%20UI%20like%20so%3A)

```
const pathToSwaggerUi = require('swagger-ui-dist').absolutePath()

app.use("/swagger", express.static(pathToSwaggerUi));
```

3\. `app.use("/swagger", express.static(path.dirname(require.resolve('swagger-ui-dist'))));` 

(`require.resolve` is built-in to Node and returns the main file of a particular package, `path.dirname` is also a built-in method and returns the path to a directory containing a given file)

src/app.js
```diff
@@ -17,6 +17,21 @@ const logger = require('../logs/logger');
 const app = express();
 require('express-async-errors');
 
+// Serve the Swagger JSON file
+// app.get('/swagger.json', (req, res) => {
+//   res.sendFile('swagger-output.json');
+// });
```
akosasante: can be removed

src/app.js
```diff
@@ -17,6 +17,21 @@ const logger = require('../logs/logger');
...
+
+// Serve Swagger UI static files
+app.use(
+  '/swagger-ui',
```
akosasante: Maybe make this route `api-docs` ?

src/app.js
```diff
@@ -17,6 +17,21 @@ const logger = require('../logs/logger');
...
+  express.static(path.join(__dirname, 'node_modules/swagger-ui-dist'))
+);
+
+// Serve your generated Swagger specification
+app.get('/swagger', (req, res) => {
```
akosasante: maybe call this route `/api-docs.json` ?

package.json
```diff
@@ -8,7 +8,9 @@
     "start": "node app.js",
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
-    "format": "prettier --write ."
+    "format": "prettier --write .",
+    "swagger": "node ./swagger.js",
+    "start-gendoc": "node ./src/app.js"
```
akosasante: This just starts the app right, do we need this `start-gendoc` script?

akosasante (commented):
As we go along we should start using the swagger-autogen comments ([docs](https://www.npmjs.com/package/swagger-autogen#building-documentation-without-starting-the-project:~:text=section%20Response%20Language-,Endpoints,-The%20way%20to)) to do things like grouping the different routes.

For example, if in UserController , I add this comment to the getSingleUser handler:

```javascript
const getSingleUser = async (req, res) => {
  // #swagger.tags = ['Users']
  const user = await User.findOne({ _id: req.params.id });
  ...
```

And then regenerate the API docs with the `npm run swagger` command then I get:



It might be useful to include this in a documentation folder so that other teammates can work on adding some of those swagger documentation comments as they add new endpoints. 

One thing right now, is that you have to enter the path to the generated docs file in the Swagger UI explorer bar at the top; otherwise it by default just loads the example docs. I'll take a look at how to make it load your docs by default, but for now just entering in the path (eg: `/swagger` ) will work fine.

src/app.js
DYA13: Removed

src/app.js
DYA13: Renamed

src/app.js
DYA13: Renamed commit 262dab076f5099c97da0c7ee9c7969914bcdca62

package.json
DYA13: We can delete it actually.

src/app.js
DYA13: Thank you very much for your detailed explanation of this issue, I have changed according to point 2, commit 82d6c10a4f5e87b0caae768677210b10170f13ee

DYA13: As for the swagger documentation on my backend I have successful response but on localhost  by the path / swagger I see the example of swagger UI but not the actual routes to login/logout 
I don't manage to see as in your screenshot
  image

src/app.js
DYA13: commit 87eeca14419b5365397d727763dcfd0e4a8d72d9

src/app.js
DYA13: commit fcca0e0f2998c05fa9dd61004284939c74df7ee0

src/app.js
DYA13: Akos , what do you think if we disable cors in app.js ? May be it will help me to see the endpoints but not swagger html as it is now happening? 
something like this `app.use(cors()); // Allow all origins `

akosasante: Sorry this is what I meant earlier, if you put in the path to the swagger json route `/swagger.json` that's defined in `app.js` into the search bar at the top; it will show the generated API docs instead of the example ones. I haven't had a chance yet but I'll look into if there's a way to by default show that instead of the example pet store endpoints. But I think as long as you document it (just a quick sentence in the readme or in a documentaion folder) so others know how to find it; this is okay to merge for now.

akosasante (approved):
Just the `start-gendoc` command left to remove, but otherwise looks good to me :+1:

package.json
DYA13: I have deleted this line , it should be there actually
`  "scripts": {
    "dev": "nodemon src/app.js",
    "start": "node app.js",
    "lint": "eslint .",
    "lint:fix": "eslint . --fix",
    "format": "prettier --write ."
  },`

DYA13: Done commit b3b7a9f8646209f3ee622108ca54cf7510a28261

src/app.js
DYA13: Akos I am sorry that I ask again but if I use the path  /swagger.json  I see the json file but not swagger one. May be I do something wrong? 

image

image

akosasante:
This is what I mean:
- navigate to localhost:8000/api-docs
- In the Swagger "Explore" bar type in the route to render the swagger json you generated: `/swagger.json`

https://github.com/Code-the-Dream-School/dd-prac-team4-back/assets/9094098/7ea04b7c-4f6e-4fcc-a64e-2914206977c9

@DYA13

PR #21, opened 2023-08-12
Wishlist/endpoints

DYA13:
Ticket PS-64


src/routes/wishlistRoutes.js
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    // Add album to wishlist and save
+    wishlist.albums.push(album._id);
+    await wishlist.save();
+
+    return res.status(200).json(wishlist);
+  } catch (error) {
+    return res.status(500).send('Server error');
+  }
```
akosasante:
```suggestion
    if (
      !mongoose.Types.ObjectId.isValid(wishlist_id) ||
      !mongoose.Types.ObjectId.isValid(album_id)
    ) {
      return res.status(400).send('Invalid wishlist_id or album_id');
    }

    // Find the Wishlist and Album documents
    const wishlist = await Wishlist.findById(wishlist_id);
    const album = await Album.findById(album_id);

    // Check if wishlist and album exist
    if (!wishlist || !album) {
      return res.status(404).send('Wishlist or album not found');
    }

    // Check if album is already in wishlist
    if (wishlist.albums.includes(album._id)) {
      return res.status(200).json(wishlist);
    }

    // Add album to wishlist and save
    wishlist.albums.push(album._id);
    await wishlist.save();

    return res.status(200).json(wishlist);
```

We shouldn't need to wrap our controller code in try/catch because we're using the `express-async-errors` library. Any erorr sthat happen and get thrown in here, will be handled by the errorHandler middleware file.

src/routes/wishlistRoutes.js
akosasante, line 1: Can we follow the same pattern as the other routes; and just do the definition in this file and have a separate `wishlistController` file for the handlers?

src/routes/wishlistRoutes.js
```diff
@@ -0,0 +1,79 @@
+const express = require('express');
+const Wishlist = require('./Wishlist');
+const Album = require('./Album');
```
akosasante: You may need to rebase on `main` and make sure you have these two models. And then make sure this `require` is using the correct path to use these models

src/routes/wishlistRoutes.js
DYA13: Yes , you are right, thank you , commit 489aa8e60f2ee602eafdc253ead5c9ed92ad8495

src/routes/wishlistRoutes.js
DYA13: I have merged the changes from main and edited the paths to use these models , commit 9beacdd0983ae162b3cbd19386518adfa1609892

src/routes/wishlistRoutes.js
DYA13: Commit 6bd3c85395f33ef870020c4635eeb7e68f3fc220

src/app.js
akosasante, line 1: Merge conflict unresolved :eyes: 

Marina's recent PR for implementing testing, splits the Express set up code into an `src/expressServer.js` file. So you should move your code that adds the `wishlistRoutes` to that file.

Then make sure to keep just the `app` startup bits in this file (see what it looks like in main: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/app.js)

src/app.js
```diff
@@ -1,5 +1,82 @@
...
+const userRouter = require('./routes/userRoutes');
+
+app.use('/api/v1/auth', authRouter);
+app.use('/api/v1/users', userRouter);
+app.use('/api/v1/wishlist', wishlistRoutes);
+
+// Error handling middleware (must be defined after all other routes and middleware)
+//add later
+=======
 const { app, connectDB } = require('./expressServer');
```
akosasante: We basically just want to keep line 77 down.

src/routes/wishlistRoutes.js
akosasante: I'm not seeing a `wishlistController` in that commit?

DYA13: commit 71a4cc9dd743548f43c63e3b42c4ca1a4dd7a58c

src/app.js
DYA13: I accepted the changes which splits Express set up code , commit ea6d18dde4f0161b150b6d61b2f6e95232f43d8d

src/app.js
DYA13: commit ea6d18dde4f0161b150b6d61b2f6e95232f43d8d

akosasante (commented):
I think with the `git merge`'s one thing that got overwritten, is now we have all the routes defined in `expressServer.js` instead of `app.js`. So you'll need to make sure you have the route defined there:

```javascript
const wishlistRoutes = require('./wishlistRoutes'); 


app.use('/api/v1/wishlist', wishlistRoutes);
```

src/controllers/wishlistController.js
```diff
@@ -0,0 +1,63 @@
+const mongoose = require('mongoose');
+const Wishlist = require('./wishlist');
```
akosasante:
```suggestion
const Wishlist = require('../models/Wishlist');
```

src/routes/wishlistRoutes.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+  // Check if album is already in wishlist
+  if (wishlist.albums.includes(album._id)) {
+    return res.status(200).json(wishlist);
+  }
+  // Add album to wishlist and save
+  wishlist.albums.push(album._id);
+  await wishlist.save();
+  return res.status(200).json(wishlist);
+});
```
akosasante: We defined this in the controller, so we should call out to the controller here:

```suggestion
router.patch('/wishlist/:wishlist_id/add_album/:album_id', addAlbumToWishlist);
```

src/routes/wishlistRoutes.js
```diff
@@ -0,0 +1,69 @@
...
+      // Remove album from wishlist and save
+      wishlist.albums.pull(album_id);
+      await wishlist.save();
+
+      return res.status(200).json(wishlist);
+    } catch (error) {
+      return res.status(500).send('Server error');
+    }
+  }
+);
```
akosasante: We defined this in the controller, so we should call out to the controller here:

```suggestion
router.patch('/wishlist/:wishlist_id/remove_album/:album_id', removeAlbumFromWishlist);
```

src/routes/wishlistRoutes.js
```diff
@@ -0,0 +1,69 @@
+const express = require('express');
+const Wishlist = require('../models/Wishlist');
+const Album = require('../models/Album');
```
akosasante:
```suggestion
const { addAlbumToWishlist, removeAlbumFromWishlist } = require('../controllers/wishlistController')
```

src/controllers/wishlistController.js
```diff
@@ -0,0 +1,63 @@
...
+    ).populate('albums');
+
+    if (!wishlist) {
+      return res.status(404).json({ error: 'Wishlist not found' });
+    }
+
+    res.json(wishlist);
+  } catch (error) {
+    res.status(500).json({ error: 'Could not update wishlist' });
+  }
```
akosasante: We don't need to wrap this in a `try/catch`, just let the error handler middlewre handle any unhandled errors:

```suggestion

    const { wishlist_id, album_id } = req.params;

    if (
      !mongoose.Types.ObjectId.isValid(wishlist_id) ||
      !mongoose.Types.ObjectId.isValid(album_id)
    ) {
      return res.status(400).json({ error: 'Invalid ID format' });
    }

    const wishlist = await Wishlist.findByIdAndUpdate(
      wishlist_id,
      { $addToSet: { albums: album_id } },
      { new: true }
    ).populate('albums');

    if (!wishlist) {
      return res.status(404).json({ error: 'Wishlist not found' });
    }

    res.json(wishlist);
```

src/controllers/wishlistController.js
```diff
@@ -0,0 +1,63 @@
...
```
akosasante:
```suggestion

    const { wishlist_id, album_id } = req.params;

    if (
      !mongoose.Types.ObjectId.isValid(wishlist_id) ||
      !mongoose.Types.ObjectId.isValid(album_id)
    ) {
      return res.status(400).json({ error: 'Invalid ID format' });
    }

    const wishlist = await Wishlist.findByIdAndUpdate(
      wishlist_id,
      { $pull: { albums: album_id } },
      { new: true }
    ).populate('albums');

    if (!wishlist) {
      return res.status(404).json({ error: 'Wishlist not found' });
    }

    res.json(wishlist);
```

remove try/catch

src/controllers/wishlistController.js
```diff
@@ -0,0 +1,63 @@
...
+    const { wishlist_id, album_id } = req.params;
+
+    if (
+      !mongoose.Types.ObjectId.isValid(wishlist_id) ||
+      !mongoose.Types.ObjectId.isValid(album_id)
+    ) {
+      return res.status(400).json({ error: 'Invalid ID format' });
+    }
+
+    const wishlist = await Wishlist.findByIdAndUpdate(
```
akosasante: One thing I missed in writing the ticket; we should probably only let users edit their own wishlist right?

So we could do the fetch here with `findOneAndUpdate`:

```javascript
await Wishlist.findOneAndUpdate({ _id: wishlist_id, user: req.user.userId }, ...)
```

same with the remove from wishlist handler

src/routes/wishlistRoutes.js
akosasante: Maybe also use the `authenticateUser` middleware here and for the remove endpoint to ensure a logged in user is calling this endpoint?

src/controllers/wishlistController.js
DYA13: added routes to expressServer.js , commit 3abf17afb6fece20abeba21504a18f9c98a5bf4a

DYA13: the path edited to models/Wishlist.js in wishlistController.js , commit 550d51a112a57d6930ebe6ec0de2cdffde0bdb73

src/routes/wishlistRoutes.js
DYA13: done , commit 3104af4e150f1aca406aec4875f740bfac867486

src/controllers/wishlistController.js
DYA13: try/catch block removed , commit 76ce464606bb3213e7ad047e9371024ae8917774

src/controllers/wishlistController.js
DYA13: commit 76ce464606bb3213e7ad047e9371024ae8917774

src/controllers/wishlistController.js
DYA13: I have updated function findOneAndUpdate with the suggested parameter , commit e87e86cbbbfc98d977c808a5dd9fb6d23ced9dc9
Akos sorry, but I didn't understand the issue regarding function removeAlbumFromWishlist, could you please give more details?

src/controllers/wishlistController.js
```diff
@@ -0,0 +1,55 @@
...
+  ).populate('albums');
+  if (!wishlist) {
+    return res.status(404).json({ error: 'Wishlist not found' });
+  }
+  res.json(wishlist);
+}
+
+// Remove an album from a wishlist
+async function removeAlbumFromWishlist(req, res) {
+  try {
```
akosasante: This one can also be unwrapped from the try/catch

src/routes/wishlistRoutes.js
akosasante: Did you push this change? I still see the handler being defined in the routes file

src/controllers/wishlistController.js
akosasante: The linked commit only removes the try/catch from the addToWishlist but not from the removeToWishlist

src/controllers/wishlistController.js
akosasante: For both `addToWislist` and `removeFromWishlist` handlers, we're fetching the wishlist, and doing the change requested. But at no point do we check if the user who is making this request, is the correct user for the wishlist id that they gave. Of course in the UI this should not be made possible but users may use the API or hackers/malvolent actors may try to spoof requests in the browser.

My suggestion was to change both of the controller handlers to use `findOneAndUpdate` _**instead of** _ `findByIdAndUpdate`. The reason is that `findOneAndUpdate` allows us to give parameters to search by apart from just `id`.

So with `await Wishlist.findOneAndUpdate({ _id: wishlist_id, user: req.user.userId }, ...)`, the result is that the wish list will only be fetched and updated if there is a wishlist with the matching id _that belongs to the requesting user_. Otherwise, it will not find anything to update.

src/controllers/wishlistController.js
DYA13: removed try/catch block , commit 4624cfaad9cbaf4a4423fae5a54028100a02d56e

src/controllers/wishlistController.js
DYA13: commit 0fd9894814d92ce9ac6f3d62afbb860e6542fcfd

src/controllers/wishlistController.js
DYA13: Akos , could you check please? commit fdbd9be819f5e0d78c1a3351855b7aca6e40e9c5

src/routes/wishlistRoutes.js
```diff
@@ -0,0 +1,16 @@
...
+  removeAlbumFromWishlist,
+} = require('../controllers/wishlistController');
+
+const router = express.Router();
+
+router.patch('/wishlist/:wishlist_id/add_album/:album_id', addAlbumToWishlist);
+
+router.patch(
+  '/wishlist/:wishlist_id/remove_album/:album_id',
+  removeAlbumFromWishlist
```
akosasante, line 13: Ohh I missed this earlier but this route is mounted in `expressServer` to : `/api/v1/wishlist`

So if we then add the route here as `/wishlist/:wishlist_id/add_album/:album_id` it's going to mean that the caller has to call: `localhost:8000/api/v1/wishlist/wishlist/123/add_album/123`. I haven't tested it myself but I'm guessing that's what you've had to do in Postman?

We should just simplify it to:

```suggestion
router.patch('/:wishlist_id/add_album/:album_id', addAlbumToWishlist);

router.patch(
  '/:wishlist_id/remove_album/:album_id',
  removeAlbumFromWishlist
```

src/controllers/wishlistController.js
```diff
@@ -0,0 +1,51 @@
...
+// Add an album to a wishlist
+async function addAlbumToWishlist(req, res) {
+  const { wishlist_id, album_id } = req.params;
+  if (
+    !mongoose.Types.ObjectId.isValid(wishlist_id) ||
+    !mongoose.Types.ObjectId.isValid(album_id)
+  ) {
+    return res.status(400).json({ error: 'Invalid ID format' });
+  }
+  const wishlist = await Wishlist.findByIdAndUpdate(
```
akosasante:
```suggestion
  const wishlist = await Wishlist.findOneAndUpdate(
```

akosasante (commented):
One last fix for the routes and the controller, then you should be good to go :+1:

src/controllers/wishlistController.js
DYA13: Suggested changes applied  👍

akosasante (approved):
LGTM 🚀

PR #25, opened 2023-08-14
Controller, router for retrieving an album by id created

DYA13:
Ticket PS-51, the endpoint that returns details about a particular record has been created

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -75,9 +75,6 @@ app.get('/health', (req, res) => {
   res.status(200).send('OK');
 });
 
-// Use the albumRouter for /albums endpoint
-app.use('/albums', albumRouter);
-
 // Error handling middleware (must be defined after all other routes and middleware)
 app.use(notFoundMiddleware); // Not found middleware to handle invalid routes
 app.use(errorHandlerMiddleware); // Error handler middleware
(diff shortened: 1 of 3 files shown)


src/app.js
```diff
@@ -76,6 +75,8 @@ app.get('/health', (req, res) => {
   res.status(200).send('OK');
 });
 
+// Use the albumRouter for /albums endpoint
+app.use('/albums', albumRouter);
```
akosasante: Isn't this already defined on line 71 under the `/api/v1/albums` route?

src/routes/albumRoutes.js
```diff
@@ -1,5 +1,7 @@
 const express = require('express');
 const router = express.Router();
+const albumController = require('./albumController');
```
akosasante: Instead of requiring the full `albumController` we can just add the handler to the code on line 10-15 to follow the existing pattern in this file

akosasante (changes requested):
Sorry @DYA13 I hadn't realized when prepping this ticket and when you asked earlier that this code was already implemented from another PR. Let's just close this ticket. I'll be taking a look at the open tickets and making sure they're up-to-date with latest changes. Apologies for the miss here.

src/app.js
DYA13: Yes you are right , commit 29537938d7eeb13ee49abc5790edae8da5475fdb

src/routes/albumRoutes.js
DYA13: Done , commit 6bd3c85395f33ef870020c4635eeb7e68f3fc220

PR #27, opened 2023-08-17
order schema

mvstoyan:


mvstoyan:
I'm working on the next ticket, create an order, and will check how work the order scheme

mvstoyan:
@akosasante  I added a function called updateOrderStatus to automatically update order statuses every minute using setInterval. I also modified the middleware hooks to call this function before find and findOne operations on the Order model.

However, there's an issue: the order statuses are not updating as expected. I've confirmed that the middleware hooks are indeed triggering, but the updates aren't happening. I'm not sure what could be causing this problem.

src/models/Orders.js
```diff
@@ -0,0 +1,105 @@
...
+  // The quantity field specifies the number of items in the order
+  quantity: {
+    type: Number,
+    required: true, // Quantity is required
+    min: 1, // Minimum quantity allowed is 1
+  },
+});
+
+// Create the OrderItem model using the defined schema
+const OrderItem = mongoose.model('OrderItem', orderItemSchema);
```
akosasante: It's fine to leave this; but we probably won't ever need to create an OrderItem directly/individually apart from an `Order` so we don't _need_ to declare it as a model. But it doesn't hurt!

src/models/Orders.js
```diff
@@ -0,0 +1,105 @@
...
+
+// Schema for the main Order model
+const orderSchema = new mongoose.Schema({
+    // The user field references the User model and is required
+  user: {
+    type: mongoose.Schema.Types.ObjectId,
+    ref: 'User', // Referencing the User model
+    required: true,
```
akosasante: For all the `required` fields, it's not necessary but can be helpful for error messaging, to include a message to describe what was required (eg: `required: [true, 'Please provide user']`)

src/models/Orders.js
```diff
@@ -0,0 +1,105 @@
...
+  // The subtotal field represents the total price before taxes
+  subtotal: {
+    type: Number,
+    required: true,
+  },
+  // The total field represents the total price after taxes
+  total: {
+    type: Number,
+    required: true,
+  },
```
akosasante: Hopefully there wouldn't ever be a case where someone passes in negative numbers to the API; but just in case, we should set some validation here `min: 0`

src/models/Orders.js
```diff
@@ -0,0 +1,105 @@
...
+    required: true, // Quantity is required
+    min: 1, // Minimum quantity allowed is 1
+  },
+});
+
+// Create the OrderItem model using the defined schema
+const OrderItem = mongoose.model('OrderItem', orderItemSchema);
+
+// Schema for the main Order model
+const orderSchema = new mongoose.Schema({
```
akosasante: The order schema is missing the `orderItems` as an embedded part of the document. That way we can easily know what was ordered as part of a given Order:

```javascript
const orderSchema = new mongoose.Schema({
	... all the existing stuf...,
	orderItems: [OrderItemSchema],
})
```

src/models/Orders.js
```diff
@@ -0,0 +1,105 @@
...
+  try {
+    // Get the current date and subtract 1 minute
+    const oneMinutesAgo = new Date(Date.now() - 1 * 60 * 1000);
+    
+    // Update orders with a status of "pending" that were created 1 minute ago or earlier,
+    // and set the order status to "cancelled"
+    const result = await Order.updateMany(
+      { orderStatus: 'pending', createdAt: { $lte: oneMinutesAgo } },
+      { $set: { orderStatus: 'cancelled' } }
+    );
```
akosasante: We can leave it like this for now for testing; but that's a pretty short time to cancel an order :sweat\_smile: . I can imagine a situation as a user where  I proceeded to Checkout (order is created) and then had to go and find my credit card, and it took me more than 1 minute and I come back and see that my order has been set to "cancelled".

src/models/Orders.js
akosasante, line 1: Just for the sake of following a consistent naming pattern, let's rename this file `Order.js` singular like the other models please

src/models/Orders.js
mvstoyan: screen

src/models/Orders.js
mvstoyan: min

src/models/Orders.js
mvstoyan: orderItems

src/models/Orders.js
mvstoyan: This was only for testing purposes, as I was on the verge of falling asleep every 10 minutes while testing😴. I will fix this once we're done with the shopping cart and payments.

akosasante:
> @akosasante I added a function called updateOrderStatus to automatically update order statuses every minute using setInterval. I also modified the middleware hooks to call this function before find and findOne operations on the Order model.
> 
> However, there's an issue: the order statuses are not updating as expected. I've confirmed that the middleware hooks are indeed triggering, but the updates aren't happening. I'm not sure what could be causing this problem.

Hmm, when I pulled your branch earlier to test; the `setInterval` was working for orders I created and was setting them to `cancelled` if they were more than a minute old when the interval triggered. I'm not sure why it was not working on your end. But just having it happen at the time when a fetch happens I think is reasonable for now since we don't really need live/realtime view of the orders in the frontend.

src/models/Orders.js
akosasante: We could add this type of messaging for each of the required fields

akosasante (commented):
small comments remaining (file name, and required fields).

As well, the CI seems to show that formatting needs to be run on this file. Make sure to always run `npm run format` and `npm run lint:fix` . (Or if you pull the latest changes from main, then every time you do a git commit, the library [husky](https://duncanlew.medium.com/getting-started-with-husky-and-lint-staged-for-pre-commit-hooks-c2764d8c9ae) will do this automatically for you using [git pre-commit hooks](https://www.atlassian.com/git/tutorials/git-hooks)).

src/models/Orders.js
mvstoyan: Renamed

src/models/Orders.js
mvstoyan: screen5

mvstoyan:
@akosasante I wanted to showcase the updateOrderStatus function and explain what I had in mind. However, in this branch, I didn't have the relevant code. So, I suspect I might have taken the wrong step and merged with PR90. You might have become quite confused between these branches. I'm genuinely sorry. Now, I have all my code here, but additional files have also appeared. For instance, alongside my orderController.js, an ordersController.js with comments has appeared.

In any case, regarding the updateOrderStatus function, I do see results with setInterval(updateOrderStatus, 1 * 60 * 1000). I see updates reflected in the database and in the terminal console. However, when I try to implement just the .pre('find' and .pre('findOne' as per the ticket, without using setInterval, they don't seem to work. The console.log doesn't even trigger, and I can't understand the reason behind this.
screen8

src/controllers/orderController.js
```diff
@@ -0,0 +1,45 @@
+const { Order, OrderItem } = require('../models/Order');
+const { StatusCodes } = require('http-status-codes');
+const stripe = require('stripe')(process.env.STRIPE_SECRET_KEY);
+const { BadRequestError } = require('../errors');
+
+const createOrder = async (req, res) => {
+  const { orderItems, subtotal, tax, total } = req.body;
+
+  // Create an array of OrderItem objects
+  const orderItemObjects = orderItems.map((item) => new OrderItem(item));
```
akosasante: Since we took out the `OrderItem` model; we can remove this code (it won't work currently it will give an error that says `no constructor found for OrderItem` or something like that

src/controllers/orderController.js
```diff
@@ -0,0 +1,45 @@
+const { Order, OrderItem } = require('../models/Order');
```
akosasante: Since we took out the `OrderItem` model definition; there's nothing to import here as `OrderItem`, can remove that and just leave `{ Order }`

src/controllers/orderController.js
```diff
@@ -0,0 +1,45 @@
...
+  if (!orderItems?.length) {
+    throw new BadRequestError(
+      'Unable to create order because the order items are missing.'
+    );
+  }
+
+  // Create an Order document in the database
+  const order = await Order.create({
+    user: req.user.userId,
+    orderItems: orderItemObjects,
```
akosasante: In which case this can just be `orderItems: orderItems` (or shorthand, just `orderItems`)

And since the OrderSchema, includes the `OrderItemsSchema`, the values will still undergo validation without us needing to create a separate model/table for them.

src/models/Order.js
```diff
@@ -0,0 +1,110 @@
...
+    );
+    console.log('Update result:', result);
+  } catch (error) {
+    console.error('Error updating orders:', error);
+  }
+};
+
+// Call the updateOrderStatus function every 1 minute (60 * 1000 milliseconds)
+setInterval(updateOrderStatus, 1 * 60 * 1000);
+
```
akosasante, line 89: This `setInterval` code worked for me :+1:  I see the \`console.log("update result") logging in the terminal every minute.

I created an order using the `POST /api/v1/orders` endpoint. Then when the order was a minute old, when the next `setInterval` invokation occurred it updated the order to cancelled (verified by checking the db at: [https://cloud.mongodb.com/v2/64c51ca2903b6767054538fa#/metrics/replicaSet/64c51ee71446053e9f3c7b7f/explorer/test/orders/find)](https://cloud.mongodb.com/v2/64c51ca2903b6767054538fa#/metrics/replicaSet/64c51ee71446053e9f3c7b7f/explorer/test/orders/find)). and the update result log that logged next looked like:

```
Update result: {
  acknowledged: true,
  modifiedCount: 1,
  upsertedId: null,
  upsertedCount: 0,
  matchedCount: 1
}
```

src/models/Order.js
```diff
@@ -0,0 +1,110 @@
...
+});
+
+// Middleware: Update order statuses before executing a findOne operation
+orderSchema.pre('findOne', async function (next) {
+  console.log('Pre-findOne started');
+  // Call the updateOrderStatus function to update order statuses
+  await updateOrderStatus();
+  console.log('Pre-findOne finished');
+  next();
+});
```
akosasante, line 106: Both of these worked for me as well! 

I created an order through the `POST /api/v1/orders` endpoint. Then, since there's not yet an endpoint for fetching orders, I instead opened a terminal window running node with all the `.env` variables: `node -r dotenv/config`

Then I ran the following in the node REPL to test that the above was working (I first commented out the `setInterval` just to make sure I was only testing the middlewares):  
 

```
> const connectDb = require('./src/db/connect')
undefined
> const db = await connectDb(process.env.MONGO_URL)
undefined

# Do a findById fetch (this should call the `pre-findone` middleware, but since 1 minute hasn't passed yet, nothing will be updated)
> await Order.Order.findById("64e1218480db8b4c9f15ff4b")
Pre-findOne started
Update result: {
  acknowledged: true,
  modifiedCount: 0,
  upsertedId: null,
  upsertedCount: 0,
  matchedCount: 0
}
Pre-findOne finished
{
  _id: new ObjectId("64e1218480db8b4c9f15ff4b"),
  user: new ObjectId("64d44bc8337399ccf9ad7e4d"),
  orderStatus: 'pending',
  tax: 0.14,
  subtotal: 50,
  total: 57,
  orderItems: [
    {
      album: new ObjectId("64d2a94c793389a43fc5a8ec"),
      quantity: 2,
      _id: new ObjectId("64e1218480db8b4c9f15ff4c")
    },
    {
      album: new ObjectId("64d2a94c793389a43fc5a8e7"),
      quantity: 5,
      _id: new ObjectId("64e1218480db8b4c9f15ff4d")
    }
  ],
  createdAt: 2023-08-19T20:09:40.331Z,
  updatedAt: 2023-08-19T20:09:40.735Z,
  __v: 0,
  paymentIntentId: 'pi_3NgvSqKnb0YvaiB30BpqeFkt'
}


# Now wait for 1 minute, and do the same thing again but this time the order will have been updated to 'cancelled':
> await Order.Order.findById("64e1218480db8b4c9f15ff4b")
Pre-findOne started
Update result: {
  acknowledged: true,
  modifiedCount: 1,
  upsertedId: null,
  upsertedCount: 0,
  matchedCount: 1
}
Pre-findOne finished
{
  _id: new ObjectId("64e1218480db8b4c9f15ff4b"),
  user: new ObjectId("64d44bc8337399ccf9ad7e4d"),
  orderStatus: 'cancelled',
  tax: 0.14,
  subtotal: 50,
  total: 57,
  orderItems: [
    {
      album: new ObjectId("64d2a94c793389a43fc5a8ec"),
      quantity: 2,
      _id: new ObjectId("64e1218480db8b4c9f15ff4c")
    },
    {
      album: new ObjectId("64d2a94c793389a43fc5a8e7"),
      quantity: 5,
      _id: new ObjectId("64e1218480db8b4c9f15ff4d")
    }
  ],
  createdAt: 2023-08-19T20:09:40.331Z,
  updatedAt: 2023-08-19T20:11:32.542Z,
  __v: 0,
  paymentIntentId: 'pi_3NgvSqKnb0YvaiB30BpqeFkt'
}
```

src/controllers/orderController.js
mvstoyan: Removed

src/controllers/orderController.js
mvstoyan: Removed

src/controllers/orderController.js
mvstoyan: fixed

src/models/Order.js
mvstoyan: postman
terminal
bd

src/models/Order.js
mvstoyan: setInterval is working for me, but code doesn't work without it, and even console.log doesn't work. I`m wondering where the mistake is.

src/controllers/orderController.js
```diff
@@ -0,0 +1,42 @@
+const { Order, OrderItem } = require('../models/Order');
```
akosasante: Per the Github Action annotation; we can remove `OrderItem` here since it's not used and is not exported by Order.js, and just have `const { Order } = require('../models/Order');`

src/models/Order.js
akosasante: Hmm, what do you mean the code doesn't work without it? 

Without `setInterval` then the order status won't be automatically updated just based on time. However, the middlewares should still work; whenever a `Order.findById`, `Order.findOne`, or `Order.find` function is called. Are you observering a different outcome? Could you outline what steps you took to trigger the middleware (in the case without setInterval) to help me understand what you mean?

akosasante (approved):
one lint error to fix; but after that feel free to merge if you don't have any further questions!

src/controllers/orderController.js
mvstoyan: Removed

src/models/Order.js
mvstoyan: @akosasante I think I've found another solution. I'll make a commit so you can take a closer look. Using expiresAt in MongoDB instead of setInterval would automatically delete data when it becomes obsolete. This is more reliable and resource-efficient, especially for unpaid orders. Do you think this could be a better option? I can revert to the code with setInterval if this isn't a good idea in this case?

src/models/Order.js
```diff
@@ -60,47 +60,29 @@ const orderSchema = new mongoose.Schema(
     // The paymentIntentId field stores the Stripe state of the purchase
     paymentIntentId: String, // Payment intent ID is a string
     orderItems: [orderItemSchema],
+    expiresAt: {
+      type: Date,
+      default: Date.now() + 60 * 1000, // Initial time for 1 minute
+      index: { expires: '1m' }, // Index for automatic deletion after 1 minute
+    },
```
akosasante: Oh interesting. What does this do / why did you remove the other code? It looks like this will just fully delete any order that's older than 1 minutes, regardless of status; I don't _think_ we want that?

akosasante (changes requested):
I'm not sure if we want to merge the change from the latest commit; could you explain to me what those changes do since I'm not familiar with TTL/expiry in Mongo used in this way.

src/models/Order.js
mvstoyan: @akosasante I've considered using TTL (Time To Live) to automatically remove orders from the database, and I'd like to propose the following approach:

Orders with the status 'pending' will automatically change to 'canceled' after 1 minute using setInterval.

For orders that are 'canceled' and are no longer needed, we will add an 'expiresAt' field. These canceled orders will be automatically removed from the database 2 minutes after their cancellation. This helps reduce the database load and keeps it in good shape.

This approach will allow us to efficiently manage orders and store only relevant data in the database. If this approach works for us, we can adjust the timing, but for testing purposes, I'll keep it at 1 minute and 2 minutes. If this approach isn't suitable, I can remove this code, and the rest of the code will remain unchanged.
expires

akosasante: Ah that's a great idea! 

  
I think my concern is what you have currently doesn't tell Mongo to only delete orders with the `"cancelled"` status as far as I can tell :thinking:

akosasante: Ah I did some reading. So my understanding is that it will only do the automatic deletion if that value is set; is that what you got too? 

So the idea is:

*   by default the order is created with value `pending`
*   if it's still `pending` after **1 minute** then we automatically set the value to `cancelled`
*   if after **2 minutes** past it's creation date, it's still around, it will be deleted automatically

What happens if:

*   order is created (status  = `pending`)
*   order is updated by UI to status = `payment_successful`
*   2 minutes pass. Won't it still get deleted ? I guess in step 2 we have to ensure that we set `expiresAt` to `null` right?

akosasante (approved):
If my understanding of the expiry stuff matches yours, then this is good to go ✅

src/models/Order.js
mvstoyan: expiresAt
@akosasante code sets up a special rule in MongoDB, automatically deletes orders with a 'cancelled' status after 2 minutes.

**orderSchema.index** - starting to create index in the order schema.

**{ expiresAt: 1 }** - this index is for the 'expiresAt' field, and it should be sorted in ascending order.

**{ expireAfterSeconds: 120** - time limit in seconds. If an order has a 'cancelled' status and it's been more than 2 minutes since it was created, MongoDB will delete it from the database.

**partialFilterExpression: { orderStatus: 'cancelled' }** - telling MongoDB that this rule only applies to documents where the 'orderStatus' is 'cancelled'.

So, this code ensures that only cancelled orders get automatically deleted from the database after 2 minutes from their creation. This helps keep the database clean and free up space from unnecessary data.

akosasante (commented):
Nice, looks good 👍

src/models/Order.js
```diff
@@ -62,9 +62,13 @@ const orderSchema = new mongoose.Schema(
...
       type: Date,
-      // Set expiresAt to the future by 2 minutes for orders with the "cancelled" status
+      // Set expiresAt only for orders with the "cancelled" status
       default: function () {
-        return new Date(Date.now() + 2 * 60 * 1000); // 2 minutes
+        if (this.orderStatus === 'cancelled') {
+          return new Date(Date.now() + 2 * 60 * 1000); // 2 minutes
+        }
+        // If the status is not "cancelled", expiresAt is not set
+        return undefined;
```
akosasante: I'm not _100% sure_ if this will get automatically set if you're _updating_ an order (it will for sure happen if you create a `cancelled` order). But we can find out once we get to that step and it seems fine to me to keep it as you have it 👍

PR #29, opened 2023-08-18
Ps90 create order endpoint

mvstoyan:


mvstoyan:
Hi, Akos. I need help understanding -  How can I test a payment with Stripe? Orders go through, but the status doesn't change. How to link payments to orders?

src/controllers/orderController.js
```diff
@@ -0,0 +1,48 @@
...
+
+    // Perform necessary checks and validations
+    if (orderItems.length === 0) {
+      throw new BadRequestError('Unable to create order because the order items are missing.'); 
+    }
+
+    // Create an Order document in the database
+    const order = await Order.create({
+      user: req.user.userId,
+      orderItems: orderItemObjects,
```
akosasante: Does the `Order` model have an `orderItems` field set up in its schema?

src/controllers/orderController.js
```diff
@@ -0,0 +1,48 @@
...
+      subtotal,
+      tax,
+      total,
+      orderStatus: 'pending', // Set initial order status as pending
```
akosasante: It's fine to do this explicitly here; another option would be making the default value of this field if not provided, `pending`

src/controllers/orderController.js
```diff
@@ -0,0 +1,48 @@
...
+const { BadRequestError } = require('../errors');
+
+const createOrder = async (req, res) => {
+  const { orderItems, subtotal, tax } = req.body;
+
+  try {
+    // Create an array of OrderItem objects
+    const orderItemObjects = orderItems.map(item => new OrderItem(item));
+    // Calculate the total price including tax
+    const total = subtotal + (subtotal * tax);
```
akosasante: Hmm, the frontend will have to calculate the total anyway in order to display it to the user in the cart page; I'm not sure that it's necessary to recalculate it here in the backend versus just have the frontend give it to us in the request body.

src/controllers/orderController.js
```diff
@@ -0,0 +1,48 @@
...
+
+    // Perform necessary checks and validations
+    if (orderItems.length === 0) {
```
akosasante: I would suggest using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here just in case the caller either forgot to pass in the field completely or passed in null/undefined: `if (orderItems?.length === 0)`

src/controllers/orderController.js
```diff
@@ -0,0 +1,48 @@
...
+    });
+
+    // Create a payment intent with Stripe
+    const paymentIntent = await stripe.paymentIntents.create({
+      amount: total * 100, // Stripe requires amount in cents
+      currency: 'usd',
+      description: `Order #${order._id}`,
+    });
+
+    // Send the payment intent client secret and order information to the client
```
akosasante: We need to also save the `paymentIntent` to our Order entry in the db to keep it on record (and later we can also get live updates about it from Stripe via webhooks).

`order.paymentId = ..., order.save`

src/controllers/orderController.js
```diff
@@ -0,0 +1,48 @@
...
+    res.status(StatusCodes.CREATED).json({ clientSecret: paymentIntent.client_secret, order });
+  } catch (error) {
+    if (error instanceof BadRequestError) {
+      res.status(error.statusCode).json({ error: error.message });
+    } else {
+      res.status(StatusCodes.BAD_REQUEST).json({ error: error.message });
+    }
+  }
```
akosasante: Any reason to wrap this in a `try/catch` vs just letting the error handling middleware cover it? And could you explain why we want to return bad request 400 for _**all**_ error cases here?

akosasante (commented):
Looking good.

Yes if you're able to create orders, and the order in our db should also store the `paymentItnetnId`. I can see the test payments in the Stripe dashboard: [https://dashboard.stripe.com/test/payments](https://dashboard.stripe.com/test/payments;). You should check and make sure you're able to see them as well. 

The next step is for the frontend to submit the actual payment to stripe and then call the backend to update the order status. The backend ticket for updating the order is: [https://codethedreamteam4-2023.atlassian.net/browse/PS-93](https://codethedreamteam4-2023.atlassian.net/browse/PS-93) (we'll reach out to Raheim to verify that he's not planning on working on that one).

For live-updates I'll have a ticket for implementing webhooks but it'll be outside of the critical path so we'll want to make sure the essential tickets are completed first.

src/controllers/orderController.js
mvstoyan: The code has been fixed in branch PS89, and now the Order model includes the orderItems field as described in the schema.
orderItems

src/controllers/orderController.js
mvstoyan: pending
controller

src/controllers/orderController.js
mvstoyan: removed

src/controllers/orderController.js
mvstoyan: Снимок экрана 2023-08-18 184230

src/controllers/orderController.js
mvstoyan: // Save the paymentIntent to the Order model
  order.paymentIntentId = paymentIntent.id;
  await order.save();

src/controllers/orderController.js
mvstoyan: Wrapping this block in a try/catch is not necessary, I have removed it. Regarding returning a 400 Bad Request status for all errors, regardless of their type, I believed that it offers a clear signal to the client that something is amiss with the request, but I agree, that was a strange idea. I'll adjust it to only return a 400 Bad Request status when the client sends a request with incorrect data.
screen3

src/controllers/orderController.js
akosasante: Is this changed here in this branch or in the PS089 one? I don't see the changes you screenshotted here.

mvstoyan: PS089, there my orderController.js. Sorry)

akosasante (approved):
Looks good! 

If you don't mind, please run `npm generate-api-docs` to generate some basic api docs for the frontend team to refer to as they've started some work on the checkout pages. 

(We can flesh out the api docs more in [https://codethedreamteam4-2023.atlassian.net/browse/PS-112](https://codethedreamteam4-2023.atlassian.net/browse/PS-112) but running this command now at least adds them to the documentation even if some details may be missing)

PR #30, opened 2023-08-18
[PS-72] Review schema created

NatachaKey:


src/routes/reviewRoutes.js
```diff
@@ -0,0 +1,15 @@
...
+
+const {
+  createReview,
+  updateReview,
+} = require('../controllers/reviewController');
+
+router.route('/album/:albumId').post(authenticateUser, createReview);
+
+//? for Akos- does it matter which type of request we add here- put or patch? because I used put but it didn't overwright the id of the review
+router.route('/:id').put(authenticateUser, updateReview);
```
akosasante: `PUT` is _usually_ when the caller (frontend) will pass the whole model and the server will replace the entry in the db. So even if they just want to upate the rating; they'd still pass in the whole object. So let's say they pass in: 

```json
{ 
  "review": {
    "id": ...,
    "rating": <new rating>,
    "title": ...,
    "comment": ...,
    "user": ...,
    "album": ...
  }
}
```

`PUT` doesn't necessarily mean that the id will change or get overwritten. It is usually used to represent: "update the record at this id with all this content, or if it doesn't exist create a new one". Aka "[upsert](https://en.wiktionary.org/wiki/upsert#:~:text=upsert%20(plural%20upserts),updates%20them%20if%20they%20do.)"

`PATCH` is _usually_ when the caller will pass just the partial bits of the model that they want updated; eg, if they:

```
{
  "review": {
    "rating": <new rating>
  }
}
```

Now these are not hard and fast rules, they are just guidance/common practice: [https://rapidapi.com/blog/put-vs-patch/](https://rapidapi.com/blog/put-vs-patch/)

In this case, most of our other update routes are using `PATCH` so I think it will be fine to use patch here to remain consistent. I think in the ticket I wrote `PUT` with the thinking that likely the frontend will pass in the whole object anyway for simplicity. We can still use patch evne if that's the case.

src/controllers/reviewController.js
```diff
@@ -0,0 +1,88 @@
...
+
+//create review
+const createReview = async (req, res) => {
+  //( frontend sends an album property in the req.body) ! we MUST PROVIDE which album we are requesting- "album": "64d2a94c793389a43fc5a8ec",
+  const { album: albumId } = req.body; //check /look for the album in req.body and assign it to albumId // in postman in REQUEST { album: albumId } : {"album": "647e247e69c32ece45e23978"}
+  //check if album exists in db
+  const isValidAlbum = await Album.findOne({ _id: albumId });
+  if (!isValidAlbum) {
+    throw new CustomError.NotFoundError(`No album with id : ${albumId}`);
+  }
```
akosasante: Since we're not actually using all the fields of the returned value of `Album.findOne` this is a little unoptimized. We can tell Mongo that we just want to check if the record exists, but we don't need it to return _all_ the fields. 

[https://mongoosejs.com/docs/api/model.html#Model.exists()](https://mongoosejs.com/docs/api/model.html#Model.exists()) \<-- will return `{_id: <ObjectId> }`  if found, or `null` otherwise. It takes the same arguments as `findOne`

```javascript
const isValidAlbum = await Album.exists({ _id: albumId });
if (!isValidAlbum) {...}
```

src/controllers/reviewController.js
```diff
@@ -0,0 +1,88 @@
...
+  //check if there is already review for this album from this user
+  const alreadySubmitted = await Review.findOne({
+    album: albumId,
+    user: req.user.userId,
+  });
+
+  if (alreadySubmitted) {
+    throw new CustomError.BadRequestError(
```
akosasante: `BadRequest` 400 is fine here. But we can be even more precise/follow RESTful practices by creating a new CustomError for the status `409`. 409 is used when there's a "Conflict" : https://http.dev/409

src/controllers/reviewController.js
```diff
@@ -0,0 +1,88 @@
+const Review = require('../models/Review');
+const Album = require('../models/Album');
+const { StatusCodes } = require('http-status-codes');
+const CustomError = require('../errors');
+
+//create review
+const createReview = async (req, res) => {
+  //( frontend sends an album property in the req.body) ! we MUST PROVIDE which album we are requesting- "album": "64d2a94c793389a43fc5a8ec",
+  const { album: albumId } = req.body; //check /look for the album in req.body and assign it to albumId // in postman in REQUEST { album: albumId } : {"album": "647e247e69c32ece45e23978"}
```
akosasante: Will we be getting this from the body or from the path params? Since this request is happening on `/api/v1/reviews/album/:albumId` ?

src/controllers/reviewController.js
```diff
@@ -0,0 +1,88 @@
...
+
+// Check if the requesting user is the author of the review
+  //option 1
+  if (review.user.toString() !== req.user.userId) {
+    throw new CustomError.UnauthorizedError(
+      'You are not authorized to update this review'
+    );
+  }
+  //option 2
+  // checkPermissions(req.user, review.user);
```
akosasante: This works as well; only difference is that `checkPermissions` allows the action if it's an admin. I don't know if that would really make sense in this case?

src/controllers/reviewController.js
```diff
@@ -0,0 +1,88 @@
...
+  //option B:
+  //   review.rating = req.body.hasOwnProperty('rating')
+  //     ? req.body.rating
+  //     : review.rating;
+  //   review.title = req.body.hasOwnProperty('title')
+  //     ? req.body.title
+  //     : review.title;
+  //   review.comment = req.body.hasOwnProperty('comment')
+  //     ? req.body.comment
+  //     : review.comment;
```
akosasante, line 77: Yeah if we make it clear in the API docs (upcoming ticket) that these are required fields; then we can probably stick with lines 65-68 because then frontend shouldn't be passing in empty/missing fields. 

Also in this case, since the model schema defines rating, title, and comment as required fields; the saving should fail anyway if the frontend did accidentally pass in missing/empty/null fields.

akosasante (commented):
Leaving some small suggestions; but no changes needed. LGTM! just check if you need to merge in main.

akosasante:
And to clarify this also covers [PS-73](https://codethedreamteam4-2023.atlassian.net/browse/PS-73) and [PS-74](https://codethedreamteam4-2023.atlassian.net/browse/PS-74) right?

src/routes/reviewRoutes.js
NatachaKey: @akosasante so with put request we should pass in the req.body all the required fields and it will throw an error if we don't, right? and with patch we just pass partial fields we need to update?
UPD: just read the article, I think that's correct

NatachaKey: put overwrights everything so if the fields are not required and we pass less properties than it had initially- it will fully replace the item in a db and we'll have only properties we passed in our put request.

NatachaKey: does 'upsert' means smth in between 'insert' and 'update'? @akosasante

src/controllers/reviewController.js
NatachaKey: @akosasante oh I see, and findOne returns the first document (=all the fields) https://mongoosejs.com/docs/api/query.html#Query.prototype.findOne():~:text=The%20result%20of%20the%20query%20is%20a%20single%20document%2C%20or%20null%20if%20no%20document%20was%20found.

src/controllers/reviewController.js
NatachaKey: @akosasante thank you so much!! added this error 409

src/controllers/reviewController.js
NatachaKey: @akosasante I don't understand quiet well this part- shouldn't it be just `router.route('/).post(authenticateUser, createReview); `instead of `router.route('/album/:albumId').post(authenticateUser, createReview);` ? because when we /api/v1/reviews/album/:albumId we don't actually **refer** to a specific album with this specific albumId ?

src/controllers/reviewController.js
NatachaKey: @akosasante oh I thought that checkPermissions checks both things- 1) if the user is an admin and 2)if the user who makes the request (req.user) is the user who left the review (review.user), am I wrong?  
image

NatachaKey:
> And to clarify this also covers [PS-73](https://codethedreamteam4-2023.atlassian.net/browse/PS-73) and [PS-74](https://codethedreamteam4-2023.atlassian.net/browse/PS-74) right?

@akosasante Yes, that's it //also added getAllReviews and getSingleReview routes and controllers

src/routes/reviewRoutes.js
akosasante: > @akosasante so with put request we should pass in the req.body all the required fields and it will throw an error if we don't, right? and with patch we just pass partial fields we need to update? UPD: just read the article, I think that's correct

Yes that's correct! 👍 

'Upsert' == update + insert. It means if the entry is found in the db, then update it; and if it's not, then insert one using that data/request body.

With a regular "update", if the entry isn't found in the db, you would just return a "not found" error and not try to now create that entry.

akosasante: > put overwrights everything so if the fields are not required and we pass less properties than it had initially- it will fully replace the item in a db and we'll have only properties we passed in our put request.

Exactly! The id of the object would still be the same; but all the fields of that object will be replaced with what was given in the request body.

src/controllers/reviewController.js
akosasante: Yep exactly!

src/controllers/reviewController.js
akosasante: Yep that's correct! I was pointing out that in Option 1, on lines46-48 we are _only_ checking if the requesting user is the author of the review. We're not checking if they're an admin. So with Option 1, an admin _cannot_ update another user's review (which I think makes sense!).

With Option 2, `checkPermissions` will return and not throw an error, if the requesting user matches the author of the review OR if the requesting user is an adming. With Option 2 an admin _can_ update another user's review.

src/controllers/reviewController.js
akosasante: I think it was fine to have `router.route('/album/:albumId').post(authenticateUser, createReview);` I was wondering in that case if we should be using that `:albumId` from the path in our controller.

So instead of looking in the body for `const { album: albumId } = req.body;`, we'd have `const { albumId } = req.params;`. Since the client is already telling us which album they want to create a review for based on the params and they don't need to repeat that info in the body. We would have to then make sure we do add it to `req.body` before creating the review; similar to what you're doing for the user id here: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/e76b227a58c8dc3ad5a0c65cd07c05c1a4385c83/src/controllers/reviewController.js#L27](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/e76b227a58c8dc3ad5a0c65cd07c05c1a4385c83/src/controllers/reviewController.js#L27)

The change you made to have the create endpoint be just on the `api/v1/reviews` route is fine as well though! In that case we have to get that album info from the body. :+1:

src/controllers/reviewController.js
```diff
@@ -0,0 +1,119 @@
...
+   //review+ about which album 
+   path: 'album', //we pass what we want to reference //line 26 in Review.js
+   select: 'artistName albumName image price', //and what properties we want to get from Album model (Album.js)
+
+    //and to populate user data: review+user who wrote it
+    // path: 'user', //we pass what we want to reference //in Review.js
+    // select: 'email name username', //and what properties we want to get from album model
+  });
+  res.status(StatusCodes.OK).json({ reviews, count: reviews.length });
+};
```
akosasante: We're fine to leave this endpoint+controller, since I can imagine there can be cases where the frontend want sto get _all_ the reviews. But for [PS-73,](https://codethedreamteam4-2023.atlassian.net/browse/PS-73) we specifically want an endpoint that gives the client just the reviews for a particular product/album. That way we reduce the size of the response we're returning and the client doesn't have to filter out the results list themselves. Filtering at the db-level is generally faster than filtering in client-side javascript code.

akosasante (commented):
Looking good;  just looking for an endpoint to fulfill the criteria in PS-73

src/routes/reviewRoutes.js
NatachaKey: @akosasante thank you so much! Love your explanations!!!

src/controllers/reviewController.js
NatachaKey: @akosasante should I add review property yo Album.js and also add virtuals to retrieve reviews for a single album?

akosasante: That would work I think. I believe a simpler approach for now would just be to use the `Review` model since it already has an `album` property. So we can do something like `Review.find({ album: albumId })` to fetch all reviews for a particular album and return those review objects. 

Your approach does have the benefit of then the client only has to make a single request on the album details page for that information.

src/controllers/reviewController.js
```diff
@@ -6,7 +6,11 @@ const CustomError = require('../errors');
 //create review
 const createReview = async (req, res) => {
   //( frontend sends an album property in the req.body) ! we MUST PROVIDE which album we are requesting- "album": "64d2a94c793389a43fc5a8ec",
-  const { album: albumId } = req.body; //check /look for the album in req.body and assign it to albumId
+  const { albumId } = req.params; //check /look for the albumId in req.params 
+  
+  //add it to req.body before creating the review
+  req.body.album = req.album.albumId;
```
akosasante:
```suggestion
req.body.album = albumId
```

src/controllers/reviewController.js
NatachaKey: @akosasante did it! just commited to GitHub, so I think this was https://codethedreamteam4-2023.atlassian.net/browse/PS-76

deleteCollectionMongo.js
akosasante, line 1: Any particular reason this was needed?

src/controllers/reviewController.js
akosasante: Oh yep you're right! I think I read it quickly on got mixed up between "Create for a specific album" and "Get for a specific album" :white\_check\_mark:

src/routes/reviewRoutes.js
```diff
@@ -7,11 +7,16 @@ const {
...
 } = require('../controllers/reviewController');
 
 router.route('/').get(getAllReviews);
 
+
 router.route('/album/:albumId').post(authenticateUser, createReview);
 
 router.route('/:id').patch(authenticateUser, updateReview).get(getSingleReview)
+
+router.route('/album/:albumId').get(getAllReviewsForThisProduct)
```
akosasante: This can just be tacked onto line 16:

`router.route('/album/:albumId').post(authenticateUser, createReview).get(getAllReviewsForThisProduct)`

akosasante (approved):
Small comments, but looks good to me to merge!

deleteCollectionMongo.js
NatachaKey: @akosasante oh I just needed to test how reviews controllers work and I thought it might be helpful for others who might need to do the same thing with orders/users/etc.- but maybe it's not appropriate to leave it in code- in order to maintain it clean?

akosasante: It's fine I was just curious; feel free to leave it in since it's a separate script not part of the running app :+1:

PR #31, opened 2023-08-24
Admin Page EJS

mvstoyan:


src/views/adminPage.ejs
```diff
@@ -0,0 +1,37 @@
...
+        <th>Album Title</th>
+        <th>Artist</th>
+        <th>Price</th>
+        <th>Select</th>
+      </tr>
+    </thead>
+    <tbody>
+      <% function formatPrice(price) {
+          return price.toLocaleString('en-US', { style: 'currency', currency: 'USD' });
+      } %>
```
akosasante: Just a suggestion, but generally, I would define functions at the top of the EJS file, so it's easy to find them all and they're not interspersed with the html template code.

src/public/style.css
akosasante, line 1: :star\_struck:

src/controllers/adminController.js
```diff
@@ -0,0 +1,11 @@
+const mongoose = require('mongoose');
```
akosasante: Github actions lint step is pointing out that this import is probably not needed

src/controllers/adminController.js
```diff
@@ -0,0 +1,11 @@
...
+const Album = require('../models/Album');
+
+const adminHomePage = async (req, res) => {
+  const allAlbums = await Album.find({});
```
akosasante: Since the first task that we're trying to enable for the admin right now is editing prices for albums with a price of $0.00; maybe it'll be helpful to have those at the top of the table? So we could do an `order by` here to return the fetched albums in ascending order of price?

akosasante (approved):
A couple small suggestions for tweaking this but looks great!

akosasante:
@mvstoyan After you merge; could you please post the updated .env values (for the admin creds and stripe key) in the team channel and replace the old pinned post ? Thank you!

src/controllers/adminController.js
mvstoyan: Removed

src/controllers/adminController.js
mvstoyan: This is good idea, thank you very much!

mvstoyan: screen

PR #33, opened 2023-08-27
bulk-edit-prices

mvstoyan:
When I attempting to send a PATCH request to /admin/api/albums, an error occurs. The server log indicates HTTP status codes 401 and 404. Please can you assist me in understanding why this error is happening when trying to execute the PATCH request and how to resolve it.


src/routes/adminRoutes.js
```diff
@@ -2,8 +2,10 @@ const express = require('express');
 const router = express.Router();
 const expressBasicAuth = require('express-basic-auth');
 const adminController = require('../controllers/adminController');
-
-router.get('/', adminController.adminHomePage);
```
akosasante, line 6: Did this get moved somewhere else? I'm not able to load the admin page without this (I get a "Route does not exist" page)

src/routes/adminRoutes.js
```diff
@@ -13,4 +15,9 @@ router.use(
   })
 );
 
+
+router
+  .route('/admin/api/albums')
```
akosasante: Because we "mount" this router in `expressServer.js` to the `/admin` route: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/PS97-bulk-edit-prices/src/expressServer.js#L122

That means that any routes we define in this file will be _appended_ to `<SERVER_URL>/admin`

So what you have here will be trying to handle calls from `/admin/admin/api/albums`.

But the call in the template is trying to call `/admin/api/album`. Express thinks that path doesn't exist and then returns a 404

```suggestion
  .route('/api/albums')
```

src/routes/adminRoutes.js
```diff
@@ -13,4 +15,9 @@ router.use(
...
+  .patch(authenticateUser, authorizePermissions('admin'), adminController.updatePriceOfAlbums);
```
akosasante: Since this admin route is already using the basic auth middleware; we don't need to also add the `authenticateUser` middleware. In addition, both `authenticateUser` and `authorizePermissions` are currently written in a way that's more specific to cookie-based authorization. For now it's sufficient to just rely on the basic auth for these special admin server-rendered pages.

```suggestion
  .patch(adminController.updatePriceOfAlbums);
```

src/views/adminPage.ejs
```diff
@@ -29,9 +34,9 @@
       <% }); %>
     </tbody>
   </table>
-
-  <script>
-    // JS
+  <div id="successMessage" style="color: green;"></div>
+  <div id="errorMessage" style="color: red;"></div>
+  <script src="/admin.js">
```
akosasante: Since this table can currently be quite long; may be better to put these error/success messages above the table instead of at the bottom? They're easy to miss right now.

src/public/admin.js
```diff
@@ -0,0 +1,50 @@
...
+    try {
+      const response = await fetch('/admin/api/albums', {
+        method: 'PATCH',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({
+          albumIds: selectedAlbums,
+          newPrice: newPrice,
+        }),
```
akosasante: Currently, you're having the `updatePriceOfAlbums` albums from the album controller, handle this request. But it's expecting the request body shape to be different than what you're sending here.

Screenshot 2023-08-26 at 10 30 37 PM

Screenshot 2023-08-26 at 10 30 40 PM

You could either:
- have some code in the admin controller that transforms the req.body before calling `updatePriceOfAlbums(req, res)`
- write code in admin controller that just completely handles this request payload on its own
- update the code here in `admin.js` to send the request in the shape expected by `updatePriceOfAlbums` (if that endpoint is not yet documented, I believe Natalia implemented it so you can double check with her)

src/routes/adminRoutes.js
mvstoyan: Sorry, accidentally.

src/routes/adminRoutes.js
mvstoyan: Thank you very much

src/public/admin.js
mvstoyan: fixed. Thank you for your help❤️

akosasante (approved):
Great job!

PR #34, opened 2023-08-27
modifyed swagger user endpoints

NatachaKey:


src/expressServer.js
akosasante, line 1: :star:

src/controllers/userController.js
```diff
@@ -8,6 +8,18 @@ const {
 } = require('../utils');
 
 const getAllUsers = async (req, res) => {
+  /*
+     #swagger.summary = 'Fetch all registered users in a database whose role is user (exclude admins)'
+     #swagger.parameters['role'] = {
+        description: 'Role of the user to fetch',
+     }
```
akosasante: The endpoint doesn't currently take any parameters as far as I can see; so I think we can remove this. That is, the user can't pass in `?role=user` query params or any params in the path right? The endpoint just _always_ only return non-admin users.

src/controllers/userController.js
```diff
@@ -8,6 +8,18 @@ const {
...
+     #swagger.responses[200] = {
+				description: 'Users successfully fetched.',
+        schema: [{ $ref: '#/definitions/PasswordlessUser'' }]
```
akosasante: You have an extra quotation mark here that's causing `swagger-autogen` to print out a warning about this being unparseable when I run the generate api docs command

```suggestion
        schema: [{ $ref: '#/definitions/PasswordlessUser' }]
```

akosasante: Actually, the endpoint doesn't return an array directly. When I use postman to test this endpoint I get an object with the key `users` and the value of that is the array of users:

```
{ 
  "users": [ array of users here ]
}
```

So we should make that clear to the reader of these api docs:

```suggestion
        schema: { users: [{ $ref: '#/definitions/PasswordlessUser' }] }
```

src/controllers/userController.js
```diff
@@ -26,7 +38,7 @@ const getSingleUser = async (req, res) => {
 				schema: { $ref: '#/definitions/PasswordlessUser' }
 		 }
 		 #swagger.responses[404] = { description: 'No user with id found.' }
-		 #swagger.responses[403] = { description: 'Requester forbidden to fetch this user.' }
```
akosasante, line 29: Why was this removed? Isn't it possible to get this error if `checkPermissions` finds that an inappropriate user is making this request?

src/controllers/userController.js
```diff
@@ -26,7 +38,7 @@ const getSingleUser = async (req, res) => {
 				schema: { $ref: '#/definitions/PasswordlessUser' }
```
akosasante: I missed in this example one that this endpoint returns `{ user: < passwordless user> }` so this should be updated to:

```suggestion
				schema: { user: { $ref: '#/definitions/PasswordlessUser' } }
```

src/controllers/userController.js
```diff
@@ -39,6 +51,17 @@ const getSingleUser = async (req, res) => {
 };
 
 const showCurrentUser = async (req, res) => {
+  /*
+     #swagger.summary = 'Fetch a user by id'
```
akosasante, line 37:
```suggestion
     #swagger.summary = 'Returns information about the requesting user based on cookie session'
```

src/controllers/userController.js
```diff
@@ -39,6 +51,17 @@ const getSingleUser = async (req, res) => {
...
+     #swagger.parameters['id'] = {
+        description: 'Mongo ObjectID of the user to fetch',
+     }
```
akosasante: This endpoint does not take any query or path parameters; it just uses the cookie to find the requesting user's id.

src/controllers/userController.js
```diff
@@ -50,6 +73,17 @@ const showCurrentUser = async (req, res) => {
 
 // Update the information of the current user
 const updateCurrentUser = async (req, res) => {
+  /*
+     #swagger.summary = 'Fetch a user by id, update their data and return a new token user.'
```
akosasante:
```suggestion
     #swagger.summary = 'Fetch a user by id, update their data and return a new user.'
```

We just return a regular full (except password) user; not the tokenized one that we put in the cookie

src/controllers/userController.js
```diff
@@ -50,6 +73,17 @@ const showCurrentUser = async (req, res) => {
...
+     #swagger.parameters['id'] = {
+        description: 'Mongo ObjectID of the user to fetch',
+     }
```
akosasante: This endpoint does not take any query or path parameters; it just uses the cookie to find the requesting user's id.

src/controllers/userController.js
```diff
@@ -67,6 +101,19 @@ const updateCurrentUser = async (req, res) => {
 };
 
 const updateUserPassword = async (req, res) => {
+  /*
+     #swagger.summary = 'Fetch a user by id, update their password and save it to the database.'
+     #swagger.parameters['id'] = {
+        description: 'Mongo ObjectID of the user to fetch',
+     }
```
akosasante: This endpoint does not take any query or path parameters; it just uses the cookie to find the requesting user's id.

src/controllers/userController.js
```diff
@@ -67,6 +101,19 @@ const updateCurrentUser = async (req, res) => {
 };
 
 const updateUserPassword = async (req, res) => {
+  /*
+     #swagger.summary = 'Fetch a user by id, update their password and save it to the database.'
```
akosasante:
```suggestion
     #swagger.summary = 'Fetch the current requesting user by id, update their password and save it to the database.'
```

src/controllers/userController.js
```diff
@@ -67,6 +101,19 @@ const updateCurrentUser = async (req, res) => {
...
+     #swagger.parameters['id'] = {
+        description: 'Mongo ObjectID of the user to fetch',
+     }
+     #swagger.responses[200] = {
+				description: 'User successfully fetched and password updated.',
+				schema: { $ref: '#/definitions/PasswordlessUser' }
+		 }
+		 #swagger.responses[400] = { description: 'Error. Need to provide both name and email values.' }
```
akosasante, line 90:
```suggestion
		 #swagger.responses[400] = { description: 'Error. Need to provide both new and old password values' }
```

src/controllers/userController.js
```diff
@@ -88,14 +135,40 @@ const updateUserPassword = async (req, res) => {
...
   res.status(StatusCodes.OK).json({ message: 'User deleted successfully' });
 };
 
 //Fetching a user from the database, including all the albums they've purchased
 const getCurrentUserWithPurchasedAlbums = async (req, res) => {
+  /*
+     #swagger.summary = 'Fetch a user by id and all albums they purchased.'
+     #swagger.parameters['id'] = {
+        description: 'Mongo ObjectID of the user to fetch',
+     }
```
akosasante: This endpoint does not take any query or path parameters; it just uses the cookie to find the requesting user's id.

src/controllers/userController.js
NatachaKey: @akosasante got it!! thank you

NatachaKey: does parameters mean here what we have in req.params? @akosasante

src/controllers/userController.js
NatachaKey: @akosasante I am trying to test this getAllUsers route and get in postman
Captura de pantalla 2023-08-29 a las 0 10 40
//maybe i am testing the wrong route- where do you get ```
{ 
  "users": [ array of users here ]
}
``` ?

File: src/controllers/userController.js (hunk 3 above)
Comment (NatachaKey) [reply src/controllers/userController.js:29]: @akosasante  I forgot that checkPermissions also can throw an error, will add it again sorry

File: src/controllers/userController.js (hunk 1 above)
Comment (akosasante) [reply src/controllers/userController.js]: Yes, in these, `parameters` means both path/route parameters (eg, for an express route like `/users/:userId` , userId would be a path parameter; which like you said would be used in the controller with `req.params`) and also refers to query parameters (eg for an endpoint that allows something like `/albums/filter?limit=2&order=asc`, limit and order would be query parameters, and would be used in the controller with `req.query`). 

In swagger-autogen syntax, we would document path parameters like this: https://www.npmjs.com/package/swagger-autogen#parameters

```
#swagger.parameters['userId'] = {
  in: 'path',
  description: "The user id that you want to fetch by",
  required: true,
  type: "string"
}
```

```
#swagger.parameters['limit'] = {
  in: 'query', // actually, by default it assumes query, so you could leave this out, but it's clearer to always include it
  type: 'number',
  required: false,
  minimum: 1
}
```

File: src/controllers/userController.js (hunk 2 above)
Comment (akosasante) [reply src/controllers/userController.js]: To clarify, I'm referring to the response body. For example, in your screenshot the JSON returned is an object, with a key of `"users"` and then the list of users. Which is to say, it returns an object and not just directly the list; otherwise we would see in the response body:

```
[
  { "_id": blah, name: blah, etc},
  { "_id": blah2, name: blah2, etc}
]
```

But instead we get back what you have there which is:

```
{
  "users": [
    { "_id": blah, name: blah, etc},
    { "_id": blah2, name: blah2, etc},
  ]
}
```

We need our documentation to describe that as well so that on the frontend they'll know that they have to do something like `response.data.users` instead of `response.data`

File: src/controllers/userController.js [hunk 13]
```diff
@@ -39,6 +48,14 @@ const getSingleUser = async (req, res) => {
 };
 
 const showCurrentUser = async (req, res) => {
+  /*
+     #swagger.summary = 'Returns information about the requesting user based on cookie session
```
akosasante: missing quotation mark at the end is making swagger-autogen parse this weirdly; it's including different stuff than what you have here, but adding the quote mark should fix it (downside of having this syntax in comments is that ide/linter won't catch things like this for us sadly :disappointed: )

```suggestion
     #swagger.summary = 'Returns information about the requesting user based on cookie session'
```

src/controllers/userController.js
```diff
@@ -50,6 +67,14 @@ const showCurrentUser = async (req, res) => {
 
 // Update the information of the current user
 const updateCurrentUser = async (req, res) => {
+  /*
+     #swagger.summary = 'Fetch a user by id, update their data and return a new user.'
```
akosasante:
```suggestion
     #swagger.summary = 'Fetch the requesting user, update their data and return the updated user.'
```

src/controllers/userController.js
NatachaKey: @akosasante  thanks so much!!!!

src/controllers/userController.js
NatachaKey: @akosasante got it! thank you!!

src/controllers/userController.js
```diff
@@ -8,6 +8,15 @@ const {
 } = require('../utils');
 
 const getAllUsers = async (req, res) => {
+  /*
+     #swagger.summary = 'Fetch all registered users in a database whose role is user (exclude admins)'
+     #swagger.responses[200] = {
+				description: 'Users successfully fetched.',
+        schema: [{ $ref: '#/definitions/PasswordlessUser' }]
```
akosasante:
```suggestion
        schema: { users: [{ $ref: '#/definitions/PasswordlessUser' }] }
```

src/controllers/userController.js
```diff
@@ -39,6 +48,14 @@ const getSingleUser = async (req, res) => {
 };
 
 const showCurrentUser = async (req, res) => {
+  /*
+     #swagger.summary = 'Returns information about the requesting user based on cookie session
+     #swagger.responses[200] = {
+				description: 'User successfully fetched.',
+				schema: { $ref: '#/definitions/PasswordlessUser' }
```
akosasante:
```suggestion
				schema: {user: { $ref: '#/definitions/PasswordlessUser' }}
```

src/controllers/userController.js
```diff
@@ -50,6 +67,14 @@ const showCurrentUser = async (req, res) => {
 
 // Update the information of the current user
 const updateCurrentUser = async (req, res) => {
+  /*
+     #swagger.summary = 'Fetch a user by id, update their data and return a new user.'
+     #swagger.responses[200] = {
+				description: 'User successfully fetched and updated.',
+				schema: { $ref: '#/definitions/PasswordlessUser' }
```
akosasante:
```suggestion
				schema: {user: { $ref: '#/definitions/PasswordlessUser' }}
```

src/controllers/userController.js
```diff
@@ -67,6 +92,16 @@ const updateCurrentUser = async (req, res) => {
 };
 
 const updateUserPassword = async (req, res) => {
+  /*
+     #swagger.summary = 'Fetch the current requesting user by id, update their password and save it to the database.'
+     #swagger.responses[200] = {
+				description: 'User successfully fetched and password updated.',
+				schema: { $ref: '#/definitions/PasswordlessUser' }
```
akosasante: This endpoint just returns a message object (`res.status(StatusCodes.`_`OK`_`).json({ msg: 'Success! Password Updated.' });`

```suggestion
				type: 'object'
				properties: {
				  msg: { type: 'string' }
				}
```

src/controllers/userController.js
```diff
@@ -88,14 +123,37 @@ const updateUserPassword = async (req, res) => {
...
 };
 
 //Fetching a user from the database, including all the albums they've purchased
 const getCurrentUserWithPurchasedAlbums = async (req, res) => {
+  /*
+     #swagger.summary = 'Fetch a user by id and all albums they purchased.'
+     #swagger.parameters['id'] = {
+        description: 'Mongo ObjectID of the user to fetch',
+     }
+     #swagger.responses[200] = {
```
akosasante, line 136: We could include a schema here; it would require making a new object in `swagger.js` to represent a user with their albums (similar to how PasswordlessUser is defined there.)

akosasante (commented):
I noticed a weird bug (not in your code; but in the swagger-autogen library itself) that seems to cause the request body to not be correctly captured in some cases for the write endpoints (PUT/PATCH/POST). So far the only workaround i figured out was putting the swagger comments at the _end_ of the function body instead of the beginning as instructed. I'll look into this further but just wanted to let you know in case you run into it. What I observed was the request body for an endpoint like `updateUserPassword` was showing "descirption: any, schema: any" instead of the expected `oldPassword: "any", newPassword: "any"`. 

example of moving the comments to the bottom:

```javascript
const updateUserPassword = async (req, res) => {
  // Function to update the current user's password
  const { oldPassword, newPassword } = req.body;
  if (!oldPassword || !newPassword) {
    // Check if oldPassword and newPassword are provided
    throw new CustomError.BadRequestError('Please provide both values');
  }
  const user = await User.findOne({ _id: req.user.userId }); // Find the user in the database based on the current user's ID
  // Compare the old password provided with the user's current password
  const isPasswordCorrect = await user.comparePassword(oldPassword);
  if (!isPasswordCorrect) {
    // Throw an UnauthenticatedError if the old password is incorrect
    throw new CustomError.UnauthenticatedError('Invalid Credentials');
  }
  user.password = newPassword; // Update the user's password

  await user.save(); // Save the updated user to the database
  // Send a JSON response with the status code 200 OK and a success message
  res.status(StatusCodes.OK).json({ msg: 'Success! Password Updated.' });

  /*
	 #swagger.summary = 'Fetch the current requesting user by id, update their password and save it to the database.'
	 #swagger.responses[200] = {
			description: 'User successfully fetched and password updated.',
			type: 'object',
			properties: {
        msg: { type: 'string' }
			}
	 }
	 #swagger.responses[400] = { description: 'Error. Need to provide both new and old password values' }
	 #swagger.responses[401] = { description: 'Error. Invalid credentials.' }
*/
};
```

src/controllers/userController.js
NatachaKey: @akosasante fixed

src/controllers/userController.js
NatachaKey: @akosasante just tried to do that but not able to get user with albums-could you please take a look

src/controllers/userController.js
```diff
@@ -13,29 +13,37 @@ const getAllUsers = async (req, res) => {
   // Find all users with the role 'user' in the database and exclude the 'password' field
   const users = await User.find({ role: 'user' }).select('-password');
   res.status(StatusCodes.OK).json({ users }); // Send a JSON response with the status code 200 OK and the users
+  /*
+     #swagger.summary = 'Fetch all registered users in a database whose role is user (exclude admins)'
```
akosasante, line 17: Looks great! We can also add a little blurb in the description (when they expand that endpoint in the docs) to indicate that this endpoint can only be called by admins. Swagger uses some flavour of [Markdown](https://www.markdownguide.org/basic-syntax/), so we can use `**` to bold text:

```suggestion
     #swagger.summary = 'Fetch all registered users in a database whose role is user (exclude admins)'
     #swagger.description = '**ROLE REQUIRED:** admin'
```

src/controllers/userController.js
```diff
@@ -46,6 +54,14 @@ const showCurrentUser = async (req, res) => {
   }
 
   res.status(StatusCodes.OK).json({ user });
+  /*
+     #swagger.summary = 'Returns information about the requesting user based on cookie session'
+     #swagger.responses[200] = {
+				description: 'User successfully fetched.',
+				schema: { user: [{ $ref: '#/definitions/PasswordlessUser' }] }
```
akosasante: this should be a single user not an array:

```suggestion
				schema: { user: { $ref: '#/definitions/PasswordlessUser' } }
```

src/controllers/userController.js
```diff
@@ -64,6 +80,14 @@ const updateCurrentUser = async (req, res) => {
   attachCookiesToResponse({ res, user: tokenUser });
   // Send a JSON response with the status code 200 OK and the updated user
   res.status(StatusCodes.OK).json({ user });
+  /*
+     #swagger.summary = 'Fetch the requesting user, update their data and return the updated user.'
+     #swagger.responses[200] = {
+				description: 'User successfully fetched and updated.',
+				schema: { user: [{ $ref: '#/definitions/PasswordlessUser' }] }
```
akosasante: This should document returning a single user object not an array (under the `user` key)

```suggestion
				schema: { user: { $ref: '#/definitions/PasswordlessUser' } }
```

src/controllers/userController.js
```diff
@@ -85,13 +109,37 @@ const updateUserPassword = async (req, res) => {
...
-};
 
+  /*
+     #swagger.summary = 'Fetch the current requesting user by id, update their password and save it to the database.'
+     #swagger.responses[200] = {
+				description: 'User successfully fetched and password updated.',
+				type: 'object'
+				properties: {
+				  msg: { type: 'string' }
+				}
```
akosasante: my earlier advice here was incorrect; this should just be a `schema` object with an example of the expected response (which swagger-autogen will analyze to determine the types of the fields):

```suggestion
				schema: { msg: 'Success! Password Updated.' }
```

src/controllers/userController.js
```diff
@@ -85,13 +109,37 @@ const updateUserPassword = async (req, res) => {
...
   const userId = req.params.id;
+  if (!userId) {
+    // Throw a NotFoundError if the user is not found
+    throw new CustomError.NotFoundError(`No user with id: ${req.params.id}`);
+  }
   // Find the user by ID and delete
   await User.findByIdAndDelete(userId);
   res.status(StatusCodes.OK).json({ message: 'User deleted successfully' });
+  /*
+     #swagger.summary = 'Fetch a user by id and delete user.'
```
akosasante, line 134: Like with the getAllUsers endpoint, we can add to the description here to let frontend know that only admin users can hit this endpoint:

```suggestion
     #swagger.summary = 'Fetch a user by id and delete user.'
     #swagger.description = '**ROLE REQUIRED:** admin'
```

swagger.js
```diff
@@ -16,10 +16,10 @@ const doc = {
     // Here we can define objects to use in our swagger documentation
     // These can be used as examples for request bodies, response bodies, etc
     PasswordlessUser: {
-      name: 'John Doe',
-      username: 'johndoe',
-      email: 'john@example.com',
-      role: { '@enum': ['user', 'admin'] },
+      $name: 'John Doe',
```
akosasante, line 20: Let's add the id field here as well just to fully document the expected response:

```suggestion
      $id: '64d6ca92a15d2e18ab96a2a3',
      $name: 'John Doe',
```

swagger.js
```diff
@@ -31,6 +31,34 @@ const doc = {
         },
       },
     },
+    UserWithAlbums: {
+
```
akosasante:
```suggestion
      $id: '64d6ca92a15d2e18ab96a2a3',
```

src/controllers/userController.js
```diff
@@ -111,6 +159,18 @@ const getCurrentUserWithPurchasedAlbums = async (req, res) => {
     user: userWithAlbums,
     purchasedAlbumCount: userWithAlbums.purchasedAlbums.length,
   });
+  /*
+     #swagger.summary = 'Fetch a user by id and all albums they purchased.'
+     #swagger.parameters['id'] = {
+        description: 'Mongo ObjectID of the user to fetch',
+        schema:  { $ref: '#/definitions/UserWithAlbums' }
+     }
```
akosasante: This endpoint doesn't take the id in the path/body/query paramaters, it just grabs the user id from the authenticated cookie. So we can remove these lines

swagger.js
```diff
@@ -31,6 +31,34 @@ const doc = {
...
+        $albumName: '...Baby one more time',
+        price: 20,
+        image: '/uploads/example.jpg',
+        releaseDate: '2023-07-28 T00:00:00.000+00:00',
+        category: 'pop',
+        spotifyUrl: 'https://api.spotify.com/v1/albums/6r1lh7fHMB499vGKtIyJLy',
+        averageRating: 4,
+        numOfReviews: 2,
+      
+    },
```
akosasante: I think some of your brackets got a bit mis-aligned here. 

Looking at a request I made on this endpoint in the docs (you could alternatively do the same in Postman)



the shape of the response is:

```
{
  "user": {
    "_id": "64d6ca92a15d2e18ab96a2a3",
    "name": "Akos123123",
    "email": "akos123@example.com",
    "role": "user",
    "__v": 0,
    "username": "akos92",
    "purchasedAlbums": [
      {
        "_id": "64ef50c8c5551074444547bc",
        "album": {
          "_id": "64d2a94c793389a43fc5a8d6",
          "artistName": "F",
          "albumName": "AUSTIN",
          "price": 0,
          "image": "https://i.scdn.co/image/ab67616d0000b27371cae34ad5a39bdab78af13e",
          "releaseDate": "2023-07-28T00:00:00.000Z",
          "spotifyUrl": "https://api.spotify.com/v1/albums/6r1lh7fHMB499vGKtIyJLy",
          "averageRating": 0,
          "numOfReviews": 0,
          "__v": 0,
          "createdAt": "2023-08-08T20:45:00.942Z",
          "updatedAt": "2023-08-17T10:10:42.861Z",
          "id": "64d2a94c793389a43fc5a8d6"
        },
        "user": "64d6ca92a15d2e18ab96a2a3",
        "__v": 0
      }
    ],
    "id": "64d6ca92a15d2e18ab96a2a3"
  },
  "purchasedAlbumCount": 1
}
```

We can basically copy that into this swagger.js file:

```suggestion
    UserWithAlbums: {
        $id: '64d6ca92a15d2e18ab96a2a3',
        $name: 'John Doe',
        $username: 'johndoe',
        $email: 'john@example.com',
        $role: { '@enum': ['user', 'admin'] },
        profileImage: {
            url: 'https://example.com/profile-image.jpg',
        },
        creditCardInfo: {
            hashedNumber: '1234567890',
            expiry: '12/24',
            preferredPaymentOption: {
                '@enum': ['credit card', 'paypal', 'google pay'],
            },
        },
        $purchasedAlbums: [{
            $_id: '64ef50c8c5551074444547bc',
            $user: '64d6ca92a15d2e18ab96a2a3',
            $album: {
                $id: '64d2a94c793389a43fc5a8d6',
                $albumName: '...Baby one more time',
                $artistName: 'Britney Spears',
                $createdAt: '2023-08-08T20:45:00.942Z',
                $updatedAt: '2023-08-17T10:10:42.861Z',
                $price: 20.99,
                image: 'https://i.scdn.co/image/ab67616d0000b27371cae34ad5a39bdab78af13e',
                spotifyUrl: 'https://api.spotify.com/v1/albums/6r1lh7fHMB499vGKtIyJLy',
                releaseDate: '2023-07-28T00:00:00.000Z',
                averageRating: 0,
                numOfReviews: 0,
            },
        }],
    },
```

Then, in the controller, we can tell swagger-autogen to use this as the response for the `withAlbums` endpoint by saying: 

```
  /*
     #swagger.summary = 'Fetch a user by id and all albums they purchased.'
     #swagger.responses[200] = {
				description: 'User and their albums purchased fetched successfully.',
				schema: { user: { $ref: '#/definitions/UserWithAlbums' }, purchasedAlbumCount: 1}
		 }
		  #swagger.responses[404] = { description: 'No user with id found.' }
  */
```

src/controllers/userController.js
NatachaKey: @akosasante thank you!! so glad to learn new things!!!

swagger.js
NatachaKey: @akosasante thank you, Akos, just a little question- to test this rote I login, create an order but when a make a GET request. to http://localhost:8000/api/v1/users/showMe/with
Captura de pantalla 2023-08-31 a las 11 42 50
Captura de pantalla 2023-08-31 a las 11 42 12
Albums I don't get album that you do in swagger (on the image)- what am I doing wrong?

akosasante: This is expected. We aren't yet creating the `PurchasedAlbum` object when creating the order -- see this ticket: https://codethedreamteam4-2023.atlassian.net/browse/PS-125

For testing I just created a PurchasedAlbum object for my user in a node console. You could also do it directly in the Mongo Atlas UI

akosasante (approved):
Looks great, thank you for all the hard work on this one! :tada: 

Very last thing is just make sure all the CI is happy

(`npm run generate-api-docs` \<-- make all the latest changes are reflected, `npm run lint:fix && npm run format && npm run test` \<-- make sure all the CI should be okay)

swagger.js
NatachaKey: @akosasante wow!!! thank you-i've just tried it (creating an entry in mongo db). just curious- to create a PurchasedAlbum object from node console did you do smth like -in a root  added newFileForCreatingPurchasedAlbum.js  and   executed it it with `node  newFileForCreatingPurchasedAlbum.js` - ```
require('dotenv').config();
const PurchasedAlbum = require('./src/models/PurchasedAlbum');

// Connect to the MongoDB database
mongoose.connect(process.env.MONGO_URL, { useNewUrlParser: true, useUnifiedTopology: true })
  .then(() => {
    console.log('Connected to the database');
    
    // Replace with the actual user and album IDs
    const userId = '111111111';
    const albumId = '2222222';
    
    // Create a PurchasedAlbum object
    const purchasedAlbum = new PurchasedAlbum({
      album: albumId,
      user: userId
    });

    // Save the purchased album to the database
    purchasedAlbum.save()
      .then(savedPurchasedAlbum => {
        console.log('PurchasedAlbum saved:', savedPurchasedAlbum);
        mongoose.disconnect();
      })
      .catch(error => {
        console.error('Error saving PurchasedAlbum:', error);
        mongoose.disconnect();
      });
  })
  .catch(error => {
    console.error('Error connecting to the database:', error);
  });
```
?

NatachaKey:
> npm run test

thank you for such a great and detailed explanations! just ran all these commands - it only fails in jest files, probable need to fix that

PR #35, opened 2023-08-27
API endpoint for deleting a review  PS-75

DYA13:


src/controllers/reviewController.js
```diff
@@ -123,9 +123,47 @@ const getSingleReview = async (req, res) => {
...
+    if (error instanceof CustomError) {
+      // Handle custom errors
+      res.status(error.statusCode).json({ error: error.message });
+    } else {
+      // Handle other errors
+      res
+        .status(StatusCodes.INTERNAL_SERVER_ERROR)
+        .json({ error: 'Internal server error' });
+    }
+  }
```
akosasante: Don't need to use try/catch to handle errors, because any errors that are thrown in a controller function in this codebase will be handled in the [ErrorHandler middleware](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/middleware/error-handler.js).  If there's some specific functionality or messaging that you're needing then we can either update the middleware or throw a custom error with the particular message that you want.

```suggestion
  // Fetch the existing review
  const review = await Review.findOne({ _id: reviewId });

  // Check if the review exists
  if (!review) {
    throw new CustomError.NotFoundError(`No review with id ${reviewId}`);
  }

  // Check if the requesting user is the author of the review
  if (review.user.toString() !== req.user.userId) {
    throw new CustomError.ForbiddenError(
      'You are not authorized to delete this review'
    );
  }

  // Delete the review
  await review.remove();

  res.status(StatusCodes.OK).json({ message: 'Review deleted successfully' });
```

akosasante (approved):
Just need to unwrap the code from the try/catch. Then you should be good to merge, great job!

PR #36, opened 2023-08-28
 PS-94API endpoint for viewing a particular order

NatachaKey:


src/routes/orderRoutes.js
```diff
@@ -1,8 +1,24 @@
...
 
-router.post('/', authenticateUser, createOrder);
+router
+  .route('/')
+  .post(authenticateUser, createOrder)
+  .get(authenticateUser, authorizePermissions('admin'), getAllOrders);
+
+router
+  .route('/:id')
+  .get(authenticateUser, authorizePermissions('admin'), getSingleOrder)
```
akosasante: For fetching a single order, I think we'll need to allow non-admins to do this too; we can do `checkPermisisons` inside the controller to make sure it's the right user or an admin.

```suggestion
  .get(authenticateUser, getSingleOrder)
```

src/models/Order.js
```diff
@@ -1,7 +1,7 @@
 const mongoose = require('mongoose');
 
 // mongoose schema for the individual order items
-const orderItemSchema = new mongoose.Schema({
+const OrderItemSchema = mongoose.Schema({
```
akosasante: Was there a reason to change this from using the `new ...` constructor?

src/models/Order.js
```diff
@@ -70,8 +75,8 @@ const orderSchema = new mongoose.Schema(
 const updateOrderStatus = async () => {
   try {
     // Get the current date and subtract 1 minute
-    const oneMinutesAgo = new Date(Date.now() - 1 * 60 * 1000);
-
+    //const oneMinutesAgo = new Date(Date.now() - 1 * 60 * 1000);
+    const oneMinutesAgo = new Date(Date.now() - 72000000); // for tests -delete later
```
akosasante: This is fine; should probably chat with the team to determine what's a reasonable amount of time. And we could even make it an envirionment variable so that we can more easily have a longer time for local dev vs in production or vice vers.

src/models/Order.js
```diff
@@ -17,14 +17,19 @@ const orderItemSchema = new mongoose.Schema({
...
     user: {
       type: mongoose.Schema.Types.ObjectId,
       ref: 'User', // Referencing the User model
       required: [true, 'Please provide user'],
     },
+    email: {
+      type: mongoose.Schema.Types.ObjectId,
+      ref: 'User', // Referencing the User model
+      required: [true, 'Please provide email'],
+    },
```
akosasante: Since we're already referencing the user on line `23-27`, I'm not sure that we need this new field here. We could instead just populate the user field with it's email when we fetch the order?

Eg:

```javascript
Order.findById(orderId).populate({path: 'user', options: { select: { password: 0 } } })
```

Which will return an order object, with an `order.user` field that has all the user info except for the user's password.

src/models/Order.js
```diff
@@ -85,10 +90,12 @@ const updateOrderStatus = async () => {
 };
 
 // Call the updateOrderStatus function every 1 minute (60 * 1000 milliseconds)
-const intervalId = setInterval(updateOrderStatus, 1 * 60 * 1000);
+
+//const intervalId = setInterval(updateOrderStatus, 1 * 60 * 1000);
+const intervalId = setInterval(updateOrderStatus, 72000000); // 2 hours
```
akosasante: Just fyi this (72,000,000) is actually **20 hours**. 2 hours would be (2 hrs \* 60 min/hr \* 60 sec/min \* 1000 millisec/sec = 7,200,000)

src/models/Order.js
```diff
@@ -97,23 +104,23 @@ orderSchema.pre('find', async function (next) {
...
   next();
 });
-// TTL (Time To Live) index to automatically delete orders with "cancelled" status after 2 minutes
-orderSchema.index(
+// TTL (Time To Live) index to automatically delete orders with "cancelled" status after 2 hours
+OrderSchema.index(
   { expiresAt: 1 },
   {
-    expireAfterSeconds: 120,
+    expireAfterSeconds: 7200,
```
akosasante, line 122: :white\_check\_mark:

src/controllers/orderController.js
```diff
@@ -39,4 +42,33 @@ const createOrder = async (req, res) => {
...
+};
+
+const deleteOrder = async (req, res) => {
+  const { id: orderId } = req.params;
+  const order = await Order.findOneAndDelete({ _id: orderId });
+
+  if (!order) {
+    throw new CustomError.NotFoundError(`No order with id ${orderId}`);
+  }
+  checkPermissions(req.user, order.user);
```
akosasante, line 72: At this point, the order has already been deleted if it was found on line 64, so checking if this was the right user here will have no effect. If it was the wrong user, it will throw a Unauhtorized error, but we'll still have already deleted the order.

Two options:

1.  split the "find order" and "split order" steps
2.  use a transaction that will rollback if there are any errors (https://mongoosejs.com/docs/transactions.html)

```javascript
// Option 1
const order = await Order.findById(orderId)
if (!order) {
  throw new NotFoundError(`No order with id ${orderId}`);
}
checkPermissions(req.user, order.user)

await Order.findByIdAndDelete(orderId); // or more simply can just call delete on the documnent we've already fetched:  await order.remove()

res.status(StatusCodes.OK).json({ msg: 'Success! Order was deleted' })

// Option 2
const session = await Order.startSession // start a MongoDb client session
await session.withTransaction(async () => {
  // everything in this block is in a transaction, which means it will be undone if there is an error thrown
  const order = await Order.findOneAndDelete({ _id: orderId });

  if (!order) {
    throw new CustomError.NotFoundError(`No order with id ${orderId}`);
  }
  checkPermissions(req.user, order.user); // if this throws an "UnauthorizedError" the delete step above will be reverted
});

// assuming we get here, that means we haven't thrown an error, so we can return suiccess response:
res.status(StatusCodes.OK).json({msg: "Success! Order was deleted" })'
```

src/controllers/orderController.js
```diff
@@ -16,6 +18,7 @@ const createOrder = async (req, res) => {
   // Create an Order document in the database
   const order = await Order.create({
     user: req.user.userId,
+    email: req.user.email,
```
akosasante: Per comment in the Order.js file; I'm not sure that we need this

src/models/Order.js
NatachaKey: @akosasante I thought it works without `new` too. I forgot to write it in Review.js and i guess it worked ok. just added it to both Schemas

src/models/Order.js
NatachaKey: @akosasante thank you! will commit it -please take a look if it's ok

src/models/Order.js
NatachaKey: @akosasante thank you, Akos! will commit changes now

src/models/Order.js
NatachaKey: @akosasante does this mean that two hours after the order being cancelled this order will be deleted from the db? so if we are in dev mode and the status of unpaid order will change in 20h so 22h later it will be deleted from mongo, right? What does { expiresAt: 1 } mean in this case? in mongo we have 
createdAt
2023-08-08T20:45:00.942+00:00
updatedAt
2023-08-17T10:10:42.861+00:00 but can't find expiresAt- where did it come from?

NatachaKey: are we sorting all orders in mongo db depending on it's time that it has a cancelled status? basically putting the oldest one first to delete it first ?

src/controllers/orderController.js
NatachaKey: @akosasante thank you  so much for showing new approaches!!! Appreciate it ❤️

src/models/Order.js
akosasante: Ah I see, looks like you're correct it works both when invoking as a class constructor (`new`) and when calling it as a function :+1: . I just wasn't sure if you made the change for a particular reason here. But seems fine either way

src/models/Order.js
```diff
@@ -1,7 +1,10 @@
 const mongoose = require('mongoose');
+if (process.env.NODE_ENV !== 'production') {
+  require('dotenv').config();
+}
```
akosasante: Why do we only need dotenv in non-production? aren't we fetching values from `process.env` in both environments? And also doesn't the dotenv get already loaded in `expressServer.js` when the app starts anyway?

src/models/Order.js
```diff
@@ -69,10 +72,14 @@ const orderSchema = new mongoose.Schema(
 // Create a function to update order statuses
 const updateOrderStatus = async () => {
   try {
-    // Get the current date and subtract 1 minute
-    const oneMinutesAgo = new Date(Date.now() - 1 * 60 * 1000);
+    let isDevelopment = process.env.NODE_ENV !== 'production';
+    const timeDuration = isDevelopment
+      ? parseInt(process.env.DEV_TIME_DURATION)
+      : parseInt(process.env.PROD_TIME_DURATION);
```
akosasante: Maybe give these a clear name so that if someone is looking at just the `.env` file it's clear what these are used for. Eg: `DEV_ORDER_EXPIRY_DURATION` or something

src/models/Order.js
```diff
@@ -84,11 +91,16 @@ const updateOrderStatus = async () => {
...
 };
 
-// Call the updateOrderStatus function every 1 minute (60 * 1000 milliseconds)
-const intervalId = setInterval(updateOrderStatus, 1 * 60 * 1000);
+// Call the updateOrderStatus function every .... x time
+
+//const intervalId = setInterval(updateOrderStatus, process.env.PROD_TIME_DURATION); //prod 1h
+const intervalId = setInterval(
+  updateOrderStatus,
+  process.env.DEV_TIME_DURATION
```
akosasante: This should also vary by prod vs dev right? 

```suggestion
  timeDuration // in prod 1hr, in dev 20hrs
```

src/models/Order.js
```diff
@@ -69,10 +72,14 @@ const orderSchema = new mongoose.Schema(
...
 const updateOrderStatus = async () => {
   try {
-    // Get the current date and subtract 1 minute
-    const oneMinutesAgo = new Date(Date.now() - 1 * 60 * 1000);
+    let isDevelopment = process.env.NODE_ENV !== 'production';
+    const timeDuration = isDevelopment
+      ? parseInt(process.env.DEV_TIME_DURATION)
+      : parseInt(process.env.PROD_TIME_DURATION);
+    // Get the current date and subtract __x   minutes
+    const oneMinutesAgo = new Date(Date.now() - timeDuration);
```
akosasante: Since this is no longer always a minute ago, maybe rename the variable something more general that makes sense in both cases: `expiryTimeSinceCreation`

src/models/Order.js
akosasante: > does this mean that two hours after the order being cancelled this order will be deleted from the db? so if we are in dev mode and the status of unpaid order will change in 20h so 22h later it will be deleted from mongo, right?

Yes, that's correct.

> What does { expiresAt: 1 } mean in this case? in mongo we have

I'm not super familiar with Mongo + expiry/TTL (time-to-live) but my understanding from reviewing Marina's ticket is that it is basically a boolean. So 0 = false, an 1 = true. When true, it tells Mongo this document can be deleted. But since it's part of this index, this statement is saying set `expiry` flag on this document to `true` after the `expireAfterSeconds` time has passed. And the `partialFilterExpression` means this should only happen to orders with an `orderStatus` of `'cancelled'`

akosasante: Looking at : [https://www.mongodb.com/docs/manual/core/index-ttl/](https://www.mongodb.com/docs/manual/core/index-ttl/) and [https://www.mongodb.com/docs/manual/core/index-partial/](https://www.mongodb.com/docs/manual/core/index-partial/) it may be more accurate to say that yes we're creating a sorting of documents by this hidden field `expiresAt` but otherwise what I said above should still apply

NatachaKey: @akosasante got it!!! thanks for the explanations!!!

src/models/Order.js
NatachaKey: @akosasante sure! will commit changes now, just to make sure-  when we create express server in ### expressServer.js with `const app = express();`  and call it in **app.js** which gets executed by default when we run the server with _npm run dev_ -this action triggers everything related with **app**  (libraries, middlewares) so we don't need to require libraries and packages that we've already put in expressServer.js ? for examle everytime I was using environment variables in different modules- i was requring   `require('dotenv').config();`

src/models/Order.js
NatachaKey: @akosasante sure! thank you

src/models/Order.js
NatachaKey: @akosasante will commit it, please check if I did it correctly- I had to move 
 ```
let isDevelopment = process.env.NODE_ENV !== 'production';
    const timeDuration = isDevelopment
      ? parseInt(process.env.DEV_ORDER_EXPIRY_DURATION)
      : parseInt(process.env.PROD_ORDER_EXPIRY_DURATION); 
```

to the globe scope so it could be accesible both from **updateOrderStatus** and i**ntervalId**

src/models/Order.js
NatachaKey: @akosasante sure!! I was thinking about it too

src/models/Order.js
akosasante: Yes that's correct. The only time you may need to add it is for scripts that you're running without starting up the express app. For example, the `fetchAlbumSpotify.js` script. Though even in that case you can require it via command-line flags instead if you prefer: `node -r dotenv/config fetchAlbumSpotify.js`

akosasante (approved):
LGTM :rocket:  But check the lint errors before you merge since I think you're missing an import in `orderController` and do a quick test in Postman to make sure everything works as expected then feel free to merge.

src/models/Order.js
NatachaKey: @akosasante thank you!!!

PR #37, opened 2023-08-28
Loop order

NatachaKey:
tried to see in mongo db order items information

### DIFF
--- a/src/controllers/orderController.js
+++ b/src/controllers/orderController.js
@@ -6,15 +6,68 @@ const CustomError = require('../errors');
 const { checkPermissions } = require('../utils');
 
 const createOrder = async (req, res) => {
-  const { orderItems, subtotal, tax, total } = req.body;
+  //process order
+  const processOrder = async (resultsMap, item) => {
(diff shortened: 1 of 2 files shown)


src/controllers/orderController.js
```diff
@@ -6,15 +6,68 @@ const CustomError = require('../errors');
...
+    throw new CustomError.BadRequestError('No cart items provided');
+  }
+  
+  // initial accumulator value. For each iteration through cartItems, we will do some processing in `processOrder` function to update this object
+  const initialValue = { subtotal: 0, orderItems: [] };
+
+  // Now we execute the `reduce` function, telling it to run the `processOrder` function for each item in cartItems
+  // We also give it the `initialValue` so that for the first iteration, it will use subtotal=0, and orderItems=[]
+  // At the end, the `reduce` function returns the updated accumulator value, which we destrucure to get the subtotal and the orderItems variables
+  const { subtotal, orderItems } = await cartItems.reduce(
```
akosasante, line 65: In this particular case, our frontend \_should\_ already be passing us the subtotal along with the well-formatted order items; is there a reason to recalculate it on the backend?

NatachaKey: yes,I thought it about it...will close this PR -I think it's unseful

PR #38, opened 2023-08-28
Ps158 nodemailer to send email

mvstoyan:


mvstoyan:
When I run node -r dotenv/config src/mailing/test.js, I encounter an error with the code 'Invalid login: 535 5.7.3 Authentication unsuccessful.' This appears to be related to authentication (username and password) during email sending and not with the code itself. It's possible that I may have entered the credentials incorrectly in the .env file. Could you please guide me in the right direction?
mail

package.json
akosasante, line 1: Make sure to double check before merging that you've pulled in main to this branch. I'm seeing `stripe` package get deleted here which I think is unexpected

src/mailing/sender.js
```diff
@@ -0,0 +1,35 @@
+const nodemailer = require('nodemailer');
+require('dotenv').config();
```
akosasante: Do we need this usually or is it just for testing? I would expect that this would usually already have been loaded when the app starts up by `expressServer.js`

akosasante: If it's just for testing, letting you know that you can also require dotenv when running a node script by using the `-r` flag: `node -r dotenv/config src/mailing/test.js` (then in that case don't need to duplicate requiring it in the code)

src/mailing/test.js
```diff
@@ -0,0 +1,13 @@
+const sendTestEmail = require('./sender');
+const recipient = process.env.RECIPIENT_EMAIL;
```
akosasante: Since this is the recipient email just for testing may be clearer to call this environment variable `TEST_RECIPIENT_EMAIL`

akosasante (approved):
Small comments but otherwise looks good to go!

src/mailing/sender.js
mvstoyan: Removed

src/mailing/test.js
mvstoyan: Thank you very much, everything works🎉. I received a test email.
mail
test

mvstoyan:
@akosasante I'm not sure what to do in this situation. It seems like there's an error 'connect ECONNREFUSED' or 'Could not open connection to the host,'. I suspect it's an issue with my computer not being able to establish a connection with the SMTP server. There might also be a problem with the sender's email. Can you guide me in the right direction?

akosasante:
> @akosasante I'm not sure what to do in this situation. It seems like there's an error 'connect ECONNREFUSED' or 'Could not open connection to the host,'. I suspect it's an issue with my computer not being able to establish a connection with the SMTP server. There might also be a problem with the sender's email. Can you guide me in the right direction? 
> 
> 

Do some console.logging and double check that all your env variables are being loaded correctly. Since we changed the name of `RECIPIENT_EMAIl` to `TEST_RECIPIENT_EMAIL` I'm wondering if you updated it in your local `.env` file as well? I was able to send to my own email with no issues :thinking:

package.json
akosasante, line 1: Right now Jest thinks that `src/mailing/test.js` is a test that it's supposed to run since the name is `test`. Which is causing the test command to fail: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/actions/runs/6019347318/job/16329036965?pr=38](https://github.com/Code-the-Dream-School/dd-prac-team4-back/actions/runs/6019347318/job/16329036965?pr=38)

 We can either change the name of the file, or preferably just tell jest to ignore anything in the `src` folder since it should only be running test on tiles in the `test` folder anyway. In `package.json` file update these lines:

```
    "test": "jest --watchAll=false --detectOpenHandles --silent --testPathIgnorePatterns=src",
    "test:watch": "jest --watchAll=true --detectOpenHandles --testPathIgnorePatterns=src",
```

mvstoyan: Thank you very much!

mvstoyan: Yes, I updated the data in .env, and my test passed once to the spam. Now it's failing again. But if it's working for you, then it must be an issue with my computer. Could you please double-check if the test email is work for you? If it is, I will merge this branch so I don't hold up the team's work. I will continue to investigate the issue with my computer and SMTP after.

akosasante (approved):
`git pull`ed to make sure my local is up-to-date with what you have here. 

ran `node -r dotenv/config ./src/mailing/test.js`

✅ received an email in my gmail account that I had set as the recipient in the .env files
✅  output in the terminal also indicated success:

```
aasante@aasante-C02F75HHMD6R backend % node -r dotenv/config ./src/mailing/test.js 
Test email sent successfully: {
  accepted: [ 'akosuaasante@gmail.com' ],ng/          models/           routes/           views/            
  rejected: [],
  ehlo: [
    'SIZE 157286400',
    'PIPELINING',
    'DSN',
    'ENHANCEDSTATUSCODES',
    'AUTH LOGIN XOAUTH2',
    '8BITMIME',
    'BINARYMIME',
    'CHUNKING',
    'SMTPUTF8'
  ],
  envelopeTime: 127,
  messageTime: 517,
  messageSize: 628,
  response: '250 2.0.0 OK <DM4PR11MB7327CCEDAEB7D51E11BE75E598E6A@DM4PR11MB7327.namprd11.prod.outlook.com> [Hostname=DM4PR11MB7327.namprd11.prod.outlook.com]',
  envelope: {
    from: 'codethedream.practicum.team4@outlook.com',
    to: [ 'akosuaasante@gmail.com' ]
  },
  messageId: '<1cb5dad0-32cd-3532-dae2-3c58b67ed3e6@outlook.com>'
}

```

PR #39, opened 2023-08-29
swagger-autogen

mvstoyan:
Each auth controller handler function, configure information using comments swagger-autogen

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "morgan": "^1.10.0",
     "nodemailer": "^6.9.4",
     "passport": "^0.6.0",
+    "socket.io": "^4.7.2",
     "spotify-web-api-node": "^5.0.2",
     "stripe": "^13.0.0",
     "swagger-ui-dist": "^5.4.2",
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     "mongoose": "^7.4.1",
     "morgan": "^1.10.0",
     "passport": "^0.6.0",
-    "stripe": "^13.0.0",
     "spotify-web-api-node": "^5.0.2",
     "stripe": "^13.0.0",
     "swagger-ui-dist": "^5.4.2",
--- a/package.json
+++ b/package.json
@@ -10,8 +10,8 @@
     "lint:fix": "eslint . --fix",
     "format": "prettier --write .",
     "format:check": "prettier --check .",
-    "test": "jest --watchAll=false --detectOpenHandles --silent",
-    "test:watch": "jest --watchAll=true --detectOpenHandles",
+    "test": "jest --watchAll=false --detectOpenHandles --silent --testPathIgnorePatterns=src",
+    "test:watch": "jest --watchAll=true --detectOpenHandles --testPathIgnorePatterns=src",
     "generate-api-docs": "node ./swagger.js"
   },
   "author": "",
@@ -37,6 +37,7 @@
     "mongo-sanitize": "^1.1.0",
     "mongoose": "^7.4.1",
     "morgan": "^1.10.0",
+    "nodemailer": "^6.9.4",
     "passport": "^0.6.0",
     "spotify-web-api-node": "^5.0.2",
     "stripe": "^13.0.0",
--- a/src/controllers/authController.js
+++ b/src/controllers/authController.js
@@ -5,30 +5,6 @@ const CustomError = require('../errors');
 const { attachCookiesToResponse, createTokenUser } = require('../utils');
 
 const register = async (req, res) => {
-  /* 
-    #swagger.summary = 'Register a new user'
-    #swagger.parameters['user'] = {
-      in: 'body',
-      description: 'User registration information',
-      required: true,
-      schema: {
-        $name: "John Doe",
-        $email: "jdoe@example.com",
-        $password: "supersecretpassword",
-        $username: "jdoe99"
-      }
-    }
-    #swagger.responses[201] = {
-      description: 'User registered successfully',
-      schema: {
-        $ref: '#/definitions/TokenizedUser'
-      }
-    }
-    #swagger.responses[400] = {
-      description: 'Bad request, missing email, name, password, or username'
-    }
-  */
-
   const { name, email, password, username } = req.body;
 
   if (!email || !name || !password || !username) {
@@ -54,9 +30,51 @@ const register = async (req, res) => {
   attachCookiesToResponse({ res, user: tokenUser }); // Attach the token to cookies and send in the response
 
   res.status(StatusCodes.CREATED).json({ user: tokenUser }); // Send a successful response with user data
+  /* 
+    #swagger.summary = 'Register a new user'
+    #swagger.parameters['user'] = {
+      in: 'body',
+      description: 'User registration information',
+      required: true,
+      schema: {
+        $name: "John Doe",
+        $email: "jdoe@example.com",
+        $password: "supersecretpassword",
+        $username: "jdoe99"
+      }
+    }
+    #swagger.responses[201] = {
+      description: 'User registered successfully',
+      schema: {
+        $ref: '#/definitions/TokenizedUser'
+      }
+    }
+    #swagger.responses[400] = {
+      description: 'Bad request, missing email, name, password, or username'
+    }
+  */
 };
 
 const login = async (req, res) => {
+  const { email, password } = req.body;
+
+  if (!email || !password) {
+    throw new CustomError.BadRequestError('Please provide email and password'); // If either field is not provided, throw an error
+  }
+  const user = await User.findOne({ email }); // Find a user by email
+
+  if (!user) {
+    throw new CustomError.UnauthenticatedError('Invalid Credentials');
+  }
+
+  const isPasswordValid = await argon2.verify(user.password, password); // Check password validity
+  if (!isPasswordValid) {
+    throw new CustomError.UnauthenticatedError('Invalid Credentials'); // If password is incorrect, throw an error
+  }
+  const tokenUser = createTokenUser(user);
+  attachCookiesToResponse({ res, user: tokenUser });
+
+  res.status(StatusCodes.CREATED).json({ user: tokenUser });
   /* 
     #swagger.summary = 'Login as a user'
     #swagger.parameters['user'] = {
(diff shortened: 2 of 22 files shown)


src/expressServer.js
```diff
@@ -101,7 +101,7 @@ app.use(passport.session());
 const connectDB = require('./db/connect');
 
 //routers
-const authRouter = require('./routes/authRoutes');
+const authRouter = require('./routes/authRoutes'/* #swagger.tags = ['Authentication'] */);
```
akosasante: This comment needs to be on the `app.use` where the route is defined (line 114); not here where it is imported.

```suggestion
const authRouter = require('./routes/authRoutes'); 
```

src/controllers/authController.js
```diff
@@ -5,6 +5,36 @@ const CustomError = require('../errors');
...
+      required: true,
+      schema: {
+        type: 'object',
+        properties: {
+          name: { type: 'string' },
+          email: { type: 'string' },
+          password: { type: 'string' },
+          username: { type: 'string' }
+        }
+      }
```
akosasante: The swagger-autogen library wants us to write the schema not in the formal OpenApi/Swagger spec way, but rather exactly as the example object you would want to see in the docs, and then it will figure out the tyeps from there: [https://github.com/swagger-autogen/swagger-autogen#schema-and-definitions](https://github.com/swagger-autogen/swagger-autogen#schema-and-definitions)

```suggestion
      schema: {
        $name: "John Doe",
        $email: "jdoe@example.com",
        $password: "supersecretpassword",
        $username: "jdoe99"
      }
```

We use the dollar sign in front of the field to denote that it is required

src/controllers/authController.js
```diff
@@ -5,6 +5,36 @@ const CustomError = require('../errors');
...
+    }
+    #swagger.responses[201] = {
+      description: 'User registered successfully',
+      schema: {
+        $ref: '#/definitions/User' // Reference to the User object definition
```
akosasante: These definitions come from the `swagger.js` file. For example, we've defined a `PasswordlessUser` object there that is used in the UserController API docs. 

But for this endpoint, we don't return a full user from the db, rather we return a tokenized version of it. So we should first create the definition in `swagger.js` for a TokenizedUser, and then refer to it here with `$ref: '#/definitions/TokenizedUser'`

Also, it looks like swagger-autogen gets confused with comments iniside of the swagger comments, so you may end up having to either remove or move `// Reference to the User object definition` to above line 24.

src/controllers/authController.js
```diff
@@ -5,6 +5,36 @@ const CustomError = require('../errors');
...
+      }
+    }
+    #swagger.responses[400] = {
+      description: 'Bad request, missing email, name, password, or username'
+    }
+    #swagger.responses[409] = {
+      description: 'Conflict, email already exists'
+    }
```
akosasante: I agree that this should return a 409 but looking at the code on line 46, right now we're just returning a 400. So either have the docs match what's in the code or vice versa

src/controllers/authController.js
```diff
@@ -33,6 +63,33 @@ const register = async (req, res) => {
...
+      in: 'body',
+      description: 'User email and password for login',
+      required: true,
+      schema: {
+        type: 'object',
+        properties: {
+          email: { type: 'string' },
+          password: { type: 'string' }
+        }
+      }
```
akosasante:
```suggestion
      schema: {
        email: "jdoe@example.com",
        password: "supersecretpassword"
      }
```

src/controllers/authController.js
```diff
@@ -55,8 +112,25 @@ const login = async (req, res) => {
...
-
 const logout = async (req, res) => {
+  /*
+    #swagger.summary = 'Logout a user'
+    #swagger.parameters['token'] = {
+        in: 'cookie',
+        description: 'JWT token in the cookie',
+        required: true,
+        type: 'string'
+    }
```
akosasante: oh interesting I didn't realize we could define the cookie value here. Were you able to get it to work with the "Try It" feature? Like did it actually send the cookie entered ? 

I think otherwise, we can just leave this out, and the browser can handle setting/fetching the cookie as it usually would.

akosasante: I'll look into a bit more today to understand better how it works with swagger-autogen :thinking:

src/controllers/authController.js
```diff
@@ -55,8 +112,25 @@ const login = async (req, res) => {
...
+    #swagger.responses[200] = {
+        description: 'User logged out successfully',
+        schema: { type: 'object', properties: { msg: { type: 'string' } } }
+    }
+    #swagger.security = [{ "JWT": [] }]  // Specify that JWT token is required for access
+    #swagger.responses[401] = {
+        description: 'Unauthorized',
+        schema: { type: 'object', properties: { msg: { type: 'string' } } }
+    }
```
akosasante: I'm not seeing in the logout function code in which case we return a 401?

src/controllers/authController.js
mvstoyan: Removed 409

src/controllers/authController.js
mvstoyan: tokenizedUser
tokenizedUser1


removed comments also.

src/controllers/authController.js
mvstoyan: schema

src/controllers/authController.js
mvstoyan: I read the Swagger documentation, and it doesn't explicitly specify how to document the 'token' parameter, which in my code is expected to be a JWT token in a cookie. I included the 'token' parameter with the description 'JWT token in the cookie' in the code comments because it might be useful information for other developers. I understand that my approach might not be entirely accurate😅, and I couldn't test it due to the errors in the code above.

src/controllers/authController.js
mvstoyan: Removed

package.json
akosasante, line 1: Make sure to update branch before merging so we're not accidentally deleting this stripe library

swagger.js
```diff
@@ -31,6 +31,15 @@ const doc = {
...
     },
+    TokenizedUser: {
+      type: 'object',
+      properties: {
+        token: { type: 'string' },
+        user: {
+          $ref: '#/definitions/PasswordlessUser',
+        },
+      },
+    },
```
akosasante: For swagger-autogen we don't need to tell it the type/properties, it will infer those. Rather it wants us to give it an example of what this object will look like and it will figure out the types from there. So for example, if I make a call to the register endpoint (via postman, or the api docs, etc) I get a response like:

```json
{
  "user": {
    "name": "John Doe",
    "userId": "64ef5885c2a489217c571253",
    "role": "user"
  }
}
```

The nice thing about swagger-autogen is we can just copy paste that exactly into our swagger.js definition here:

```suggestion
    TokenizedUser: {
      user: {
        name: "John Doe",
        userId: "64ef5885c2a489217c571253",
        role: { '@enum': ['user', 'admin'] },
      },
    },
```

Unlike the user controller endpoints, the auth endpoints don't return a full user object so we can't just use a reference to the PasswordlessUser definition above as that would make us generate misleading documentation.

src/controllers/authController.js
akosasante: I think for now, I would recommend explaining it via the description box:

```
#swagger.description = 'This endpoint checks for a signed JWT token in an HTTP-only cookie and clears it. It checks the `token` field in the cookie.'
```

akosasante: What you have works and does follow the Swagger docs :white\_check\_mark:  My recommendation to change it is mainly just to make the API docs easier to navigate for everyone. 

The browser already takes care of storing and passing along values in cookies. But if we add your line here, then the user using the api docs has to manually grab the token value and paste it into the box to try out any authenticated endpoint, which feels like extra friction that we don't really need.


Here's what I currently have to do in order to try the logout endpoint

https://github.com/Code-the-Dream-School/dd-prac-team4-back/assets/9094098/d54729f1-5b61-4105-a328-2d9f56a10898


Here's an example of testing login and logout without the jwt token input. Everything works because the browser takes care of passing along the token in the cookie:

https://github.com/Code-the-Dream-School/dd-prac-team4-back/assets/9094098/341a27a8-0925-4570-9f45-265980555f1a

mvstoyan: Thank you very much, so helpful for me ❤️

swagger.js
mvstoyan: swagger

src/controllers/authController.js
mvstoyan: swagger1

src/controllers/authController.js
```diff
@@ -55,8 +103,16 @@ const login = async (req, res) => {
...
 
 //logout endpoint
-
 const logout = async (req, res) => {
+  /*
+    #swagger.summary = 'Logout a user'
+    #swagger.description = 'This endpoint checks for a signed JWT token in an HTTP-only cookie and clears it. It checks the `token` field in the cookie.'
+    #swagger.responses[200] = {
+        description: 'User logged out successfully',
+        schema: { type: 'object', properties: { msg: { type: 'string' } } }
```
akosasante: this is showing:



for swagger-autogen we want to define the schema literally as it will be shown:

```suggestion
        schema: { msg: 'user logged out!' }
```


https://github.com/Code-the-Dream-School/dd-prac-team4-back/assets/9094098/4d8bb1a8-a60d-4e81-a9f4-3291cd32f360

akosasante (approved):
small fix left; but then you're good to merge

mvstoyan:
@akosasante i have a problem with swagger-output.json, can't fix this file. I need a help i think😅

mvstoyan:
@akosasante  Node CI failed for me. Its show me what need to fix. But i dont know how to do this correct. This is my code screen. 
Снимок экрана 2023-09-02 160227
next screen - Node CI  
Снимок экрана 2023-09-02 155809
next screen - main branch 
Снимок экрана 2023-09-02 160409
What should I do in this situation?

akosasante:
```
[error] swagger-output.json: SyntaxError: Unexpected token, expected "," (12:12)
[error]   10 |   "schemes": [
[error]   11 |     "http",
[error] > 12 |   "servers": [
[error]      |            ^
[error]   13 |     {
[error]   14 |       "url": "http://localhost:8000/"
[error]   15 |     }
```

It looks like there's a `]` close bracket missing after line 11 to close the `"schemes"` array.

Since this file is a generated file; often the easiest way to fix issues is not to do it manually, but instead just regenerate it. So I would do the following:

*   `npm run generate-api-docs` -- this will recreate `swagger-output.json` and it should fix any issues that were in there
*   `npm run prettier` -- will format all the files (including `swagger-output.json` ) in a way that makes the CI happier

Then you can double check before pushing that everything will work in CI by doing the same steps that CI does (`npm run lint:check && npm run format:check`)

mvstoyan:
@akosasante Can you check one more time plz, before I will merge this branch.

src/controllers/authController.js
```diff
@@ -5,6 +5,30 @@ const CustomError = require('../errors');
...
+      description: 'User registered successfully',
+      schema: {
+        $ref: '#/definitions/TokenizedUser'
+      }
+    }
+    #swagger.responses[400] = {
+      description: 'Bad request, missing email, name, password, or username'
+    }
+  */
+
```
akosasante: I think you're running into a similar bug with the swagger-autogen library as @NatachaKey  did when working on the user controller docs. I'm still not yet sure why this happens and I'll be submitting a bug report to the library repo. But in the meanwhile, it seems to be fixed when you move the swagger doc comments to the _bottom_ of the function istead of the top (for example look at what was done for usercontroller here: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/PS111-API-docs-auth-controller/src/controllers/userController.js#L27-L48)](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/PS111-API-docs-auth-controller/src/controllers/userController.js#L27-L48))

It only seems to be an issue on the login and register handlers right now but you can move logout comments to the bottom too if you prefer.

akosasante (approved):
One small fix then should be good. Feel free to merge after that

akosasante (approved):
All looks good on my end now 🚀

PR #41, opened 2023-08-29
Swagger albums

NatachaKey:
responses and parameters are not seen as expected


package.json
akosasante, line 1: Double check after you pull `main` into this branch; but I don't expect to see the `stripe` package being deleted.

src/controllers/albumController.js
```diff
@@ -4,17 +4,48 @@ const CustomError = require('../errors');
...
+        schema: [{ $ref: '#/definitions/Album'' }]
+		 }
+		 
+  */
   req.body.user = req.user.userId;
   const album = await Album.create(req.body);
   res.status(StatusCodes.CREATED).json({ album });
 };
 
+// question for Akos - what does minimin means on line 20?
```
akosasante: I can't see what you're referring to

src/controllers/albumController.js
```diff
@@ -4,17 +4,48 @@ const CustomError = require('../errors');
 const logger = require('../../logs/logger');
 
 const createAlbum = async (req, res) => {
+  /*
+     #swagger.summary = 'Create new album and save it to the database'
+    
+     #swagger.responses[201] = {
+				description: 'Albums was successfully created.',
+        schema: [{ $ref: '#/definitions/Album'' }]
```
akosasante: We need to write the definition in our `swagger.js` so that the autogen knows what shape we're referring to. Similar to how `PasswordlessUser` is defined in that file:

Something like the following (though please double check if these are the correct field names / what the fields would look like by comparing against a real response from Postman)

```
Album: {
      albumName: 'The Dark Side of the Moon',
      artistName: 'Pink Floyd',
      price: 9.99,
      image: '/uploads/example.jpg',
      releaseDate: '2022-03-01T00:00:00.000Z',
      category: 'rock',
      spotifyUrl: 'https://open.spotify.com/album/4LH4d3cOWNNsVw41Gqt2kv',
      averageRating: 4.5,
      numOfReviews: 2
    }
```

akosasante: According to the return value of this function, we return an object with key "album" and value being the album that was just created: `res.status(StatusCodes.CREATED).json({ album });`

So this should not be an array here:

```suggestion
        schema: { album: { $ref: '#/definitions/Album' } }
```

akosasante: Also note that you had an extra quotation mark after definitions/Album that was causing swagger-autogen to not be able to parse somethings correctly here

src/controllers/albumController.js
```diff
@@ -4,17 +4,48 @@ const CustomError = require('../errors');
 const logger = require('../../logs/logger');
 
 const createAlbum = async (req, res) => {
+  /*
```
akosasante, line 10: swagger-autogen seems to get a bit confused about `req.body.user = req.user.userId` and thinks that that means that the API request actually has a body where the caller passes in a user object. But this is actually coming from our cookie, and the body for this request is all the info for an album. 

We can define the fields expected in a definition object called `NewAlbum` in the swagger.js file. This probably doesn't include all the same fields as the `Album` that's returned (eg: the caller doesn't give us the average rating or number of reviews in the creation request). 

Then here in the controller, we can tell swagger-autogen about this as a body paramater:

```
     #swagger.parameters['newAlbum'] = {
        in: 'body',
        description: 'Album information to use for creation.',
        required: true,
        type: 'object',
        schema: { $ref: '#/definitions/NewAlbum' }
     }
```

src/controllers/albumController.js
```diff
@@ -4,17 +4,48 @@ const CustomError = require('../errors');
...
 
+// question for Akos - what does minimin means on line 20?
 const getAllAlbums = async (req, res) => {
-  const albums = await Album.find({});
+  /*
+     #swagger.summary = 'Fetch all albums in a database'
+    
+     #swagger.responses[200] = {
+				description: 'Albums  successfully fetched.',
+        schema: { albums: [{ $ref: '#/definitions/Album' }] }
```
akosasante: According to the return value of this function we're returning an object with two fields: `'albums'` which has a list of albums, and `'count'` which is a number

```suggestion
        schema: { albums: [{ $ref: '#/definitions/Album' }], count: 1 }
```

src/controllers/albumController.js
```diff
@@ -4,17 +4,48 @@ const CustomError = require('../errors');
...
+  /*
+     #swagger.summary = 'Fetch an album  by id'
+     #swagger.parameters['id'] = {
+        description: 'Mongo ObjectID of the album to fetch',
+     }
+     #swagger.responses[200] = {
+				description: 'Album successfully fetched.',
+				schema: { $ref: '#/definitions/Album' }
+		 }
+		 #swagger.responses[404] = { description: 'No album with this id wasfound.' }
```
```suggestion
		 #swagger.responses[404] = { description: 'No album with this id was found.' }
```

src/controllers/albumController.js
```diff
@@ -4,17 +4,48 @@ const CustomError = require('../errors');
...
 
 const getSingleAlbum = async (req, res) => {
+  /*
+     #swagger.summary = 'Fetch an album  by id'
+     #swagger.parameters['id'] = {
+        description: 'Mongo ObjectID of the album to fetch',
+     }
+     #swagger.responses[200] = {
+				description: 'Album successfully fetched.',
+				schema: { $ref: '#/definitions/Album' }
```
```suggestion
				schema: { album: { $ref: '#/definitions/Album' } }
```

src/controllers/albumController.js
```diff
@@ -24,6 +55,18 @@ const getSingleAlbum = async (req, res) => {
...
+  /*
+     #swagger.summary = 'Fetch an album  by id and update it'
+     #swagger.parameters['id'] = {
+        description: 'Mongo ObjectID of the album to fetch',
+     }
+     #swagger.responses[200] = {
+				description: 'Album successfully fetched and updated.',
+				schema: { $ref: '#/definitions/Album' }
+		 }
+		 #swagger.responses[404] = { description: 'No album with this id wasfound.' }
```
```suggestion
		 #swagger.responses[404] = { description: 'No album with this id was found.' }
```

src/controllers/albumController.js
```diff
@@ -24,6 +55,18 @@ const getSingleAlbum = async (req, res) => {
...
+
```
akosasante, line 93: For this endpoint, we do `runValidators: true` which means that mongoose will throw a `ValidationError` if the caller tries to pass in any values to update that don't match the schema. Our errorhandler middleware will return a 400 in that case, so we should document that as well

`#swagger.responses[400] = { description: 'validation error' }`

src/controllers/albumController.js
```diff
@@ -24,6 +55,18 @@ const getSingleAlbum = async (req, res) => {
 };
 
 const updateAlbum = async (req, res) => {
+  /*
+     #swagger.summary = 'Fetch an album  by id and update it'
+     #swagger.parameters['id'] = {
+        description: 'Mongo ObjectID of the album to fetch',
+     }
```
akosasante, line 79: In addition to the `id` path parameter, we also have the body parameter that the caller needs to pass in. Swagger-autogen doesn't know exactly what that would look like based on the code so we do need to explicitly define that here. We can probably use the same definition that we will create in `swagger.js` for the createAlbum endpoint since I assume the same values are allowed here for the update?

```
     #swagger.parameters['album properties to update'] = {
      in: 'body',
      description: 'Album information to use for update',
      required: true,
      type: 'object',
      schema: { $ref: '#/definitions/NewAlbum' }
     }
```

src/controllers/albumController.js
```diff
@@ -38,6 +81,14 @@ const updateAlbum = async (req, res) => {
 //will be user to let admin update price of several albums on the frontend
 
 const updatePriceOfAlbums = async (req, res) => {
+  /*
+     #swagger.summary = 'Update prices of albums passed in req.body'
+
+     #swagger.responses[200] = {
+				description: 'Albums prices  were successfully updated.',
+				chema: { albums: [{ $ref: '#/definitions/Album' }] }
```
```suggestion
				schema: { albums: [{ $ref: '#/definitions/Album' }] }
```

src/controllers/albumController.js
```diff
@@ -55,6 +106,15 @@ const updatePriceOfAlbums = async (req, res) => {
 
 //Fetching an album from the database, including all the users that have purchased it
 const getAlbumWithAllUsersWhoPurchasedIt = async (req, res) => {
+  /*
+     #swagger.summary = 'Show all users that purchased this particular album'
+
+     #swagger.responses[200] = {
+				description: 'Album with users who purchased it was fetched  successfully ',
+				chema: { albums: [{ $ref: '#/definitions/Album' }] }
```
akosasante: Probably need to define a schema definition in `swagger.js` for `AlbumWithUsers` so that we can display an example of what this response would look like

```suggestion
				schema: { albums: [{ $ref: '#/definitions/Album' }] }
```

akosasante: Also, according to the repsonse value of this function, we're returning two keys in the object of the response `album` which you have here, and `purchasingUsersCount`

src/controllers/albumController.js
```diff
@@ -4,17 +4,48 @@ const CustomError = require('../errors');
...
   req.body.user = req.user.userId;
   const album = await Album.create(req.body);
   res.status(StatusCodes.CREATED).json({ album });
 };
 
+// question for Akos - what does minimin means on line 20?
 const getAllAlbums = async (req, res) => {
-  const albums = await Album.find({});
+  /*
+     #swagger.summary = 'Fetch all albums in a database'
```
```suggestion
     #swagger.summary = 'Fetch all albums in a database (with price > $0)'
```

src/controllers/albumController.js
```diff
@@ -75,10 +135,20 @@ const getAlbumWithAllUsersWhoPurchasedIt = async (req, res) => {
   });
 };
 
+// question for Akos: how can we test this route in postman? it is set to show 10 results per page and don't know how to see ne 2nd the 3d pages..etcпш
 const getFilteredAlbums = async (req, res) => {
+  /*
+     #swagger.summary = 'Fetch first 10 albums in a database with price more than 0'
```
```suggestion
     #swagger.summary = 'Fetch paginated list of albums with price > 0, with query parameters for sorting and filtering'
```

src/controllers/albumController.js
```diff
@@ -75,10 +135,20 @@ const getAlbumWithAllUsersWhoPurchasedIt = async (req, res) => {
...
+
+     #swagger.responses[200] = {
+				description: 'Filtered by price more than 0 albums were successfully fetched.',
```
```suggestion
				description: 'Filtered by query parameters and with price more than 0 albums were successfully fetched.',
```

src/controllers/albumController.js
```diff
@@ -75,10 +135,20 @@ const getAlbumWithAllUsersWhoPurchasedIt = async (req, res) => {
   });
 };
 
+// question for Akos: how can we test this route in postman? it is set to show 10 results per page and don't know how to see ne 2nd the 3d pages..etcпш
```
akosasante: On line 172, the default page size is set to 10 (`limit`). But you can pass in a larger limit if you want to see more results on the first page. Eg: http://localhost:8000/api/v1/albums/filter?limit=100. If we really wanted to we could write code to have a `limit=infinity` to get all the results, but you can already get all the results from the `getAllAlbums` endpoint.

To view results for subsequent pages, you can use the `offset` query parameter: [https://support.smartbear.com/qacomplete/docs/developer/api/rest/api/reference/paging.html](https://support.smartbear.com/qacomplete/docs/developer/api/rest/api/reference/paging.html)

`limit` = number of results on this page/response

`offset` = number of results from db to skip.

So if we want the first 10 albums we would use the query parameters `limit=10&offset=0`

Then for the second page, we would skip the first 10 albums so: `limit=10&offset=10`

Then for the third page we, would add limit again, to skip the first 20 albums: `limit=10&offset=20`

etc

src/controllers/albumController.js
```diff
@@ -38,6 +81,14 @@ const updateAlbum = async (req, res) => {
 //will be user to let admin update price of several albums on the frontend
 
 const updatePriceOfAlbums = async (req, res) => {
+  /*
+     #swagger.summary = 'Update prices of albums passed in req.body'
+
```
akosasante: We should also define what the expected input is; as swagger-autogen isn't able to easily figure it out from this function's body:

```
	#swagger.parameters['albums to update prices'] = {
		in: 'body',
		description: 'Array of information to use to update album prices',
		required: true,
		type: 'array',
		schema: [{ id: "5f9d7b3b9d9d9d9d9d9d9d9d", price: 9.99 }]
 }
```

src/controllers/albumController.js
```diff
@@ -75,10 +135,20 @@ const getAlbumWithAllUsersWhoPurchasedIt = async (req, res) => {
   });
 };
 
+// question for Akos: how can we test this route in postman? it is set to show 10 results per page and don't know how to see ne 2nd the 3d pages..etcпш
 const getFilteredAlbums = async (req, res) => {
+  /*
+     #swagger.summary = 'Fetch first 10 albums in a database with price more than 0'
+
```
akosasante, line 152: We'll want to describe all of the query parameters that are allowed. And we'll tell swagger-autogen to not try to figure the query paramters on its own. We can also document the default values for the query parameters where applicable

```
     #swagger.autoQuery = false
     #swagger.parameters['limit'] = {
        description: 'Number of albums to fetch',
        default: 10,
        in: 'query'
     }
     #swagger.parameters['offset'] = {
        description: 'Number of albums to skip (use with limit to page through results)',
        default: 0,
        in: 'query'
     }

     #swagger.parameters['order'] = {
        description: 'Order of results',
        in: 'query',
        schema: {
          '@enum': ['asc', 'desc']
        }
     }

     #swagger.parameters['albumName'] = {
        description: 'String to filter results by matching any part of album name (case-insensitive)',
        in: 'query'
     }

     #swagger.parameters['artistName'] = {
        description: 'String to filter results by matching any part of artist name (case-insensitive)',
        in: 'query'
     }
```

For the use of `enum`, the docs say we need to enable OpenApi 3.x.x: [https://github.com/swagger-autogen/swagger-autogen#enums](https://github.com/swagger-autogen/swagger-autogen#enums)

You can do that in `swagger.js` by passing in a vlaue to the `swaggerAutogen` function at the bottom: `swaggerAutogen({ openapi: '3.1.0' })(outputFile, routes, doc);`

src/controllers/albumController.js
```diff
@@ -75,10 +135,20 @@ const getAlbumWithAllUsersWhoPurchasedIt = async (req, res) => {
...
+     #swagger.responses[200] = {
+				description: 'Filtered by price more than 0 albums were successfully fetched.',
+        schema: { albums: [{ $ref: '#/definitions/Album' }] }
```
akosasante: This endpoint also returns a `count` field:

```suggestion
        schema: { albums: [{ $ref: '#/definitions/Album' }], count: 1 }
```

src/controllers/albumController.js
NatachaKey: @akosasante just did it- but can't test it in postman - i log in, create an order and then do GET request to {{MUSICURL}}/albums/64f0777055b548585ca3c18c/listOfUsersWhoPurchasedThisAlbum and get the error that there's no such album
Captura de pantalla 2023-08-31 a las 13 21 39
Captura de pantalla 2023-08-31 a las 13 21 30

src/controllers/albumController.js
NatachaKey: @akosasante just did it- but can't test it in postman - i log in, create an order and then do GET request to {{MUSICURL}}/albums/64f0777055b548585c
Captura de pantalla 2023-08-31 a las 13 21 30
a3c18c
Captura de pantalla 2023-08-31 a las 13 21 39
/listOfUsersWhoPurchasedThisAlbum and get the error that there's no such album

src/controllers/albumController.js
NatachaKey: @akosasante thank you, Akos , for such a great explanation!!!!

src/controllers/albumController.js
```diff
@@ -7,11 +7,35 @@ const createAlbum = async (req, res) => {
...
+            #swagger.parameters['newAlbum'] = {
+        in: 'body',
+        description: 'Album information to use for creation.',
+        required: true,
+        type: 'object',
+        schema: { $ref: '#/definitions/NewAlbum' }
+     } 
+     #swagger.responses[201] = {
+				description: 'Albums was successfully created.',
+        schema: [{ $ref: '#/definitions/Album' }]
```
akosasante: The create endpoint returns a single album, rather than a list that this line is documenting and it will be wrapped in an object with the key `"album"`



so this will need to be:

```suggestion
        schema: { album: { $ref: '#/definitions/Album' } }
```

src/controllers/albumController.js
```diff
@@ -7,11 +7,35 @@ const createAlbum = async (req, res) => {
   req.body.user = req.user.userId;
   const album = await Album.create(req.body);
   res.status(StatusCodes.CREATED).json({ album });
+  /*
+     #swagger.summary = 'Create new album and save it to the database'
+            #swagger.parameters['newAlbum'] = {
```
akosasante, line 13: I found that in this case we might need to add `#swagger.autoBody = false` to make sure swagger-autogen doesn't try to generate the parameter `"user"` based on finding `req.user` in the code.

```suggestion
            #swagger.autoBody = false
            #swagger.parameters['newAlbum'] = {
```

src/controllers/albumController.js
```diff
@@ -7,11 +7,35 @@ const createAlbum = async (req, res) => {
...
+        description: 'Album information to use for creation.',
+        required: true,
+        type: 'object',
+        schema: { $ref: '#/definitions/NewAlbum' }
+     } 
+     #swagger.responses[201] = {
+				description: 'Albums was successfully created.',
+        schema: [{ $ref: '#/definitions/Album' }]
+		 }
+
```
akosasante, line 26: This endpoint also has the chance for a `400` error if any validation fails when saving the album (eg: if caller passes in a price that's negative or something)

`#swagger.responses[400] = { description: 'validation error' }`

akosasante: This endpoint can only be used by admins. Let's add a description indicating that:

`#swagger.description = '**ROLE REQUIRED:** admin'`

src/controllers/albumController.js
```diff
@@ -33,6 +69,26 @@ const updateAlbum = async (req, res) => {
...
+          #swagger.parameters['album properties to update'] = {
+      in: 'body',
+      description: 'Album information to use for update',
+      required: true,
+      type: 'object',
+      schema: { $ref: '#/definitions/NewAlbum' }
+     }
+     #swagger.responses[200] = {
+				description: 'Album successfully fetched and updated.',
+				schema: { $ref: '#/definitions/Album' }
```
akosasante: We return the response wrappd in an object with the key `"album"`



so this should be 

```suggestion
				schema: { album: { $ref: '#/definitions/Album' } }
```

src/controllers/albumController.js
```diff
@@ -51,6 +107,20 @@ const updatePriceOfAlbums = async (req, res) => {
   logger.info(`${bulkWriteResponse.modifiedCount} Albums updated successfully`);
   const updatedAlbums = await Album.find({ _id: { $in: ids } }); // Fetches the updated albums from the database using the _id values in the ids array. //see Implicit $in in mongoose docs
   res.status(StatusCodes.OK).json({ albums: updatedAlbums });
+  /*
+     #swagger.summary = 'Update prices of albums passed in req.body'
+	#swagger.parameters['albums to update prices'] = {
```
akosasante, line 116: swagger-autogen is getting tripped up by the `req.body.map` in this endpoint so it thinks there's some parameter called `"map(update..."`. we can tell it to ignore that and just go with what you've explicitly defined here:

```suggestion
	#swagger.autoBody = false
	#swagger.parameters['albums to update prices'] = {
```

src/controllers/albumController.js
```diff
@@ -73,9 +143,18 @@ const getAlbumWithAllUsersWhoPurchasedIt = async (req, res) => {
     album: usersThatPurchasedThisAlbum,
     purchasingUsersCount: usersThatPurchasedThisAlbum.purchasedByUsers.length,
   });
+  /*
+     #swagger.summary = 'Show all users that purchased this particular album'
+
+     #swagger.responses[200] = {
+				description: 'Album with users who purchased it was fetched  successfully ',
+				schema: { albums: [{ $ref: '#/definitions/AlbumWithUsers' }], purchasingUsersCount: 1 }
```
akosasante: I made a request to **`http://localhost:8000/api/v1/albums/64d2a94c793389a43fc5a8d6/listOfUsersWhoPurchasedThisAlbum`**

and I got:

```
{
  "album": {
    "_id": "64d2a94c793389a43fc5a8d6",
    "artistName": "F",
    "albumName": "AUSTIN",
    "price": 0,
    "image": "https://i.scdn.co/image/ab67616d0000b27371cae34ad5a39bdab78af13e",
    "releaseDate": "2023-07-28T00:00:00.000Z",
    "spotifyUrl": "https://api.spotify.com/v1/albums/6r1lh7fHMB499vGKtIyJLy",
    "averageRating": 0,
    "numOfReviews": 0,
    "__v": 0,
    "createdAt": "2023-08-08T20:45:00.942Z",
    "updatedAt": "2023-08-17T10:10:42.861Z",
    "purchasedByUsers": [
      {
        "_id": "64ef50c8c5551074444547bc",
        "album": "64d2a94c793389a43fc5a8d6",
        "user": {
          "_id": "64d6ca92a15d2e18ab96a2a3",
          "name": "Akos123123",
          "email": "akos123@example.com",
          "role": "user",
          "__v": 0,
          "username": "akos92",
          "id": "64d6ca92a15d2e18ab96a2a3"
        },
        "__v": 0
      }
    ],
    "id": "64d2a94c793389a43fc5a8d6"
  },
  "purchasingUsersCount": 1
}
```

It returns a single album (the one we passed in the id for) and that album can have an array of users. So our response schema here should match:

```suggestion
				schema: { album: { $ref: '#/definitions/AlbumWithUsers' }, purchasingUsersCount: 1 }
```

swagger.js
```diff
@@ -31,6 +32,104 @@ const doc = {
...
+        },
+      ],
+    },
+    Album: {
+      $id: '64d2ae2308a725b72bd5c0cf',
+      $albumName: 'The Dark Side of the Moon',
+      $artistName: 'Pink Floyd',
+      price: 9.99,
+      image: '/uploads/example.jpg',
+      releaseDate: '2022-03-01T00:00:00.000Z',
```
akosasante, line 78: also comes with `createdAt` and `updatedAt` fields:

```suggestion
      releaseDate: '2022-03-01T00:00:00.000Z',
      createdAt: '2022-03-01T00:00:00.000Z',
      updatedAt: '2022-03-01T00:00:00.000Z',
```

swagger.js
```diff
@@ -31,6 +32,104 @@ const doc = {
...
+      category: 'rock',
+      $spotifyUrl: 'https://open.spotify.com/album/4LH4d3cOWNNsVw41Gqt2kv',
+      averageRating: 4.5,
+      numOfReviews: 2,
+    },
+    NewAlbum: {
+      $id: '64d2ae2308a725b72bd5c0cf',
```
akosasante: We can remove this line. When we're creating or updating an album we don't pass in an id of our own. Eg if you're creating a new album you would pass in:

```
{
  albumName: ...,
  artistName: ...,
  etc
}
```

but you wouldn't pass in ID since that would be created by the backend/db

swagger.js
```diff
@@ -31,6 +32,104 @@ const doc = {
...
+      $albumName: 'The Dark Side of the Moon',
+      $artistName: 'Pink Floyd',
+      price: 9.99,
+      image: '/uploads/example.jpg',
+      releaseDate: '2022-03-01T00:00:00.000Z',
+      category: 'rock',
+      $spotifyUrl: 'https://open.spotify.com/album/4LH4d3cOWNNsVw41Gqt2kv',
+    },
+    AlbumsWithUsers: {
```
akosasante: In the albumController file you only refer to `AlbumWithUsers` (no `'s'` on the Album). Since swagger-autogen couldn't find a matching definition in this file, it just defaults to "string" in the docs:



You're using `#/definitions/AlbumWithUsers` in the `getAlbumWithAllUsersWhoPurchasedIt`, which returns an object like:

```
{
  "album": {
    "_id": "64d2a94c793389a43fc5a8d6",
    "artistName": "F",
    "albumName": "AUSTIN",
    "price": 0,
    "image": "https://i.scdn.co/image/ab67616d0000b27371cae34ad5a39bdab78af13e",
    "releaseDate": "2023-07-28T00:00:00.000Z",
    "spotifyUrl": "https://api.spotify.com/v1/albums/6r1lh7fHMB499vGKtIyJLy",
    "averageRating": 0,
    "numOfReviews": 0,
    "__v": 0,
    "createdAt": "2023-08-08T20:45:00.942Z",
    "updatedAt": "2023-08-17T10:10:42.861Z",
    "purchasedByUsers": [
      {
        "_id": "64ef50c8c5551074444547bc",
        "album": "64d2a94c793389a43fc5a8d6",
        "user": {
          "_id": "64d6ca92a15d2e18ab96a2a3",
          "name": "Akos123123",
          "email": "akos123@example.com",
          "role": "user",
          "__v": 0,
          "username": "akos92",
          "id": "64d6ca92a15d2e18ab96a2a3"
        },
        "__v": 0
      }
    ],
    "id": "64d2a94c793389a43fc5a8d6"
  },
  "purchasingUsersCount": 1
}
```

But this code here seems to be defining a _user_ with their list of _purchased albums_. I think that's what we had in the user docs for UserWithAlbums. But this should kind of be the inverse of that.

src/controllers/albumController.js
NatachaKey: @akosasante Akos, I am wondering where do we get this open.spotify.com... url because when we fetch the data in mongo we save https://api.spotify.com/v1/albums/6r1lh7fHMB499vGKtIyJLy
Captura de pantalla 2023-09-01 a las 0 58 29
 the same url we get from api docs https://developer.spotify.com/documentation/web-api/reference/get-an-album

ans so i am wonderying how this url passes the validator in Album.js?   ```
spotifyUrl: {
      type: String,
      required: true,
      validate: {
        validator: function (v) {
          return /^https:\/\/open\.spotify\.com\/album\/[a-zA-Z0-9]+$/.test(v); //ensures that the URL starts with https://open.spotify.com/album/ followed by alphanumeric characters.
        },
        message: 'Invalid Spotify URL',
      },
```

Comment (NatachaKey) [reply src/controllers/albumController.js:21]: // for mi -strart commiting / updating from here

File: swagger.js (hunk 27 above)
Comment (NatachaKey) [reply swagger.js]: @akosasante thank you!!!  will commit it now, just to make sure- to write down the definition correctly- we should first make a request to the url and then convertin to js format and check in models (User and Album) which fields are required and put a $ sign ?

**Comment by NatachaKey** [comment]
@akosasante I was trying to resolve conflicts - but everytime I refresh 2 conflicts in sagger.js are still there and to fix swagger-output-i think it's better to delete in from main and generate a new one? o is it a bad idea?

**Comment by akosasante** [comment]
@NatachaKey Yes that should be fine. It's similar to when there's conflicts in `package-lock.json`. Since that (and `swagger-output.json`)are _generated_ files that are dependent on something else -- `package.json` and `swagger.js` respectively -- it's better to fix any conflicts the source file (`swagger.js`). And then just regenerate the generated file. You can delete and generate a new one (or I think it will also just work if you don't delete it and just go ahead and do `npm run generate-api-docs` after fixing `swagger.js` first.

File: package.json [hunk 28]
```diff
@@ -21,6 +21,7 @@
     "bcryptjs": "^2.4.3",
     "cookie-parser": "^1.4.6",
     "cors": "^2.8.5",
+    "date-fns": "^2.30.0",
```
akosasante: Did this come from merging `main` or something you added?

akosasante (approved):
one question about package.json change but otherwise looks good to me

package.json
NatachaKey: @akosasante oh i will delete it, i added this package because I was not sure how to add a Date() property- didn't know that it should be a string

PR #42, opened 2023-08-29
create Wishlist function 

DYA13:
Modifications to Wishlist endpoint

### DIFF
--- a/src/controllers/wishlistController.js
+++ b/src/controllers/wishlistController.js
@@ -11,19 +11,22 @@ const createWishlist = async (req, res) => {
   if (!existingWishlist) {
     const newWishlist = new Wishlist({
       user: req.user.userId,
-      albums: [albumId],
+      albums: [], // start with an empty wishlist
     });
+
+    if (albumId) {
+      newWishlist.albums.push(albumId);
+    }
+
     await newWishlist.save();
     return res.status(StatusCodes.CREATED).json({ wishlist: newWishlist });
   }
 
-  if (existingWishlist.albums.includes(albumId)) {
-    console.log(`Album ${albumId} is already in the wishlist.`);
-    return res.status(StatusCodes.OK).json({ wishlist: existingWishlist });
+  if (albumId && !existingWishlist.albums.includes(albumId)) {
+    existingWishlist.albums.push(albumId);
+    await existingWishlist.save();
   }
 
-  existingWishlist.albums.push(albumId);
-  await existingWishlist.save();
   return res.status(StatusCodes.OK).json({ wishlist: existingWishlist });
 };
 
--- a/src/controllers/wishlistController.js
+++ b/src/controllers/wishlistController.js
@@ -30,31 +30,20 @@ const createWishlist = async (req, res) => {
 // Add an album to a wishlist
 async function addAlbumToWishlist(req, res) {
   const { wishlist_id, album_id } = req.params;
-  console.log('Received request to add album to wishlist');
   if (
     !mongoose.Types.ObjectId.isValid(wishlist_id) ||
     !mongoose.Types.ObjectId.isValid(album_id)
   ) {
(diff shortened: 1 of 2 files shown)


src/controllers/wishlistController.js
```diff
@@ -41,11 +66,12 @@ async function removeAlbumFromWishlist(req, res) {
   if (!wishlist) {
     return res.status(404).json({ error: 'Wishlist not found' });
   }
-
+  console.log('Album added to wishlist successfully');
```
akosasante:
```suggestion
  console.log('Album removed from wishlist successfully');
```

src/controllers/wishlistController.js
```diff
@@ -1,5 +1,31 @@
 const mongoose = require('mongoose');
 const Wishlist = require('../models/Wishlist');
+const { StatusCodes } = require('http-status-codes');
+
+//Create wishlist
+const createWishlist = async (req, res) => {
+  const { albumId } = req.body; //frontend sends the albumId in the request body
```
akosasante: For creating the wishlist I'm not sure if it makes sense for the frontend to be _required_ to send a single album when first creating the wishlist :thinking: . 

What we may want to do 

*   create a wishlist automatically when user first registers
*   create a wishlist automatically if one doesn't alreay exist when user logs in
*   have an endpoint that allows a user to fetch their own wishlist, and if one doesn't exist it will be created on-the-fly

I'll write tickets to flesh those out a bit more.

In the meanwhile for this ticket, let's just create an empty wishlist (if the user doesn't already have one). Then the frontend can make subsequent calls to the "add album to wishlist" to add items after the wishlist has been successfully created.

DYA13: Thank you Akos, I have updated the create wishlist function, could you please check ? commit 10962af5bdcb4d17c5e054b3207777865192fd48

src/controllers/wishlistController.js
```diff
@@ -1,5 +1,34 @@
...
+
+  if (!existingWishlist) {
+    const newWishlist = new Wishlist({
+      user: req.user.userId,
+      albums: [], // start with an empty wishlist
+    });
+
+    if (albumId) {
+      newWishlist.albums.push(albumId);
+    }
```
akosasante: For now, since we're creating an empty wishlist, let's remove this block

src/controllers/wishlistController.js
```diff
@@ -1,5 +1,34 @@
 const mongoose = require('mongoose');
 const Wishlist = require('../models/Wishlist');
+const { StatusCodes } = require('http-status-codes');
+
+//Create wishlist
+const createWishlist = async (req, res) => {
+  const { albumId } = req.body; //frontend sends the albumId in the request body
```
akosasante: For now, since we're creating an empty wishlist, let's remove this line

src/controllers/wishlistController.js
```diff
@@ -1,5 +1,34 @@
...
+    }
+
+    await newWishlist.save();
+    return res.status(StatusCodes.CREATED).json({ wishlist: newWishlist });
+  }
+
+  if (albumId && !existingWishlist.albums.includes(albumId)) {
+    existingWishlist.albums.push(albumId);
+    await existingWishlist.save();
+  }
```
akosasante: For now, since we're creating an empty wishlist, let's remove this line

src/controllers/wishlistController.js
DYA13: Done

src/controllers/wishlistController.js
DYA13: Done

src/controllers/wishlistController.js
DYA13: Done

DYA13: Checked in Postman, it works 👍

akosasante (approved):
:ship:

PR #43, opened 2023-08-31
Stripe PaymentIntent is fixed

mvstoyan:


src/controllers/orderController.js
```diff
@@ -27,6 +27,11 @@ const createOrder = async (req, res) => {
     amount: total * 100, // Stripe requires amount in cents
     currency: 'usd',
     description: `Order #${order._id}`,
+    metadata: {
+      userId: req.user.userId,
+      orderId: order._id, // key-values
+      totalQuantity: orderItems.length, // the total number of albums purchased in order
```
akosasante: If we have orderItems like:

```javascript
[
 { album: ..., quantity: 3 }
 { album: ..., quantity: 2 }
]
```

The user has bought 2 albums, but a total quantity of items of 5. 

Maybe we can include both values here in the metadata in case they are of use later.

akosasante (commented):
Thanks for picking this up quickly, just one small note

src/controllers/orderController.js
mvstoyan: screen4
console.log("Total Quantity:", orderItems.reduce((total, item) => total + item.quantity, 0));
console.log("Total Albums:", orderItems.length);  
I trying to check on console.log, and test in postman, i think its work.

akosasante (approved):

PR #44, opened 2023-08-31
create a PurchasedAlbum entry

mvstoyan:


akosasante (approved):
Thanks for the quick add!

PR #45, opened 2023-08-31
Socket

DYA13:


package.json
akosasante, line 1: Make sure to do a `git pull origin main` followed by `npm install` (or use "Update Branch" if it shows up here in the Github PR) before merging so that we're not accidentally deleting `stripe` as it looks like that's what's currently being shown.

DYA13:
It seems there are some problems with the endpoints , it doesn't let me run all checks successfully
● GET /api/v1/users/:user_id endpoint › should return a valid user without the password field if found
  ● GET /api/v1/users/:user_id endpoint › should return a 404 status if user is not found

DYA13:
Akos , I have noticed that from yesterday after last pulling from main my terminal becomes crazy when I run npm run dev

https://github.com/Code-the-Dream-School/dd-prac-team4-back/assets/88336329/52a227af-0db4-4976-b8e2-51bdb5c990b2

akosasante:
@DYA13 For myself, when I was seeing those errors in my local dev it was because I hadn't added all the latest values to my `.env` file: [https://codethedream.slack.com/archives/C05G08542H3/p1693435063248899](https://codethedream.slack.com/archives/C05G08542H3/p1693435063248899)

Specifically, it's looking for `DEV_ORDER_EXPIRY_DURATION`, but because it didn't find one it was passing a null value where the code was expecting an integer/date.

akosasante:
> It seems there are some problems with the endpoints , it doesn't let me run all checks successfully ● GET /api/v1/users/:user\_id endpoint › should return a valid user without the password field if found ● GET /api/v1/users/:user\_id endpoint › should return a 404 status if user is not found

Sorry, I think there's a mistake in my instructions. When exporting from `expressServer.js` we should be exporting the HTTP server (which has socket stuff connected) rather than the socket server. This is currently causing the app to start up with none of the REST API endpoints available.

akosasante:
Sorry, accidentally clicked "Close PR" :facepalm:

src/expressServer.js
```diff
@@ -131,4 +131,26 @@ app.use(express.static(__dirname + '/public'));
...
+
+  socket.on('msg_from_client', function (from, msg) {
+    console.log('Message is ' + from, msg);
+  });
+  socket.on('disconnect', function () {
+    console.log('Disconnected');
+  });
+});
+
+module.exports = { app: socketServer, connectDB };
```
akosasante:
```suggestion
module.exports = { app: server, connectDB };
```

package.json
DYA13: Done 👍

akosasante (approved):
Looks great; I was able to test in Postman sending messages and listening for events :+1:

PR #46, opened 2023-09-01
swagger orders

DYA13:
Akos could you please check create order swagger code , I don't manage to adjust it in order I could be able to insert necessary fields to make it work ? Thank you very much


akosasante:
FYI, Looking at your latest commit message, rather than doing `npm run format:check -- --write` you can use the script we have in `package.json` for formatting and writing the changes to file:  `npm run format`

Github CI has to do `npm run format:check` to just do the check but not write the changes, because it runs on a separate machine so any changes it _did_ make would only be temporary anyway.

src/controllers/orderController.js
```diff
@@ -57,11 +57,68 @@ const createOrder = async (req, res) => {
...
 
 const getAllOrders = async (req, res) => {
   const orders = await Order.find({});
   res.status(StatusCodes.OK).json({ orders, count: orders.length });
+  /*
+     #swagger.summary = 'Fetch all orders in a database'
+     #swagger.description = '**ROLE REQUIRED:** admin'
+     #swagger.responses[200] = {
+				description: 'Orders successfully fetched.',
+        schema: { $ref: '#/definitions/Oder' }
```
akosasante:
```suggestion
        schema: { $ref: '#/definitions/Order' }
```

src/controllers/orderController.js
```diff
@@ -57,11 +57,68 @@ const createOrder = async (req, res) => {
   res
     .status(StatusCodes.CREATED)
     .json({ clientSecret: paymentIntent.client_secret, order });
+  /*
+#swagger.summary = 'Create a new order and process payment'
+#swagger.description = 'Creates a new order and processes payment using Stripe.'
+#swagger.tags = ['Orders']
+#swagger.parameters['body'] = {
```
akosasante: For swagger-autogen comments, we don't use the full Swagger/OpenAPI syntax to define objects. Rather, the library wants us to just describe the actual shape of the object with example data, and it will infer the type and all that.

src/controllers/orderController.js
```diff
@@ -57,11 +57,68 @@ const createOrder = async (req, res) => {
...
+      tax: {
+        type: 'number',
+        description: 'Tax amount',
+      },
+      total: {
+        type: 'number',
+        description: 'Total amount',
+      },
+    },
+  },
```
akosasante:
```suggestion
  schema: {
    $orderItems: [{
      album: "5f7f1f7a5c3f2b2d9c1b0b9d",
      quantity: 2
    }],
    $subtotal: 9.99,
    $tax: 0.75,
    $total: 10.74
  },
```

akosasante (commented):
Anything that you refer to by `$ref: '#/definitions/...'` needs to exist in the `definitions` block of the swagger-autogen document defined in `swagger.js`

[https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/swagger-orders/swagger.js#L15](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/swagger-orders/swagger.js#L15)

In there you want to basically paste in an example of what an expected response would look like with sample fields. Swagger-autogen will use that information to build the documented response example and model. 

You can do the same for the body parameter rather than defining it in-line here in the controller.

akosasante:
https://github.com/swagger-autogen/swagger-autogen#schema-and-definitions

src/controllers/orderController.js
DYA13: Akos , I have three problems in swagger-output.json in lines 531, 539, 565. As for me , it seems the same like other schemas. Could you please check? thank you

swagger.js
```diff
@@ -32,6 +32,53 @@ const doc = {
...
+          {
+            album: '5f7f1f7a5c3f2b2d9c1b0b9d',
+            quantity: 2,
+          },
+        ],
+        subtotal: 9.99,
+        tax: 0.75,
+        total: 10.74,
+      },
+    },
```
akosasante: When I make a call to the GET /orders/{id} endpoint I get a response like this:



So we'll want to make sure our documented example response has the same shape:

```suggestion
      Order: {
        $_id: 'idabc123456',
        $user: {
          $id: 'idabc123456',
          $name: 'John Doe',
          $username: 'johndoe',
          $email: 'john@example.com',
          $role: { '@enum': ['user', 'admin'] }
        },
        $orderStatus: { '@enum': ['pending', 'payment_successful', 'payment_failed', 'cancelled', 'complete'], default: 'pending' },
        orderItems: [
          {
            $album: '5f7f1f7a5c3f2b2d9c1b0b9d',
            $quantity: 2,
            $_id: 'idabc12345'
          },
        ],
        $subtotal: 9.99,
        $tax: 0.75,
        $total: 10.74,
        $createdAt: '2023-08-08T20:45:00.942Z',
        $updatedAt: '2023-08-08T20:45:00.942Z',
        paymentIntentId: 'pi_abc123456'
      },
    },
```

The `$` in front of a field indicates that this is a required field or in the case of a response body, a field that is _always_ populated and returned.

The `@enum` field allows us to document that there is a finite set of options for this field (the `@` sign in front indicates that we're using raw Swagger/OpenAPI syntax here)

The resulting documentation should look like:

Screenshot 2023-09-02 at 10 46 47 PM

swagger.js
```diff
@@ -32,6 +32,53 @@ const doc = {
         },
       },
     },
+    definitions: {
```
akosasante: You have a `definitions` object here nested under the original definitions object on line 15, which won't be parsed correctly by swagger-autogen.

All objects, regardless of which controller uses them, should be part of the same `definitions` object.

swagger.js
```diff
@@ -32,6 +32,53 @@ const doc = {
...
+              quantity: 3,
+            },
+          ],
+          subtotal: 29.97,
+          tax: 2.5,
+          total: 32.47,
+        },
+      ],
+      count: 2,
+    },
```
akosasante: When I make a call to the `GET /orders` endpoint, I get a response like this:



It is an object with two keys: `orders` and `count`. So we can remove line 52 since there is no top level `id` field.

As well, instead of having to write what an order looks like again, we can potentially just use: `orders: [ { $ref: '#/definitions/Order' } ],`

The only difference might be that the `user` field here is just the ID, vs in the GET endpoint it's the full user :thinking:  in which case maybe we will just have to rewrite the object here after all.

src/controllers/orderController.js
```diff
@@ -57,11 +57,36 @@ const createOrder = async (req, res) => {
...
+    in: 'body',
+    description: 'Order information including order items, subtotal, tax, and total',
+    required: true,
+    schema: {
+      $ref: '#/definitions/Order' 
+    },
+  }
+  #swagger.responses[201] = {
+    description: 'Order created successfully',
+    schema: { $ref: '#/definitions/Order' }
```
akosasante: The endpoint returns an object like this: `{ "order": < the order object we put in the definitions block in swagger.js> }`. We can make sure the docs match by matching that shape here:

```suggestion
    schema: { order: { $ref: '#/definitions/Order' } }
```

src/controllers/orderController.js
```diff
@@ -92,6 +128,17 @@ const deleteOrder = async (req, res) => {
...
 
   res.status(StatusCodes.OK).json({ msg: 'Success! Order was deleted' });
+  /*
+  #swagger.summary = 'Delete an order by id'
+  #swagger.description = '**ROLE REQUIRED:** user'
+  #swagger.parameters['id'] = {
+    description: 'Mongo ObjectID of the order to delete',
+  }
+  #swagger.responses[200] = {
+    description: 'The order was successfully deleted.',
```
akosasante, line 138: We can also include an example of what the response body would look like:

```suggestion
    description: 'The order was successfully deleted.',
    schema: { msg: 'Success! Order was deleted' }
```

src/controllers/orderController.js
```diff
@@ -92,6 +128,17 @@ const deleteOrder = async (req, res) => {
   await Order.findByIdAndDelete(orderId); // or more simply can just call delete on the documnent we've already fetched:  await order.remove()
 
   res.status(StatusCodes.OK).json({ msg: 'Success! Order was deleted' });
+  /*
+  #swagger.summary = 'Delete an order by id'
+  #swagger.description = '**ROLE REQUIRED:** user'
```
akosasante: This endpoint will also work for admins

```suggestion
  #swagger.description = '**ROLE REQUIRED:** user or admin'
```

src/controllers/orderController.js
```diff
@@ -57,11 +57,36 @@ const createOrder = async (req, res) => {
...
+  /*
+  #swagger.summary = 'Create a new order and process payment'
+  #swagger.description = 'Creates a new order and processes payment using Stripe.'
+  #swagger.tags = ['Orders']
+  #swagger.parameters['body'] = {
+    in: 'body',
+    description: 'Order information including order items, subtotal, tax, and total',
+    required: true,
+    schema: {
+      $ref: '#/definitions/Order'
```
akosasante: The schema that we expect the client to give us in the body request isn't actually exactly the same as the schema that we return in the response:



For example, we don't expect the client to pass in the \_id, paymentIntentId, createdAt, updatedAt, or orderStatus (these are all generated by the backend/controller code), and the user isn't part of the expected request body since the controller will just grab that from the cookie. 

We should create a separate object in the `definitions` block of the `swagger.js` file that describes the shape of the "InputOrder" or "NewOrder" or however you prefer to name it. You can see that we did a similar thing in the `main` branch with `Album` object used for responses and `NewAlbum` object used for the request body: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/swagger.js#L73-L94

DYA13: Akos, I have added new order , commit 913f64ac1afde6be16a90bd348ae1bcb4b4ba7b8

swagger.js
```diff
@@ -32,6 +32,86 @@ const doc = {
...
+          quantity: {
+            type: 'integer',
+            example: 2,
+          },
+        },
+      },
+    },
+
+    OrderList: {
+      $id: '789012',
```
akosasante: Response should not have an id at top level, this line can be removed

swagger.js
```diff
@@ -32,6 +32,86 @@ const doc = {
...
+      orders: [
+        {
+          user: 'Jane Smith',
```
akosasante: Response returns a user id rather than a user's name



```suggestion
          user: '64ef5e90086e1c19b888c72c',
```

swagger.js
```diff
@@ -32,6 +32,86 @@ const doc = {
...
+          user: 'Bob Johnson',
+          orderItems: [
+            {
+              album: '5f7f1f7a5c3f2b2d9c1b0b9e',
+              quantity: 3,
+            },
+          ],
+          subtotal: 29.97,
+          tax: 2.5,
+          total: 32.47,
```
akosasante: Missing a few properties that are returned in the response:



```suggestion
          total: 16.24,
          orderStatus: { '@enum': ['pending', 'payment_successful', 'payment_failed', 'cancelled', 'complete'] },
          createdAt: '2023-08-08T20:45:00.942Z',
          updatedAt: '2023-08-08T20:45:00.942Z',
          paymentIntentId: 'pi_12345abcdef'
        },
        {
          user: 'Bob Johnson',
          orderItems: [
            {
              album: '5f7f1f7a5c3f2b2d9c1b0b9e',
              quantity: 3,
            },
          ],
          subtotal: 29.97,
          tax: 2.5,
          total: 32.47,
          orderStatus: { '@enum': ['pending', 'payment_successful', 'payment_failed', 'cancelled', 'complete'] },
          createdAt: '2023-08-08T20:45:00.942Z',
          updatedAt: '2023-08-08T20:45:00.942Z',
          paymentIntentId: 'pi_12345abcdef'
```

swagger.js
akosasante: This second definitions block still needs to be removed; so that all the Order objects are in the same existing block that starts on line 15 instead of creating a second one here.

src/controllers/orderController.js
akosasante: :+1: I see new `NewOrder` definition in swagger.js. Now just this line needs to be edited to refer to `'#/definitions/NewOrder` as well.

src/controllers/orderController.js
```diff
@@ -79,6 +104,17 @@ const getSingleOrder = async (req, res) => {
...
 
   res.status(StatusCodes.OK).json({ order });
+  /*
+  #swagger.summary = 'Fetch an order by id'
+  #swagger.parameters['id'] = {
+    description: 'Mongo ObjectID of the order to fetch',
+  }
+  #swagger.responses[200] = {
+    description: 'Order successfully fetched.',
+    schema: { $ref: '#/definitions/Order' }
```
akosasante: when the response is returned; it's nested inside an object like this: `{ "order": <order object as defined in the schema> }`. But out swagger doc here just says it will return `{ <order object> }` without that top-level key:

Real response:



vs. documented response:



so we need to make the documentation match by wrapping our schema:

```suggestion
    schema: { order: { $ref: '#/definitions/Order' } }
```

swagger.js
```diff
@@ -32,6 +32,86 @@ const doc = {
...
+            type: 'string',
+            example: 'Album ID or Name',
+          },
+          quantity: {
+            type: 'integer',
+            example: 2,
+          },
+        },
+      },
+    },
```
akosasante: In swagger-autogen, by default we are expected to describe objects/responses/requests as their literal form / or an example of what they would look like, rather than using raw Swagger/OpenAPI notation (though swagger-autogen does have an option for providing raw syntax in some cases where needed).

So right now, this block is saying that when creating a new order, we have to pass in fields like "type" and "properties" which is incorrect:

Rather, we should be writing this as exactly how we expect the frontend or any other API client to pass information to this endpoint:

```suggestion
      NewOrder: {
        $subtotal: 100,
        $tax: 0.15,
        $total: 115,
        $orderItems: [
          { album: "abcdef12345", quantity: 2 }
        ]
      },
    },
```

swagger.js
DYA13: Akos, I am sorry, I didn't catch what I should delete as I was working on the code  orders on lines from  35 till 113 , could you please explain? Thank you

src/controllers/orderController.js
DYA13: Akos I have done some changes, I am not sure if it is correct , could you check please? commit 78bfd32e45be226fb99a1b4f7ae1245e15dcfeed

swagger.js
DYA13: Done , commit 5cfda4d4f2cc2641482a265581a815de4021aaf2

swagger.js
```diff
@@ -32,6 +32,93 @@ const doc = {
...
+          subtotal: 14.99,
+          tax: 1.25,
+          total: 16.24,
+          orderStatus: { '@enum': ['pending', 'payment_successful', 'payment_failed', 'cancelled', 'complete'] },
+          createdAt: '2023-08-08T20:45:00.942Z',
+          updatedAt: '2023-08-08T20:45:00.942Z',
+          paymentIntentId: 'pi_12345abcdef'
+        },
+        {
+          user: 'Bob Johnson',
```
akosasante:
```suggestion
          user: '64ef5e90086e1c19b888c72c',
```

swagger.js
akosasante: Currently you have this variable: 

```javascript
const doc = {
  info: {},
  host: "....",
  definitions: {
    PasswordlessUser: {},
    definitions: {
      Order: {},
      NewOrder: {}
    }
  }
}
```

that second "definitions" field shouldn't be there. That's why currently swagger can't find the `Order` definition, because you have it nested under a second `definitions` block.

It should just be:

```
const doc = {
  info: {},
  host: "...",
  definitions: {
    PasswordlessUser: {},
    Order: {},
    OrderList: {},
    ...etc
  }
}
```

src/controllers/orderController.js
akosasante: You should change line 69 here to refer to: `$ref: '#/definitions/NewOrder'`

swagger.js
DYA13: Thank you , done , commit 58c655b562c8dfe5add91ccfecd671968c917aad

src/controllers/orderController.js
DYA13: Thank you, done, commit 85a91c42a5311e2e4542c65fc6bbd4f3fdf1fd59

DYA13:
I  have still three problems , but seems it is not related to orders 
image

swagger-output.json
akosasante, line 1: Have you run `npm run generate-api-docs` ? I'm not seeing the other order docs showing up here in the Github diff for the swagger-output.json.

akosasante (commented):
Locally I'm not seeing the issues you've screenshotted with `swagger-output.json` after running the generate api docs script. If the content for those endpoints is showing up correctly in the API docs page then I wouldn't worry too much about the vscode warnings for now.

Also I think somewhere in your recent commits (maybe from merging in main) we lost the changes you made to `swagger.js` where the various response/request shapes were defined.

swagger-output.json
DYA13: yes I have done , 
image

akosasante:
> I have still three problems , but seems it is not related to orders 
> 
> image

I'm not able to see exactly why it's showing you those as roblems, but everythign looks to be fine on my end for the Albums endpoints. Does that output file have a `components: {...}` block?

akosasante (commented):
@DYA13  I'm still not seeing the `swagger.js` file that was here before that would define things like `NewOrder` and `Order.js`

DYA13:
Akos , sorry , I am a bit messed with conflicts. I tried to revert changes. I am not pretty sure if I have restored all necessary files and changes. Could you please check?🙏 🥹

swagger.js
```diff
@@ -32,100 +32,135 @@ const doc = {
...
-      $role: { '@enum': ['user', 'admin'] },
-      profileImage: {
-        url: 'https://example.com/profile-image.jpg',
-      },
-      creditCardInfo: {
-        hashedNumber: '1234567890',
-        expiry: '12/24',
-        preferredPaymentOption: {
-          '@enum': ['credit card', 'paypal', 'google pay'],
+    definitions: {
```
akosasante: @DYA13 we still need to delete line 35 and line 70 and 77 so that we don't have this extra `definitions` block in here surrounding our Order object.

https://github.com/Code-the-Dream-School/dd-prac-team4-back/assets/9094098/d1fb5a58-7623-422d-8393-7dceab775685

DYA13: Akos thank you for detailed explanation ❤️

DYA13: I have added New Order, at the moment it is like an object, as it was before. If I need to change for an array, tell me please.

swagger.js
```diff
@@ -32,12 +32,64 @@ const doc = {
...
+      $subtotal: 100,
+      $tax: 0.15,
+      $total: 115,
+      $orderItems: [{ album: 'abcdef12345', quantity: 2 }],
+    },
+
+    OrderList: {
+      orders: [
+        {
+          user: 'Jane Smith',
```
akosasante: on the `OrderList` used in the `GET /orders` endpoint, the `user` field is not populated so we actually just get back the user's id not their name. Here's an example response:

```json
{
  "orders": [
    {
      "_id": "64faeb31e00e8a4c21dfec94",
      "user": "64d44bc8337399ccf9ad7e4d",
      "orderStatus": "complete",
      "tax": 0.4,
      "subtotal": 100,
      "total": 140,
      "orderItems": [
        {
          "album": "64d36bd25c5dec8e599bee05",
          "quantity": 1,
          "_id": "64faeb31e00e8a4c21dfec95"
        },
        {
          "album": "64d2ae2308a725b72bd5c0da",
          "quantity": 4,
          "_id": "64faeb31e00e8a4c21dfec96"
        },
        {
          "album": "64d2ae2308a725b72bd5c0db",
          "quantity": 2,
          "_id": "64faeb31e00e8a4c21dfec97"
        }
      ],
      "createdAt": "2023-09-08T09:36:49.933Z",
      "updatedAt": "2023-09-08T09:36:49.933Z",
      "__v": 0
    },
    {
      "_id": "64fc39e8d4e3a9f350fa6b81",
      "user": "64d44bc8337399ccf9ad7e4d",
      "orderStatus": "complete",
      "tax": 0.4,
      "subtotal": 100,
      "total": 140,
      "orderItems": [
        {
          "album": "64d36bd25c5dec8e599bee05",
          "quantity": 1,
          "_id": "64fc39e8d4e3a9f350fa6b82"
        },
        {
          "album": "64d2ae2308a725b72bd5c0da",
          "quantity": 4,
          "_id": "64fc39e8d4e3a9f350fa6b83"
        },
        {
          "album": "64d2ae2308a725b72bd5c0db",
          "quantity": 2,
          "_id": "64fc39e8d4e3a9f350fa6b84"
        }
      ],
      "createdAt": "2023-09-09T09:24:56.384Z",
      "updatedAt": "2023-09-09T09:24:56.384Z",
      "__v": 0
    },
    {
      "_id": "64fc3a61becbd34df27e3b59",
      "user": "64d44bc8337399ccf9ad7e4d",
      "orderStatus": "complete",
      "tax": 0.4,
      "subtotal": 100,
      "total": 140,
      "orderItems": [
        {
          "album": "64d36bd25c5dec8e599bee05",
          "quantity": 8,
          "_id": "64fc3a61becbd34df27e3b5a"
        },
        {
          "album": "64d2ae2308a725b72bd5c0da",
          "quantity": 84,
          "_id": "64fc3a61becbd34df27e3b5b"
        },
        {
          "album": "64d2ae2308a725b72bd5c0db",
          "quantity": 82,
          "_id": "64fc3a61becbd34df27e3b5c"
        }
      ],
      "createdAt": "2023-09-09T09:26:57.926Z",
      "updatedAt": "2023-09-09T09:26:57.926Z",
      "__v": 0
    },
    {
      "_id": "64fcdc35d7349bd5eff622e2",
      "user": "64ef5e90086e1c19b888c72c",
      "orderStatus": "payment_successful",
      "tax": 0.15,
      "subtotal": 500,
      "total": 550,
      "orderItems": [
        {
          "album": "64d36b781a3159bca4e17193",
          "quantity": 5,
          "_id": "64fcdc35d7349bd5eff622e3"
        }
      ],
      "createdAt": "2023-09-09T20:57:25.668Z",
      "updatedAt": "2023-09-09T20:57:33.322Z",
      "__v": 0,
      "paymentIntentId": "pi_3NoYDaKnb0YvaiB31RGQGLwz"
    },
    {
      "_id": "64fcdc761543ae4a89fcaa52",
      "user": "64ef5e90086e1c19b888c72c",
      "orderStatus": "payment_successful",
      "tax": 0.15,
      "subtotal": 500,
      "total": 550,
      "orderItems": [
        {
          "album": "64d36b781a3159bca4e17193",
          "quantity": 5,
          "_id": "64fcdc761543ae4a89fcaa53"
        }
      ],
      "createdAt": "2023-09-09T20:58:30.889Z",
      "updatedAt": "2023-09-09T20:58:37.257Z",
      "__v": 0,
      "paymentIntentId": "pi_3NoYEdKnb0YvaiB31bZI5Twx"
    },
    {
      "_id": "64fcdc8f4331a018d0c5aa3b",
      "user": "64ef5e90086e1c19b888c72c",
      "orderStatus": "payment_successful",
      "tax": 0.15,
      "subtotal": 500,
      "total": 550,
      "orderItems": [
        {
          "album": "64d36b781a3159bca4e17193",
          "quantity": 5,
          "_id": "64fcdc8f4331a018d0c5aa3c"
        }
      ],
      "createdAt": "2023-09-09T20:58:55.443Z",
      "updatedAt": "2023-09-09T21:03:33.139Z",
      "__v": 0,
      "paymentIntentId": "pi_3NoYF2Knb0YvaiB314UGm5o1"
    },
    {
      "_id": "64fcddc6d3ab1a4d7a54ac1c",
      "user": "64ef5e90086e1c19b888c72c",
      "orderStatus": "payment_successful",
      "tax": 0.15,
      "subtotal": 500,
      "total": 550,
      "orderItems": [
        {
          "album": "64d36b781a3159bca4e17193",
          "quantity": 5,
          "_id": "64fcddc6d3ab1a4d7a54ac1d"
        }
      ],
      "createdAt": "2023-09-09T21:04:06.154Z",
      "updatedAt": "2023-09-09T21:16:15.528Z",
      "__v": 0,
      "paymentIntentId": "pi_3NoYK2Knb0YvaiB30dlLP6TG"
    }
  ],
  "count": 7
}
```

so weshould update the definition here to match that:

```suggestion
          user: '64d44bc8337399ccf9ad7e4d',
```

swagger.js
```diff
@@ -32,12 +32,64 @@ const doc = {
...
+          orderItems: [
+            {
+              album: '5f7f1f7a5c3f2b2d9c1b0b9d',
+              quantity: 1,
+            },
+          ],
+          subtotal: 14.99,
+          tax: 1.25,
+          total: 16.24,
+        },
```
akosasante: We also can have a few more fields defined here: `orderStatus`, `_id`, `paymentIntentId`, `createdAt`, and `updatedAt`:

```suggestion
          total: 16.24,
          createdAt: '2023-08-08T20:45:00.942Z',
          updatedAt: '2023-08-08T20:45:00.942Z',
          _id: '64ef50c8c5551074444547bc',
          orderStatus: { '@enum': ['pending', 'payment_successful', 'payment_failed', 'cancelled', 'complete'] },
          paymentIntentId: 'pi_abc123456',
        },
```

swagger.js
```diff
@@ -32,12 +32,64 @@ const doc = {
...
+
+    NewOrder: {
+      $subtotal: 100,
+      $tax: 0.15,
+      $total: 115,
+      $orderItems: [{ album: 'abcdef12345', quantity: 2 }],
+    },
+
+    OrderList: {
+      orders: [
```
akosasante, line 80: In addition to the `orders` list, the `getAllOrders` endpoint also returns a field `count` [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/controllers/orderController.js#L64](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/controllers/orderController.js#L64)



We should include that in our definition here:

```suggestion
      count: 1,
      orders: [
```

akosasante (approved):
Looks good; few remaining small bits but good to merge!

PR #48, opened 2023-09-02
Create an endpoint that sends a reset password email PS 162

DYA13:
Akos, I have checked the code many times  but I don't manage why I receive error 404 in Postman. Could you please help to understand what I have done with errors? Thank you


src/controllers/authController.js
```diff
@@ -65,8 +67,43 @@ const logout = async (req, res) => {
...
+  user.passwordResetExpiresOn = new Date(resetExpires);
+  await user.save();
+
+  // Send reset password email
+  const emailTemplate = 'forgotPassword'; // to adjust this based on our email template name
+  const recipient = user.email;
+  const localVariables = {
+    resetToken,
+  };
+  await emailSender.send(emailTemplate, recipient, localVariables);
```
akosasante: This function doesn't exist in your `emailSender`. Should be:

```suggestion
  await emailSender.sendForgotPasswordEmail(emailTemplate, recipient, localVariables);
```

src/email/sender.js
```diff
@@ -0,0 +1,36 @@
...
+
+  try {
+    await transporter.sendMail(mailOptions);
+    console.log('Password reset email sent successfully');
+  } catch (error) {
+    console.error('Error sending password reset email:', error);
+  }
+}
+
+module.exports = sendForgotPasswordEmail;
```
akosasante: Right now we're making the default value of this export `sendForgotPasswordEmail` function. 

That means if you import it in a different file:  
 

`const emailSender = require('./email/sender')`

Then the value of `emailSender` is actually just the function `sendForgotPasswordEmail`. 

We will eventually be adding other sending functions to this file so we shouldn't make this just a single value, but instead export an object, so that the other future functions can be exported as well:

```suggestion
module.exports = { sendForgotPasswordEmail };
```

Then this can be used in other files like this:

`const emailSender = require('./email/sender')` (where `emailSender` will now be an object. And you can call this forgot password function by grabbing the value at this key: `emailSender.sendForgotPasswordEmail(...)`)

or like this:

`const {sendForgotPasswordEmail} = require('./email/sender')` . Here we're using object destructuring to get just the field we want out of the object. Then we can use like this: `sendForgotPasswordEmail(...)`

src/email/templates/html.ejs
```diff
@@ -0,0 +1,26 @@
...
+    <div class="email-container">
+      <h2>Password Reset Request</h2>
+      <p>Hello,</p>
+      <p>
+        We received a request to reset your password. Click the link below to
+        reset your password:
+      </p>
+      <p><a href="<%= resetLink %>">Reset Password</a></p>
+      <% /* The <%= resetLink %> placeholder should be replaced with the actual
+      URL later to the reset password page.*/ %>
```
akosasante: To do a comment in EJS, you need to use `<%# my comment %>` or `<%/* comment */%>`

[https://ejs.co/#docs:~:text=%3C%25%23%20Comment%20tag%2C%20no%20execution%2C%20no%20output](https://ejs.co/#docs:~:text=%3C%25%23%20Comment%20tag%2C%20no%20execution%2C%20no%20output)

Though I think even with that EJS will get a bit confused here about where the starting and ending `<%` are because you have that inner `<%= resetLink %>` inside the comment. So you may want to break that up a bit:

```suggestion
      <%/* The <\% resetLink %\> placeholder should be replaced with the actual
      URL later to the reset password page. */%>
```

src/email/sender.js
```diff
@@ -0,0 +1,36 @@
+const nodemailer = require('nodemailer');
+const path = require('path');
+const ejs = require('ejs');
+require('dotenv').config();
+
+async function sendForgotPasswordEmail(toEmail, resetToken) {
+  const transporter = nodemailer.createTransport({
+    service: 'gmail',
```
akosasante: This should all already be set up in the `mailing/sender.js` file: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/mailing/sender.js#L4-L10](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/mailing/sender.js#L4-L10)

We should be writing this new code in that file and just adding this new function. The function should not need to recreate a transporter object and can just use the one created at top-level of the `sender.js` file

akosasante:
I'm not getting a 404 but rather a 500. Left some notes in the review comments on how to fix

src/email/sender.js
DYA13: I have set the code from  email/sender.js in mailing.js . Should I delete  sender.js in email folder?

DYA13: commit ef22adbf4473cd09a28dba0a7a1ae5139cbdae02

src/mailing/sender.js
akosasante, line 1: Is it expected that you have the same content in `src/email/sender.js` and `src/mailing/sender.js` ? We can probably remove the `email` folder since this `mailing` one already had existing code.

src/controllers/authController.js
```diff
@@ -65,8 +67,47 @@ const logout = async (req, res) => {
...
+  await user.save();
+
+  // Send reset password email
+  const emailTemplate = 'forgotPassword'; // to adjust this based on our email template name
+  const recipient = user.email;
+  const localVariables = {
+    resetToken,
+  };
+  await emailSender.sendForgotPasswordEmail(
+    emailTemplate,
```
akosasante: I wouldn't expect to need to pass in the template name to the `sendForgotPasswordEmail` function. Will different callers be passing different email templates to that function? I think it will always use the same template, so that should be just a static variable defined inside that function itself rather than passed in from the controller

src/mailing/sender.js
```diff
@@ -31,4 +34,34 @@ async function sendTestEmail(recipient) {
...
-module.exports = sendTestEmail;
+async function sendForgotPasswordEmail(toEmail, resetToken) {
+  const transporter = nodemailer.createTransport({
+    service: 'gmail',
+    auth: {
+      user: process.env.EMAIL_USER,
+      pass: process.env.EMAIL_PASSWORD,
+    },
+  });
+
```
akosasante: We don't need to redefine the transporter each time this function is called and instead should use the one defined at the top level of this module/file (line 7).

src/mailing/sender.js
DYA13: Should I delete the folder templates in folder email as well ?

src/controllers/authController.js
DYA13: Could you check please ? commit  346fdf55e289cc9a37f919fedb7f46d0452b0ade

src/mailing/sender.js
DYA13: As far as I understand we want to delete this sender.js in email folder

src/mailing/sender.js
akosasante: Rather we should move that content over to the `mailing` folder and use it there.

src/mailing/sender.js
akosasante: Yes we'll delete the `email/sender.js`. Here in the `mailing/sender.js` we can use the outlook `transport` variable defined earlier in the file to send an email in this function from the [`codethedream.practicum.team4@outlook.com`](mailto:codethedream.practicum.team4@outlook.com) email rather than defining a new gmail transporter (unless that's needed for a particular reason).

src/mailing/sender.js
DYA13: Done , commit 62dcc9e089bdc1bb02458a59be648574de39b1f7

src/mailing/sender.js
DYA13: email/sender.js deleted 62dcc9e089bdc1bb02458a59be648574de39b1f7

DYA13:
Now seems to be ok, the only thing I am doubt about is this line 2023-09-06 22:42:34 [INFO]: GET /api/v1/users/nonexistentUserId 404 2.271 ms - 51
image

src/controllers/authController.js
```diff
@@ -65,8 +67,43 @@ const logout = async (req, res) => {
...
+  user.passwordResetExpiresOn = new Date(resetExpires);
+  await user.save();
+
+  // Send reset password email
+  const emailTemplate = 'forgotPassword'; // to adjust this based on our email template name
+  const recipient = user.email;
+  const localVariables = {
+    resetToken,
+  };
+  await emailSender(emailTemplate, recipient, localVariables);
```
akosasante: `emailSender` exports an object now (`module.exports = { sendTestEmail, sendForgotPasswordEmail };`) so your require here is an object, we need to tell it which function to call otherwise we'll get the following error when we call the endpoint:



change required:

```suggestion
  await emailSender.sendForgotPasswordEmail(emailTemplate, recipient, localVariables);
```

src/mailing/sender.js
```diff
@@ -31,4 +34,34 @@ async function sendTestEmail(recipient) {
...
+  // Load and render the HTML template
+  const htmlTemplate = await ejs.renderFile(
+    path.join(__dirname, '../email/templates/html.ejs'),
```
akosasante: `__dirname` == `<backend root>/src/mailing` 

two dots (`..`) == "go up one level" == `<backend root>/src`

So right now it's going to be looking for the template in: `<backend root>/src/email/templates/html.ejs`. But we don't have that file/folder anymore. 

Screenshot 2023-09-07 at 1 28 45 PM

Instead we should be pointing it to the `templates` folder that's here in the same `mailing` directory (so only one dot needed)

src/mailing/templates/html.ejs
akosasante, line 1: preferably, these should be in a subfolder.

For example, what happens if we need to create template files for "confirmed order" emails? We can't have them all be under `templates` directly with the same names. For better organization we can have subfolders for each use case. So I would move these html.ejs, subject.ejs, and style.css to a `src/mailing/templates/forgot_password/` folder

akosasante:
> Now seems to be ok, the only thing I am doubt about is this line 2023-09-06 22:42:34 \[INFO\]: GET /api/v1/users/nonexistentUserId 404 2.271 ms - 51 
> 
> image

That looks fine to me; all the tests are passing.

The 404 line is from the express server using `winston` to log any http requests to it. And in this test in `userController.test.js` we intentionally do a call that returns a 404 to verify that it behaves as expected in that error case: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/test/userController.test.js#L84-L95

src/mailing/sender.js
DYA13: Done commit 73944aa0890d78246a92d3482217b90b33a1951c

src/mailing/templates/html.ejs
DYA13: Done , commit 1bb3c4a3591fc94547e67b48ebf067135ca8fb38

DYA13:
Yeh the line GET /api/v1/users/nonexistentUserId 404 2.271 ms - 51  seems it  simulates the scenario where a user attempts to access information about a user that doesn't exist (nonexistentUserId).  Seems if the resource is not found , it responds with 404.  🤔

src/mailing/sender.js
```diff
@@ -31,4 +34,34 @@ async function sendTestEmail(recipient) {
...
+    service: 'gmail',
+    auth: {
+      user: process.env.EMAIL_USER,
+      pass: process.env.EMAIL_PASSWORD,
+    },
+  });
+
+  // Load and render the HTML template
+  const htmlTemplate = await ejs.renderFile(
+    path.join(__dirname, './email/templates/html.ejs'),
```
akosasante:
```suggestion
    path.join(__dirname, './templates/forgot_password/html.ejs'),
```

DYA13:
Akos I still have the problem , it says Email is not defined in mailing/sender.js line 16`const baseEmail = new Email({`

DYA13:
Jest tells that "Email is not defined"  and the same for "dendTestEmail"
image

akosasante:
Looking at Github Actions, ES Lint is saying to check line 16 (character 23) in the `sender.js` file.

That's where this line is: `const baseEmail = new Email({` 

So it's saying we're using the variable `Email` which has not been created or imported yet. 

This line came when you merged in changes from the main branch in this commit: cba0f5d#diff-8126c036a6a21b3db3916fb16588ba4c6db6cebe9bb82491304bcc87f4c2576c

  
Either when git was figuring out how to merge stuff, or possibly if you had to resolve any conflicts you may have accidentally ommitted that. 

If you take a look at the code in `main` you can see that `Email` variable comes from being imported from the `email-templates` library: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/mailing/sender.js#L2

DYA13:
Thank you Akos, yes import of Email was missing, I have added in this commit d45b8ffc6e85d734cf17c796b23a8b44698eed21

DYA13:
Akos , but I have error 404 when testing this endpoint anyway, where should I look for an error?
image

akosasante:
> Akos , but I have error 404 when testing this endpoint anyway, where should I look for an error? 
> 
> image

Hmm it says he error was `User not found` and came from line 136 (character 11) in the `authController`. So that's pointing to this part of the code: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/endpoint/resetPasswordEmail/src/controllers/authController.js#L135-L137](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/endpoint/resetPasswordEmail/src/controllers/authController.js#L135-L137)

That means it could  not ifnd a user in the connected mongo database with whichever email you passed in the request body. Have you confirmed that a user exists with that email address in the db?

DYA13:
It works now 💪
image

src/mailing/sender.js
```diff
@@ -43,5 +45,34 @@ async function sendTestEmail(to, username) {
...
+    subject: 'Password Reset Request',
+    html: htmlTemplate,
+  };
+
+  try {
+    await transporter.sendMail(mailOptions);
+    console.log('Password reset email sent successfully');
+  } catch (error) {
+    console.error('Error sending password reset email:', error);
+  }
```
akosasante: Good progress; there's a few things remaining to fix; but I think it would be best to refactor this to use the `email-templates` library at this point:

Now that natalia's `email-templates` change has been merged into here though, we can simplify this a bunch by using that library instead of having to handle all the templating and whatnot manually. It also allows the CSS to be properly inlined and take effect.

```suggestion
  try {
    await baseEmail.send({
      template: 'forgot_password',
      message: { to: toEmail },
      locals: { resetLink: `${process.env.BACKEND_BASE_URL}/reset_password?token=${resetToken}`},
    })
    console.log('Password reset email sent successfully');
  } catch (error) {
    console.error('Error sending password reset email:', error);
  }
```

*   `template` tells `email-templates` library which folder in the `templates` directory to look for the `html.ejs` and `subject.ejs` files to use for this email
*   Your `html.ejs` is expecting a `resetLink` variable, so we will pass that in the locals property for local variables. If y'all will build the "reset password" page in the backend then we want to make that link be `http://localhost:8000/<path to reset password page>?token=<reset token>` for local dev and `https://musicstore.onrender.com/<path to reset password page>?token=<reset token>` for production. So best way to do that is to store the base url in the environment variables. If the reset password page will be built in the frontend, then instead have an environment variable that switches the base url appropriately for the localhost:3000 vs beatbazaar.onrender.com URLs
*   The `baseEmail` variable contains configuration that is common to all emails we send like the fact that we're using EJS, the folder where the templates live, the folder where the styles and any other resources live, etc. 

I'll still leave comments on the current approach just for learning purposes, but I would suggest replacing the content of this function with the above code for consistency with the test email function and future email functions.

src/mailing/sender.js
```diff
@@ -43,5 +45,34 @@ async function sendTestEmail(to, username) {
...
+  });
+
+  // Load and render the HTML template
+  const htmlTemplate = await ejs.renderFile(
+    path.join(__dirname, './templates/forgot_password/html.ejs'),
+    { resetLink: `http://localhost:8000/api/v1/reset-password/${resetToken}` }
+  );
+
+  const mailOptions = {
+    from: 'your@example.com',
```
akosasante: This email **must** match the user used in the `auth` otherwise when it won't send the email (at least this is the case for most email services) and you will see an SMTP error in the backend terminal logs.

```suggestion
    from: process.env.EMAIL_USERNAME,
```

src/mailing/sender.js
```diff
@@ -43,5 +45,34 @@ async function sendTestEmail(to, username) {
...
+    auth: {
+      user: process.env.EMAIL_USER,
+      pass: process.env.EMAIL_PASSWORD,
+    },
+  });
+
+  // Load and render the HTML template
+  const htmlTemplate = await ejs.renderFile(
+    path.join(__dirname, './templates/forgot_password/html.ejs'),
+    { resetLink: `http://localhost:8000/api/v1/reset-password/${resetToken}` }
```
akosasante: Likely this URL will not be under `/api/v1` since it's not an API endpoint but rather a EJS/HTML page for the user to see and where they can input their resetted password.

src/mailing/sender.js
```diff
@@ -43,5 +45,34 @@ async function sendTestEmail(to, username) {
   });
 }
 
-// Export the transport and the sendTestEmail function
-module.exports = { sendTestEmail };
+async function sendForgotPasswordEmail(toEmail, resetToken) {
+  const transporter = nodemailer.createTransport({
+    service: 'gmail',
+    auth: {
+      user: process.env.EMAIL_USER,
```
akosasante: The value in the `.env` file that I see in the team slack is `EMAIL_USERNAME`: https://codethedream.slack.com/archives/C05G08542H3/p1693435063248899

src/mailing/sender.js
```diff
@@ -43,5 +45,34 @@ async function sendTestEmail(to, username) {
   });
 }
 
-// Export the transport and the sendTestEmail function
-module.exports = { sendTestEmail };
+async function sendForgotPasswordEmail(toEmail, resetToken) {
+  const transporter = nodemailer.createTransport({
+    service: 'gmail',
```
akosasante: Unless you're explicitly trying to test by sending from a gmail account, what we currently have in other branches is using `process.env.EMAIL_SERVICE` which is `'Outlook'` to match the email we're sending from (`codethedream.practicum.team4@outlook.com`). Are you testing from a different address? If so, that's totally fine but let's make sure to commit the version that everyone can use for testing.

src/mailing/sender.js
```diff
@@ -1,6 +1,8 @@
 const nodemailer = require('nodemailer');
 const Email = require('email-templates');
 const path = require('path');
+const ejs = require('ejs');
+require('dotenv').config();
```
akosasante, line 4: I don't think this is needed. When this file is run by the server, the dotenv library will already have been imported in `expressServer.js`

src/mailing/sender.js
```diff
@@ -1,6 +1,8 @@
 const nodemailer = require('nodemailer');
 const Email = require('email-templates');
 const path = require('path');
+const ejs = require('ejs');
```
akosasante: If you switch to using `email-templates` then this will no longer be needed

src/mailing/templates/forgot_password/html.ejs
```diff
@@ -0,0 +1,26 @@
+<!doctype html>
+<html>
+  <head>
+    <title>Password Reset</title>
+    <link rel="stylesheet" type="text/css" href="style.css" />
```
akosasante: `email-templates` library will be looking for the file starting from the `templates` directory, so we have to write this path/href relative to that folder:

```suggestion
    <link rel="stylesheet" type="text/css" href="forgot_password/style.css" />
```

src/controllers/authController.js
akosasante: Yes this line (155) that passes in `emailTemplate` can be removed. 

Your code in `sender.js` has `sendForgotPasswordEmail` only expecting two arguments: the recipient email address, and the reset token

src/controllers/authController.js
```diff
@@ -120,8 +122,47 @@ const logout = async (req, res) => {
...
+  const resetToken = crypto.randomBytes(20).toString('hex');
+  const resetExpires = Date.now() + 15 * 60 * 1000; // 15 minutes from now
+
+  // Update user's reset token and expiration date
+  user.passwordResetToken = resetToken;
+  user.passwordResetExpiresOn = new Date(resetExpires);
+  await user.save();
+
+  // Send reset password email
+  const emailTemplate = 'forgotPassword'; // to adjust this based on our email template name
```
akosasante: This should no longer be needed

src/controllers/authController.js
```diff
@@ -120,8 +122,47 @@ const logout = async (req, res) => {
...
+  const recipient = user.email;
+  const localVariables = {
+    resetToken,
+  };
+  await emailSender.sendForgotPasswordEmail(
+    emailTemplate,
+    recipient,
+    localVariables
```
akosasante: You're passing in an object `{ resetToken: ... }` but your code in sender.js is expecting just a single string value for the resetToken as the second argument.

src/controllers/authController.js
DYA13: Done , commit 66413a776311332e85d5529378bcdeda744cd532

src/mailing/sender.js
DYA13: Thank you Akos 👍

src/mailing/sender.js
DYA13: Sorry, you are right 👍

src/mailing/sender.js
DYA13: Removed import ejs , commit d5d367fe09e202e201a1512da868012240df2b8b

src/mailing/sender.js
DYA13: Removed

src/controllers/authController.js
DYA13: Edited , commit b0d0cd7fb1672665e7d2ae532d987ca63a10c327
Could you please check if I have done correctly? Thank you

src/controllers/authController.js
akosasante: I think you accidentally added it back in in the latest commit: b0d0cd7

src/mailing/sender.js
akosasante: we can also remove `require('dotenv').config();`

akosasante (approved):
Main thing is just fix the call in `authController.js` to not be passing in an extra argument (`emailTemplate`) then you should be good to go :+1:

DYA13:
Removed variable and argument emailTemplate , commit 6c7acbfa8c8b5164490ce70abb393ded24d3d6a6, commit 2587e8cc800f8bc98936f10229881a1273155d3c

PR #49, opened 2023-09-03
PS192 socket-unit-test

mvstoyan:
Test failed, I keep trying to understand hot to fix and pass

### DIFF
--- a/test/socket.test.js
+++ b/test/socket.test.js
@@ -1,5 +1,6 @@
 const { app } = require('../src/expressServer');
 const io = require('socket.io-client');
+const { intervalId } = require('../src/models/Order');
 
 let server;
 const PORT = 8001;
@@ -9,11 +10,11 @@ beforeAll(async () => {
 });
 
 afterAll(async () => {
+  clearInterval(intervalId); // Clear the interval after tests are finished
   server.close();
 });
 
 test('test:ping socket event', (done) => {
-  jest.setTimeout(30000);
   // Socket.io client for sending a message
   const senderClient = io.connect(`http://localhost:${PORT}`, {
     transports: ['websocket'], // WebSocket transport
@@ -48,7 +49,7 @@ test('test:ping socket event', (done) => {
       recipientClient.disconnect();
 
       // Small timeout before calling "done()" to give time for the disconnect event to fire
-      setTimeout(() => done(), 1000);
+      done();
     });
   });
-}, 60000);
+});
--- a/test/socket.test.js
+++ b/test/socket.test.js
@@ -1,0 +1,35 @@
+const { app } = require('../src/expressServer');
+const io = require('socket.io-client');
+
+let server;
+const PORT = 8001;
+
+beforeAll(async () => {
+  server = await app.listen(PORT);
+});
+
+afterAll(async () => {
+  server.close();
+});
+
+test('test:echo socket event', (done) => {
+  jest.setTimeout(30000);
+  const client = io.connect(`http://localhost:${PORT}`, {
+    transports: ['websocket'],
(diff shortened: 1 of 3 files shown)


test/socket.test.js
```diff
@@ -0,0 +1,35 @@
...
+  const client = io.connect(`http://localhost:${PORT}`, {
+    transports: ['websocket'],
+    forceNew: true,
+    reconnection: false,
+  });
+
+  const messageToSend = 'Hello, server!';
+
+  client.once('connect', () => {
+    client.emit('test:echo', messageToSend);
```
akosasante, line 26: `test:echo` was for the example. But in this application we don't have a `test:echo` channel/event handler defined. We do av

test/socket.test.js
```diff
@@ -0,0 +1,35 @@
...
+
+    client.on('test:echo', (response) => {
```
akosasante: `test:echo` was for the example. But in this application we don't have a `test:echo` channel/event handler defined. We do have a `test:ping` channel though: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/PS192-socket-unit-test/src/live/index.js#L6](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/PS192-socket-unit-test/src/live/index.js#L6)

If we want to create a `test:echo` handler for the sake of this unit test exercise that's totally fine. Otherwise, we should be testing the existing `test:ping` handler.

Currently this test is failing because line 28 `client.on('test:echo')` is waiting and waiting for our server to send a websocket message to the `test:echo` listener, but there's no code in our backend doing that; so the test just hangs until Jest eventually kills it.

test/socket.test.js
```diff
@@ -0,0 +1,35 @@
...
+      expect(response).toBe(messageToSend);
```
akosasante: If we create a `test:echo` event handler we will need to have it do `this.emit(<same message that was sent in>)`. `.emit` will send that message to _all_ sockets listening to that topic. 

But out current event handler for `test:ping` does something a little different: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/PS192-socket-unit-test/src/live/testHandlers.js#L5.](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/PS192-socket-unit-test/src/live/testHandlers.js#L5.) It does `this.broadcast.emit("user sent: <message that was sent in>")` so not only is the string that's being emmited is different than what was sent in, it also uses `.broadcast.emit` which is a little different from `.emit`. With `.broadcast.emit` the message is sent to _all listeners EXCEPT the one who sent the message_. This can be useful for example with chat; the person who joined the chat doesn't need to get a notification that they themselves joined the chat; that message is only useful for _everyone else in the room_.

But in that case, to test it we need to have a second client to receive the message. We can create that second client in exactly the same way as lines 17-21 here. And then we can have something like:

```javascript
  senderClient.once('connect', () => {
    senderClient.emit('test:ping', ...);
  });

  recipientClient.once('connect', () => {
    recipientClient.on('test:ping', (response) => {
      expect(response).toBe(...);

      senderClient.disconnect();
      recipientClient.disconnect();

      // small timeout before calling "done()" to give time for the disconnect event to fire
      setTimeout(() => done(), 1000)
    });
  })
```

mvstoyan: Thank you very much, work now and I understand.  But I cant understand how to stop setInterval from Order models. I tried add clearInterval(intervalId) to afterAll, but not work.
screen

mvstoyan: Thank you very much, work now and I understand.❤️

akosasante (approved):

PR #50, opened 2023-09-03
Email templates

NatachaKey:


src/mailing/sender.js
```diff
@@ -9,26 +11,28 @@ const transport = nodemailer.createTransport({
...
-        if (err) {
-          reject(err);
-        } else {
-          resolve(res);
-        }
-      }
-    );
+const baseEmail = new Email({
+  message: {
+    from: 'codethedream.practicum.team4@outlook.com',
```
akosasante: Not that this value needs to be a secret; but I think we already do have it in the `.env` files (`process.env.EMAIL_USERNAME`) so we can just continue to pull it from there?

src/mailing/sender.js
```diff
@@ -9,26 +11,28 @@ const transport = nodemailer.createTransport({
...
+  },
+  transport: transport, // Use the transport you created
+  subjectPrefix: process.env.NODE_ENV === 'production' ? '' : '(TEST) ',
+  views: {
+    root: path.resolve('../templates/test'), // this tells email-templates to look for templates starting from this path
```
akosasante: Your `templates` file is not inside `src`.

You should move the `templates` folder to inside `src` and better yet inside `src/mailing` since those templates will be only used by our email functions so it makes sense for them to live in the same place. 

---

When using `path.resolve` it will calculate the path relative to where node is running from (so usually that will mean relative to the root of the repo). So `../templates/test` is saying "go up one folder ('`..`') from where this backend repo is stored". You can double check this by adding a log somewhere in this file `console.log(path.resolve("../templates/test"));`. On my computer this logs: `/Users/aasante/dev/practicum2023/templates/test` which is _outside_ of my repo folder.

You can compare by also console.logging the Current Working Directory with `console.log(process.cwd());`. On my local computer that prints the directory that my backend repo is at: `/Users/aasante/dev/practicum2023/backend`. 

---

Here we are trying to tell Email-Templates the _root_ folder to look for all templates; not just the specific folder for a single template. So we should be directing it to the `/templates` folder not to the `/templates/test` folder. Because later on as we add more emails, there will be `/templates/forgot_password`, `/templates/order_email` etc. But all of those will use the same view root. Then in the actual email sending function (eg: sendTestEmail) we tell Email-Templates which folder to use via the `template` option.

---

Then, if your templates live at: `<backend_repo_root>/src/mailing/templates` then you can refer to it using any of the following:

*   `path.resolve('src/mailing/templates')` \<-- giving path.resolve the path from the root of the repo / the current working firectory where node is running from
*   `path.resolve(__dirname, "templates")` or `path.resolve(__dirname, "./templates")` \<-- using the special variable `__dirname` which returns the directory of _the file where the code is_ and then telling path to check the same folder/directory and navigate to `templates` folder from there

templates/test/subject.ejs
akosasante, line 1: This should have some text in here so that email-templates knows what to use as the subject line.

src/mailing/sender.js
```diff
@@ -9,26 +11,28 @@ const transport = nodemailer.createTransport({
...
+    },
+  },
+});
+
+// Function to send a test email
+async function sendTestEmail(to, username) {
+  return baseEmail.send({
+    template: 'test',
+    message: { to },
+    locals: { name: username },
```
akosasante: Your email uses the variable `username`

```
 <p>Hello, <%= username %>, this is a test email.</p>
```

but here in locals we're passing in an object that has a field `name` that has a value of whatever `username` evaluates to. So this will cause an error. We should either change the locals to `{ username: username }` or the EJS template to `<p> Hello, <%= name %>, etc`

src/mailing/sender.js
```diff
@@ -9,26 +11,28 @@ const transport = nodemailer.createTransport({
...
+  views: {
+    root: path.resolve('../templates/test'), // this tells email-templates to look for templates starting from this path
+    options: {
+      extension: 'ejs', // this tells email-templates that the file will end tith *.ejs
+    },
+  },
+});
+
+// Function to send a test email
+async function sendTestEmail(to, username) {
```
akosasante, line 38: To be able to test this, you'll want to update `src/mailing/test.js` file to pass in a `"username"` to be able to see it in the preview email.

NatachaKey: @akosasante I´ve hardcoded username- maybe it would be better to add here a username from the User model- can I do that and import it from the User.js ?  - can I do that and import it from the User.js ?

NatachaKey: I will try it now

NatachaKey: current code doesn´t work

src/mailing/templates/test/subject.ejs
akosasante, line 1: The subject-line of the email will be plain text (with potentially some interpolated variables); we cannot style it or use any HTML content there

src/mailing/templates/test/subject.ejs
```diff
@@ -0,0 +1,18 @@
...
+    // Use JavaScript to extract the subject without tags and display it
+    //doesn´t help
+    document.addEventListener('DOMContentLoaded', function () {
+      const titleElement = document.querySelector('title');
+      const subject = titleElement.textContent;
+      document.body.innerHTML = `<h1>${subject}</h1>`;
+    });
+  </script>
+</body>
+</html>
```
akosasante: So for example, if the subject will always be "TEST EMAIL". Then that would be:

```suggestion
TEST EMAIL
```

Or if we wanted the subject to be passed in from the sender function as a locals variable then we can interpolate it with EJS:

```suggestion
<%= subject %>
```

Or if we want to have some static content and some dynamic content, it could be something like:

```suggestion
Hello, <%= username %> - THIS IS A TEST EMAIL
```

src/mailing/test.js
```diff
@@ -1,10 +1,30 @@
...
-    const response = await sendTestEmail(TEST_RECIPIENT_EMAIL);
-    console.log('Test email sent successfully:', response);
+    //was trying to add a custom user name from db to the email -needs changes on line 19->  User.findOne({ _id: id })
+    const id = '64d44bc8337399ccf9ad7e4d'
+    const user = await User.findOne({ _id: id }).select('-password');
+    console.log(user);
+    
+  if (!user) {
+    // Throw a NotFoundError if the user is not found
+    throw new CustomError.NotFoundError(`No user with id: ${req.params.id}`);
```
akosasante: This test file is only ever run as a script (`node -r dotenv/config src/mailing/test.js`); it is not an Express controller handler called by the server, so we don't have the variable `req` available.

src/mailing/templates/test/subject.ejs
akosasante: You'll see in both the preview and in the actual send (if you add `send: true` to baseEmail) that the HTML doesn't get parsed for the subject:

src/mailing/test.js
```diff
@@ -1,10 +1,30 @@
...
+const User = require('../models/User'); 
+const CustomError = require('../errors');
+
 const TEST_RECIPIENT_EMAIL = process.env.TEST_RECIPIENT_EMAIL;
 
+// Connect to the MongoDB database
+mongoose.connect(process.env.MONGO_URL, {
+  useNewUrlParser: true,
+  useUnifiedTopology: true,
+});
```
akosasante: This function is asynchronous; so it is possible that by the time line 19 where we try to fetch the user by id runs; this function may not have finished connecting to mongoose. To ensure that this finishes first, you would want to put this in an async function and await it (or put it in the `async main` function and await it there)

src/mailing/test.js
```diff
@@ -1,10 +1,30 @@
...
+
 async function main() {
   try {
-    const response = await sendTestEmail(TEST_RECIPIENT_EMAIL);
-    console.log('Test email sent successfully:', response);
+    //was trying to add a custom user name from db to the email -needs changes on line 19->  User.findOne({ _id: id })
+    const id = '64d44bc8337399ccf9ad7e4d'
+    const user = await User.findOne({ _id: id }).select('-password');
```
akosasante: So for this test we will always be sending to this particular user right? It may be easier for other members to use this test script if we make this just send to the `TEST_RECIPIENT_EMAIL` user:

```suggestion
    const user = await User.findOne({ email: process.env.TEST_RECIPIENT_EMAIL });
```

(redacting the password doesn't really matter here since this is not data that we're returning in an API to any users)

If you tried this before, it may not have worked because of the above comment about the mongoose connection being async

src/mailing/templates/test/style.css
akosasante, line 1: Possible you haven't gotten to this bit yet; but in order for the CSS to take effect, you'll need to add `juice` options to the `baseEmail` object

NatachaKey: @akosasante oopss it got deleted somehow, i added it now , i can see css changes being applied. Akos I was trying to add a function  sendEmailToAllUsers() and it showed me that the emails were sent successfully but- as for me I didn´t receive it (I registered a real email ) and now it stopped working and doesn´t let me send email to a test email in .env saying that it was marked as spam- i was trying to find a solution , change subject and html but it doesn´t help
image

akosasante: Based on that error message, it looks like we've hit a sending quota on Outlook: [https://answers.microsoft.com/en-us/outlook_com/forum/all/email-refusal/e342fa28-9922-43c9-83f9-375a116b2b3e](https://answers.microsoft.com/en-us/outlook_com/forum/all/email-refusal/e342fa28-9922-43c9-83f9-375a116b2b3e)

You should still be able to test by setting `preview: true`, and `send: false` on the baseEmail object and just using the preview for now rather than the actual sent message. I'll take a look later today to see if there's anything we can do to adjust the quota or if we just need to wait for it to reset.

src/mailing/test.js
```diff
@@ -1,30 +1,64 @@
...
+      useNewUrlParser: true,
+      useUnifiedTopology: true,
+    });
+    console.log('Connected to MongoDB');
+  } catch (error) {
+    console.error('Error connecting to MongoDB:', error.message);
+  }
+}
+// Call the async function to connect to the database
+connectToDatabase();
```
akosasante: Make sure to `await` this otherwise it can still possibly not finish before you try to fetch things. That may mean you have to move it this line (`connectToDatabase`) to inside the `main()` function

src/mailing/templates/test/style.css
NatachaKey: @akosasante oh yes that helped-  successfully sent them all 
image

akosasante (approved):
LGTM

src/mailing/sender.js
```diff
@@ -1,34 +1,49 @@
...
+    root: path.resolve('src/mailing/templates'), // this tells email-templates to look for templates starting from this path
+    options: {
+      extension: 'ejs', // this tells email-templates that the file will end tith *.ejs
+    },
+  },
+  send: false,
+  preview: true,
+  juice: true,
+  juiceResources: {
+    webResources: { relativeTo: path.resolve('src/mailing/templates/test') },
```
akosasante: Remember that this \`baseEmail\` variable is the config for \_all\_ of our email templates. 

This means that with this code, for _all templates_, it's going to look for the styles in the `test` folder. If you want to put all the styles and js and stuff in a single folder, that's fine, but maybe in that case put it in a folder like: `src/mailing/templates/web_resources` or something like that.

Otherwise, my initial suggestion was that this would be: `path.resolve('src/mailing/templates')` . Then if you need to refer to a style that's in the test folder, you would in that EJS file do: `<link rel='stylesheet' href='./test/style.css' data-inline>`

src/mailing/sender.js
```diff
@@ -1,34 +1,49 @@
...
+  },
+});
+console.log(path.resolve('../templates'));
+console.log(process.cwd());
```
akosasante: can be removed before merging

akosasante (approved):
Based on conversation from yesterday's meeting I want to clarify; you can in fact merge this one to main now; the other branch that you have that is either based on this one or has merged in commits from this one, will need to pull in `main` afterwards and resolve any conflicts but otherwise it should be fine.

Waiting on the more recent PRs to be ready was more a point of caution when we had many different people branched off the same work earlier in the project.

src/mailing/sender.js
NatachaKey: @akosasante thank you!! i mush have fixed it in another branch and forgot to fix it in this one. will commit now and merge

PR #51, opened 2023-09-04
Admin swagger

mvstoyan:


mvstoyan:
I need help on how to create documentation correctly for adminPage. Please can you guide me?

src/controllers/adminController.js
```diff
@@ -3,6 +3,17 @@ const Album = require('../models/Album');
...
+  /*
+  #swagger.tags = ['Admin']
+
+  #swagger.summary = 'Admin Home Page'
+  #swagger.description = 'Fetch all albums and render the admin home page.'
+
+  #swagger.responses[200] = {
+    description: 'Admin home page fetched successfully.',
+    schema: { $ref: '#/definitions/Album' }
+  }
```
akosasante: Really the return value of this endpoint is not JSON of the albums, but rather an HTML page (eg: if you call this endpoint Postman with the proper credentials, you'll get the following):



Screenshot 2023-09-04 at 1 27 18 PM

Screenshot 2023-09-04 at 1 27 25 PM

If we want to document that we'd need to do something like:

```
  #swagger.responses[200] = {
    content: { 'text/html': {} }
  }
```

src/routes/adminRoutes.js
```diff
@@ -5,13 +5,24 @@ const adminController = require('../controllers/adminController');
...
+  }) /* #swagger.tags = ['Admin'] 
+  #swagger.summary = 'Admin Login'
+  #swagger.description = 'Authenticate as an admin to access the admin page.'
+
+  #swagger.responses[200] = {
+    description: 'Admin authenticated successfully.',
+  }
+  #swagger.responses[401] = {
+    description: 'Authentication failed.',
+  } */
```
akosasante: I don't think swagger-autogen reads any comments on middleware. Or at least, I'm not seeing any of this show up in the docs :shrug:

src/controllers/adminController.js
```diff
@@ -3,6 +3,17 @@ const Album = require('../models/Album');
...
+*/
 };
 
 const { updatePriceOfAlbums } = require('../controllers/albumController');
```
akosasante: I'm not sure if there's an easy way to tell swagger-autogen about this endpoint since we're just re-exporting the endpoint from another controller. 

You could try making this it's own controller handler that just calls into the albumController one?

```javascript
const { updatePriceOfAlbums: originalUpdatePrice } = require('../controllers/albumController');
const updatePriceOfAlbums = async (req, res) => {
  /* #swagger.tags = ['Admin'] */
  //etc any other swagger infor
  await originalUpdatePrice(req, res);
}
```

akosasante (commented):
Technically I wouldn't consider it necessary to document the admin endpoints since those aren't really expected to be called by another team or third-party client.   
 

But it's totally fine to do so if we want to here :+1:

src/controllers/adminController.js
mvstoyan: Thank you very much. I think its work. But now I cant update price and most likely there is another mistakes, it is very difficult for me to get documentation, that's why I took the admin page😅
swagger

src/controllers/adminController.js
```diff
@@ -3,9 +3,44 @@ const Album = require('../models/Album');
...
+  {
+    "id": "64d2ae2308a725b72bd5c0dc",
+    "price": 9.99
+  },
+  {
+    "id": "another_album_id",
+    "price": 19.99
+  }
+]
+}
```
akosasante:
```suggestion
    /*	#swagger.requestBody = {
            required: true,
            "@content": {
                "application/json": {
                    "schema": [{
                        "$ref": "#/definitions/UpdatePriceOfAlbums"
                      }]
                 }
         }
    }
    */
```

And then define the `[ { id: "64d2ae2308a725b72bd5c0dc", price: 9.99} ]` in a definitions block in swagger.js

The `@content` allows us to define request body shape more directly

src/controllers/adminController.js
```diff
@@ -3,9 +3,44 @@ const Album = require('../models/Album');
...
+#swagger.responses[200] = {
+  description: 'Album prices were successfully updated.',
+  content: { 'text/html': {} }
+}*/
```
akosasante: This endpoint actually returns a [JSON response](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/admin-swagger/src/controllers/albumController.js#L112) which is a list of album objects that were updated. So we can define that using the existing Album definition: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/admin-swagger/swagger.js#L80-L92

mvstoyan: @akosasante Thank you very much, its work now🎉

src/controllers/adminController.js
```diff
@@ -3,9 +3,41 @@ const Album = require('../models/Album');
...
+  }
+*/
 };
 
-const { updatePriceOfAlbums } = require('../controllers/albumController');
+const {
+  updatePriceOfAlbums: originalUpdatePrice,
+} = require('../controllers/albumController');
+const updatePriceOfAlbums = async (req, res) => {
+  /*#swagger.summary = 'Update prices of albums passed in req.body'
```
akosasante, line 21: We can add a tag here so that this shows up together with the other admin endpoint:

```suggestion
  /*#swagger.summary = 'Update prices of albums passed in req.body'
    #swagger.tags = ['Admin']
```

akosasante (approved):
:+1:

PR #55, opened 2023-09-07
PS201-calculate Average Rating

mvstoyan:


src/models/Review.js
```diff
@@ -32,4 +32,47 @@ const ReviewSchema = new mongoose.Schema(
...
+    },
+  ]);
+
+  try {
+    await this.model('Album').findOneAndUpdate(
+      { _id: albumId }, // Find the album by its unique ObjectId.
+      {
+        // Update the 'averageRating' field of the album document with the rounded
+        // value from the aggregation result. If 'averageRating' is undefined, it defaults to 0.
+        averageRating: Math.ceil(result[0]?.averageRating || 0),
```
akosasante: `Match.ceil` will always round up. So if we have an average rating of 4.1, this will be shown as 5 stars which doesn't seem very accurate. We likely instead want [Math.round](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round) which will round to nearest integer in either direction:
```suggestion
        averageRating: Math.round(result[0]?.averageRating || 0),
```

src/models/temp.js
akosasante, line 1: Will this be removed before merging?

akosasante (approved):
One small change request then can merge if CI is passing 👍

src/models/temp.js
mvstoyan: Removed

PR #56, opened 2023-09-07
reviewcontroller updated

Musab-Kara:
album detail page review controller changes


.DS_Store
akosasante, line 1: This is a file from Mac operating system that it uses to store some info about directories; it's not harmful to commit but to keep things clean we'd rather not commit it: [https://en.wikipedia.org/wiki/.DS_Store](https://en.wikipedia.org/wiki/.DS_Store)

You can follow the steps [here](https://stackoverflow.com/a/107921) or [here](https://www.linkedin.com/pulse/how-remove-dsstore-files-from-git-repositories-chandan-thakur) to clean it up (though instead of pushing to `origin master`, you'd be pushing to `origin reviewcontroller-updated`

akosasante (commented):
@Musab-Kara I'm not sure if I understand the usecase of these changes? It seems to just be changing the syntax of the delete review endpoint slightly but with no behavioural changes. Could you explain a bit what you're aiming to do here, thanks!

src/controllers/reviewController.js
```diff
@@ -143,7 +143,7 @@ const deleteReview = async (req, res) => {
   }
 
   // Delete the review
-  await review.remove();
```
akosasante, line 146: I don't think this line needs to change necessarily

src/controllers/reviewController.js
```diff
@@ -125,14 +125,14 @@ const getSingleReview = async (req, res) => {
 
 // delete review
 const deleteReview = async (req, res) => {
-  const { reviewId } = req.params; // take the id of the review from req.params
+  const { id } = req.params; // take the id of the review from req.params
```
akosasante, line 128: Ah I see now after looking at your frontend ticket; this endpoint wasn't working because it wasn't correctly destructuring the path parameter, good catch!

We could reduce this to just a single-line change by making this: `const { id: reviewId } = req.params;` that way the rest of the function could be left as it originally was, referring to the _variable_ `reviewId` that we've destructured here. Only a suggestion, either way is fine.

akosasante (approved):
Please fix the `.DS_Store` change before merging

Musab-Kara:
@akosasante I could not restore DsStore file so I opened a new PR with your recommendation. please see https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/62

closing this PR

PR #57, opened 2023-09-07
Payment email

NatachaKey:
also have an issue with applying style.css to html.ejs in orderCompleted template


src/controllers/orderController.js
```diff
@@ -5,58 +5,100 @@ const { BadRequestError } = require('../errors');
...
+});
+const db = mongoose.connection;
+
+db.on('error', (error) => {
+  console.error('Database connection error:', error);
+});
+
+db.once('open', () => {
+  console.log('Connected to the database');
+});
```
akosasante: I'm not clear on why this is needed here?

src/controllers/orderController.js
```diff
@@ -5,58 +5,100 @@ const { BadRequestError } = require('../errors');
...
-  // Send the payment intent client secret and order information to the client
-  res
-    .status(StatusCodes.CREATED)
-    .json({ clientSecret: paymentIntent.client_secret, order });
+    // Start a database transaction
+    await db.startTransaction();
+
+    // Save the order to the database
+    const savedOrder = await db.saveOrder(orderData);
+
```
akosasante: At this point, we've already saved the order earlier on 38 when it was created and again on 64 to add the payment id to it. Is anything new being saved here?

src/controllers/orderController.js
```diff
@@ -5,58 +5,100 @@ const { BadRequestError } = require('../errors');
...
-      user: req.user.userId, // user ID from the request
-    });
-  }
 
-  // Send the payment intent client secret and order information to the client
-  res
-    .status(StatusCodes.CREATED)
-    .json({ clientSecret: paymentIntent.client_secret, order });
+    // Start a database transaction
+    await db.startTransaction();
```
akosasante: Rather than creating a separate database connection and starting a transaction that way, we can use the existing connection(s) via our model: [https://mongoosejs.com/docs/transactions.html#:~:text=In%20practice%2C%20you,transaction%20error.](https://mongoosejs.com/docs/transactions.html#:~:text=In%20practice%2C%20you,transaction%20error.)

src/controllers/orderController.js
```diff
@@ -5,58 +5,100 @@ const { BadRequestError } = require('../errors');
...
+
+    // Save the order to the database
+    const savedOrder = await db.saveOrder(orderData);
+
+    // Send the order completion email
+    await sendOrderCompletedEmail(user.email, user.name);
```
akosasante: Will we also pass along the order items so that those can be shown in the email?

src/controllers/orderController.js
```diff
@@ -5,58 +5,100 @@ const { BadRequestError } = require('../errors');
...
+    // Commit the database transaction
+    await db.commitTransaction();
+
+    // Send the payment intent client secret and order information to the client
+    res
+      .status(StatusCodes.CREATED)
+      .json({ clientSecret: paymentIntent.client_secret, order });
+  } catch (error) {
+    // Handle errors and rollback the transaction if it fails
+    await db.rollbackTransaction();
```
akosasante: I don't see this function in the mongoose documentation

src/controllers/orderController.js
```diff
@@ -5,58 +5,100 @@ const { BadRequestError } = require('../errors');
...
+    await db.startTransaction();
+
+    // Save the order to the database
+    const savedOrder = await db.saveOrder(orderData);
+
+    // Send the order completion email
+    await sendOrderCompletedEmail(user.email, user.name);
+
+    // Commit the database transaction
+    await db.commitTransaction();
```
akosasante: I don't see this function in the mongoose documentation

src/mailing/sender.js
```diff
@@ -1,34 +1,70 @@
...
+// Function to send the order completion email
+async function sendOrderCompletedEmail(to, username) {
+  const html = await ejs.renderFile(
+    path.join(__dirname, 'templates', 'orderCompleted', 'html.ejs'),
+    { username: username } // Make sure username is defined
+  );
+
+  const subject = await ejs.renderFile(
+    path.join(__dirname, 'templates', 'orderCompleted', 'subject.ejs'),
+    { username: username } // Make sure username is defined
```
akosasante: Since we're using email-templates, we shouldn't need to manually find the template files. Email-templates asks us to pass `template: 'orderCompleted'` on line 63 and it will get know to grab `html.ejs` and `subject.ejs` from there.

src/mailing/sender.js
```diff
@@ -1,34 +1,70 @@
...
+  );
+  console.log('HTML Content:', html);
+  console.log('Subject:', subject);
+
+  return baseEmail.send({
+    template: 'orderCompleted',
+    message: { to },
+    locals: { username, subject },
```
akosasante: I don't think `subject` currently needs to be in the locals. Or at least, I'm not seeing it being used/interpolated in the html.ejs or subject.ejs files for this template

src/mailing/templates/orderCompleted/html.ejs
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Email Template</title>
+    <link rel='stylesheet' href='./style.css' data-inline>
```
akosasante: In `sender.js` when we set up the `baseEmail` configuration options, we tell `juice` (the library that handles parsing the CSS for these emails) to use `'src/mailing/templates'` as the base path when looking for CSS files (or other web resources) that are asked for by the HTML. So we need to link this stylesheet, relative to that path.

```suggestion
    <link rel='stylesheet' href='./orderCompleted/style.css' data-inline>
```

src/controllers/orderController.js
NatachaKey: @akosasante I was trying to implement a transaction and got some error and after attempts to find a solution included this mongo connection

src/controllers/orderController.js
NatachaKey: @akosasante I think we don't- will delete it now

src/controllers/orderController.js
NatachaKey: @akosasante yes, that was my idea, I'm just not sure yet how to write that

src/mailing/sender.js
NatachaKey: @akosasante will delete it, at first I didn´t write it but it was not working and I was figuring out - maybe there was an error, but I was wrong

src/mailing/sender.js
NatachaKey: @akosasante I see, will delete it also from // Function to send a test email
```
async function sendTestEmail(to, username) {
  return baseEmail.send({
    template: 'test',
    message: { to },
    locals: { username, subject },
  });
}
```
 because we are not using it there either

src/mailing/templates/orderCompleted/html.ejs
NatachaKey: @akosasante thank you so much!!!!!

src/controllers/orderController.js
akosasante: Yes we can do that for sure :+1:  We can pass in `order.orderItems` here or if you need other info from the order, you could even pass along the whole `order` variable (and add a matching argument where `sendOrderCompletedEmail` is defined)

deleteCollectionMongo.js
```diff
@@ -21,3 +21,5 @@ async function deleteCollection() {
 }
 
 deleteCollection();
+
+//was trying to delete order collection but it threw an error Error deleting collection: TypeError: Order.deleteMany is not a function- didn't find the solution
```
akosasante: I think because the Order file exports an object rather than a single value. So you'll need to import it by doing `const { Order } = require('./src/models/Order');

see example in controller: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/paymentEmail/src/controllers/orderController.js#L1
see export here : https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/paymentEmail/src/models/Order.js#L130

NatachaKey: @akosasante thank you so much for the explanation!!!

src/controllers/orderController.js
NatachaKey: @akosasante Akos, I was trying to do that but couldn't receive albumName and price from the populated 'album' property in OrderItemSchema- could you please take a look- the maximun I was able to reach is this   {
                "album": "64d36bd25c5dec8e599bee05",
                "quantity": 1,
                "_id": "64faef78e19c2e36bdd0db53"
            },
image
I think I need to populate 'album' in OrderItemSchema so I added this code 
image
but it didn't help

src/controllers/orderController.js
```diff
@@ -5,58 +5,103 @@ const { BadRequestError } = require('../errors');
...
-  for (const orderItem of orderItems) {
-    await PurchasedAlbum.create({
-      album: orderItem.album, // album ID from orderItem
-      user: req.user.userId, // user ID from the request
+// Populate the orderItems with full album data
+const populatedOrderItems = await Promise.all(
+  orderItems.map(async (orderItem) => {
+    return await orderItem.populate('album').execPopulate();
+  })
+);
```
akosasante: Currently this won't work because `populate` is a mongoose function that works on mongoose model instances. But the `orderItems` variable isn't one we got from the database, it's a plain Javascript object that we got from the HTTP request.

src/controllers/orderController.js
```diff
@@ -5,58 +5,103 @@ const { BadRequestError } = require('../errors');
...
+  // Create an array to store the order items with full album data
+  const orderItemsWithFullAlbum = populatedOrderItems.map((orderItem) => ({
+    album: orderItem.album.toObject(), // Use toObject() to convert the full album to a plain JavaScript object
+    quantity: orderItem.quantity,
+  }));
```
akosasante: Could you explain why we are trying to populate the album object in order to save it here? Saving the album id should be sufficient and if we need the full object then we can populate it when _fetching_ from the database instead.

src/controllers/orderController.js
```diff
@@ -5,58 +5,103 @@ const { BadRequestError } = require('../errors');
...
+    order.paymentIntentId = paymentIntent.id;
+
+    for (const orderItem of orderItems) {
+      const purchasedAlbum = new PurchasedAlbum({
+        album: orderItem.album,
+        user: req.user.userId,
+      });
+
+      await purchasedAlbum.save({ session });
+    }
```
akosasante: Let's move this block above the stripe block; that way we know we've done as much of the database saving as possible before trying to save to stripe (since the stripe part cannot be easily reversed here if the transaction fails; while all the db queries _will_ be reversed if the transaction fails)

src/controllers/orderController.js
```diff
@@ -5,58 +5,103 @@ const { BadRequestError } = require('../errors');
...
+
+
+
+
+    await session.commitTransaction();
+
+    
+    // Send the order completion email after THE ORDER HAS STATUS COMPLETE
+    await sendOrderCompletedEmail(user.email, user.name, orderItemsWithFullAlbum);
```
akosasante: Here we could send an email that says "your order is being processed"

but if we only want to send an email when the order status is complete, then instead of putting this `sendOrderCompletedEmail` here then a better place for now would be to put it into a `.post('save',...)` middleware in the Order model file. Or in an `updateOrder` endpoint that the frontend can call when the checkout process has been completed. I will also include a ticket for listening to a webhook from stripe directly for this usecase.

akosasante: If this is where we need the full album info in the order item, we can populate it from the saved order:

`const orderItemsWithFullAlbum = await order.populate({ path: 'orderItems.album'})`

akosasante: And the email function is currently expecting four arguments (to, username, orderitems, and total), so we'll want to pass in something like:

```
await sendOrderCompletedEmail(user.email, user.username, orderItemsWithFullAlbum.orderItems, orderItemsWithFullAlbum.total);
```

src/models/Order.js
```diff
@@ -16,6 +16,13 @@ const OrderItemSchema = new mongoose.Schema({
   },
 });
 
+// Add a virtual property to populate the album with all its properties
+OrderItemSchema.virtual('fullAlbum').get(function () {
+  return this.populate('album').execPopulate().then((doc) => {
+    return doc.album;
+  });
+});
+
```
akosasante: I don't think this will work; I'm getting the following error when I try to access the virtual property (order.orderItems\[0\].fullAlbum):

```
`Error while processing the order: Error: Mongoose does not support calling populate() on nested docs. Instead of `doc.arr[0].populate("path")`, use `doc.populate("arr.0.path")`
```

src/mailing/templates/orderCompleted/html.ejs
```diff
@@ -0,0 +1,42 @@
...
+                        </td>
+                    </tr>
+                    <tr>
+                        <td>
+                            <h1>Hello <%= username %>,</h1>
+                            <p>Thank you for purchasing in our music store!!!</p>
+                            <p>Here's a list of albums you bought:</p>
+                            <ul>
+                                <% orderItems.forEach(function(item) { %>
+                                    <li><%= item.album %> - <%= item.quantity %> x $<%= item._id %> </li>
```
akosasante: `item.album` will be a full album object:



and does the user need to see the orderItem.\_id or album.\_id ?

src/mailing/templates/test/html.ejs
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Email Content</title>
+    <link rel='stylesheet' href='./style.css' data-inline>
```
akosasante:
```suggestion
    <link rel='stylesheet' href='./test/style.css' data-inline>
```

src/controllers/orderController.js
NatachaKey: @akosasante just to be sure- so if this part fails 
image
setInterval inOrder.js will trigger that this order will be deleted both from Order collection and PurchasedAlbum collection?
image

src/controllers/orderController.js
NatachaKey: @akosasante created a `.post('save',...) middleware ` but I've been testing and it doesn't send email when the orderStatus is changed to 'complete' manually (directly in mongo db)- so i added he call of this save() middleware on line 62 in orderController.js -but it also doesn't help to send an email...

src/controllers/orderController.js
NatachaKey: @akosasante I was trying to get the whole object ... now added populate method to Order.js to .post('save') middleware 
```

// Middleware to send order completion email when the order status changes to "complete"
OrderSchema.post('save', async function (doc) {
  if (doc.orderStatus === 'complete') {
    const orderItemsWithFullAlbum = await doc
      .populate({ path: 'orderItems.album' })
      .execPopulate();

    // Send the order completion email
    await sendOrderCompletedEmail(
      doc.user.email,
      doc.user.name,
      orderItemsWithFullAlbum.orderItems,
      orderItemsWithFullAlbum.total
    );
  }
});
```

src/controllers/orderController.js
NatachaKey: @akosasante not sure what I'm doing wrong but trying this way I still  don't get album populated 
image
image

NatachaKey: also noticed that if i create several orders with complete status- it sends email only to the oldest one  (I was testing it by sending different quantites)

src/models/Order.js
```diff
@@ -115,6 +117,30 @@ OrderSchema.pre('findOne', async function (next) {
   console.log('Pre-findOne finished');
   next();
 });
+
+// Middleware to send order completion email when the order status changes to "complete"
+OrderSchema.post('save', async function (doc) {
+  try {
+    if (doc.orderStatus === 'complete') {
```
akosasante: I didn't notice until now that we created both a `payment_successful` and a `complete` status. We probably only need one of those (or at least I can't picture currently what the difference between the two would be or in what case we would go from one to the next).

akosasante: An issue with this check is that `doc.orderStatus === "complete"` could be true even if the order status wasn't just updated in this query.

For example, imagine that we have an order that was already completed from weeks ago, but a request comes in to update the order total (maybe there was some bug or glitch that we're fixing now). The `post('save')` middleware would be called, and `doc.orderStats === 'complete'` would pass since the doucment's orderstatus does in fact match. Instead, we want to check if the query that just ran was to update the orderstatus specifically.

I'm not sure if you can do this with `post('save')` but  I know with `post('findOneAndUpdate')` we could check which parameters were passed to the update query with : `this.getUpdate().$set.orderStatus === 'payment_successful'`

src/models/Order.js
```diff
@@ -115,6 +117,30 @@ OrderSchema.pre('findOne', async function (next) {
...
+      const user = await User.findById(doc.user); // Assuming you have a User model
```
akosasante: Since we have the user field in the order schema, we can get this info by calling `populate` instead of a separate user query (see next comment)

src/models/Order.js
```diff
@@ -115,6 +117,30 @@ OrderSchema.pre('findOne', async function (next) {
...
+
+      const orderItemsWithFullAlbum = await doc
+        .populate({ path: 'orderItems.album' })
+        .execPopulate();
```
akosasante: `execPopulate` is not a function. It was removed in mongoose v6 (we are using v7): [https://github.com/Automattic/mongoose/blob/f552fe286a42373e50970072352b66a1fd443a1c/docs/migrating_to_6.md#removed-execpopulate](https://github.com/Automattic/mongoose/blob/f552fe286a42373e50970072352b66a1fd443a1c/docs/migrating_to_6.md#removed-execpopulate)

We can populate by calling `await` + `populate(...)` and we can chain populate calls by using an array. This allows us to populate both the full user object and the full orderItems.album objects:

```javascript
const orderItemsWithFullAlbum = await doc.populate(['user', { path: 'orderItems.album' }])
```

src/models/Order.js
```diff
@@ -115,6 +117,30 @@ OrderSchema.pre('findOne', async function (next) {
...
+
+      // Send the order completion email
+      await sendOrderCompletedEmail(
+        user.email,
+        user.username, // Use the appropriate field for the username
```
akosasante: If we populate the order user with previous comment then these lines can instead be:

```suggestion
        orderItemsWithFullAlbum.user.email,
        orderItemsWithFullAlbum.user.username, // Use the appropriate field for the username
```

src/mailing/templates/orderCompleted/html.ejs
```diff
@@ -0,0 +1,42 @@
...
+                        </td>
+                    </tr>
+                    <tr>
+                        <td>
+                            <h1>Hello <%= username %>,</h1>
+                            <p>Thank you for purchasing in our music store!!!</p>
+                            <p>Here's a list of albums you bought:</p>
+                            <ul>
+                                <% orderItems.forEach(function(item) { %>
+                                    <li><%= item.album %> - <%= item.quantity %> x $<%= item.price %> </li>
```
akosasante: `item.album` will return a full object; so we'll likely want to just show the album name

`item` is an order item, we don't have the price on that schema, so `item.price` won't return anything. The price will be available on the `album` object itself.

src/mailing/sender.js
```diff
@@ -1,7 +1,7 @@
 const nodemailer = require('nodemailer');
 const Email = require('email-templates');
 const path = require('path');
-
+const ejs = require('ejs');
```
akosasante: Is this import needed?

src/mailing/orderCompleted.js
akosasante, line 1: This dev script works well to allow us to manually test the act of sending an orderCompleted email. If we want a script that allows us to test the flow of updating an order, resulting in sending an email, we could create a separate one for that:

```javascript
// <backend repo>/updateOrderToPaidScript.js 

require('dotenv').config();
const { Order } = require('./src/models/Order');
const { connectDB } = require('./src/expressServer');

async function run() {
	try {
		// get arguments from command line
		const args = process.argv.slice(2);
		// script should be run by calling this file with order id as the first argument, eg: `node payOrderScript.js <orderId>`
		const orderId = args[0];

		if (!orderId) {
			console.error('Please provide an order id as an argument.');
			process.exit(1);
		}

		await connectDB(process.env.MONGO_URL);

		// update order status to payment_successful
		// this should also result in an email being sent due to the post-save middleware in Order.js
		const updatedOrder = await Order.findByIdAndUpdate(orderId, { orderStatus: "payment_successful" }, { new: true })

		console.log('Order updated successfully.\n', updatedOrder);

		process.exit(0);
	} catch (error) {
		console.error(error);
		process.exit(1);
	}
}

run()
```

src/controllers/orderController.js
```diff
@@ -5,58 +5,77 @@ const { BadRequestError } = require('../errors');
...
-  // Create an Order document in the database
-  const order = await Order.create({
-    user: req.user.userId,
-    orderItems,
-    subtotal,
-    tax,
-    total,
-  });
+    order.paymentIntentId = paymentIntent.id;
+    //this is where the payment happens? so if it's successful -we change orderStatus to'complete'? then we should save this new status is a db - that triggers a save middleware in Order.js that sends email of order being paid?
```
akosasante: Not quite; so currently, the flow with Stripe on the backend and front end is as shown in the image in this ticket: [https://codethedreamteam4-2023.atlassian.net/browse/PS-90](https://codethedreamteam4-2023.atlassian.net/browse/PS-90)

**Step 1 (Frontend):**

*   User clicks "Proceed to Checkout" on their cart/order summary. The frontend directs them to the `/checkout` page, and will make a POST request to the `/api/v1/orders` endpoint

**Step 2 (Backend):**

*   Server will _**create**_ an order in the **pending** state. 
*   Server will tell Stripe that an order is being made (creating a "paymentIntent" object in Stripe)
*   Stripe will give the server the `paymentIntent` object. The server will _**save**_ the `id` of this object in the Order document so that we can easily associate the two. 
*   The server returns the order object, and the `clientSecret` from Stripe's paymentIntent object as part of the response

**Step 3 (Frontend):**

*   Frontend will pass the `clientSecret` to Stripe. This is how Stripe authenticates and knows that this frontend is legitimately handling a Stripe payment for the user
*   Frontend shows payment page to the user
*   User inputs their payment info (eg: credit card)
*   User submits their payment
    *   _**Note:**_ If the user doesn't submit their payment (they forgot or leave the tab for too long; eventually our `setInterval` in the backend will mark the order as 'cancelled' since they've abandoned it for too long)
*   Frontend sends the user payment info to Stripe (along with the `clientSecret` and a `redirectUrl`)
*   After Stripe processes the payment, it will cause the browser to redirect to the checkout complete page (which we gave as the `redirectUrl`)

**Step 4 (Frontend):**

*   Frontend will call `GET /api/v1/orders/:id` to get order details and will ask Stripe for the paymentIntent so that it can tell the user if the payment was successful or not. 

So for your question of "how does the backend know if the order payment is successful" we have at least two options:

**Option 1: Frontend tells the server**

*   At **Step 4** the frontend has info from Stripe about whether the purchase was successful or not. At this point it can make a `PATCH /api/v1/orders/:id` call to update the orderStatus to `payment_successful`
    *   We don't currently have this endpoint implemented in the backend
    *   We don't currently have a ticket to do this work in the frontend

**Option 2: Backend listens to a webhook from Stripe**

*   This would be this ticket: https://codethedreamteam4-2023.atlassian.net/browse/PS-202

src/mailing/orderCompleted.js
NatachaKey: @akosasante Akos, I don't undestand quite well this part- how would we test this script? does it listens to changes  like nodemon in teminal and if it detects the new order : **_orderId_**  with orderStatus: "payment_successful"  it sends an email because in Order.findByIdAndUpdate we added  function _sendOrderCompletedEmailToUser_ ?

src/mailing/sender.js
NatachaKey: @akosasante no, deleted it, thank you!!!

src/controllers/orderController.js
NatachaKey: @akosasante Akos, thank you for such a great explanation!!! finally undestood it and assigned to myself PS 202 to see how his stripe hook works

NatachaKey: https://stripe.com/docs/api/payment_intents/object - stripe docs

NatachaKey:
@akosasante Akos, I couldn't test manually the email for paid orders- I tried different ways - first i run 
image
while I have in html.ejs the following: 
<ul>
                                <% orderItems.forEach(function(item) { %>
                                    <li><%= item.album.albumName%> - <%= item.quantity %> x $<%= item.album.price %> </li>
                                <% }); %>
                                <li>Total amount: $<%= total %> </li>
                            </ul>

image
I populated orderItems.album object as well as user object here on line 45 
image
but i get an error 
image
it says I get null (reading 'albumName') so when in html.ejs I put item.album - at least I get the whole object, but can't get item.album.albumNime and item.album.price properties and don't know why 
image
image
Could you please take a look why I don't get them? 
I also lost access to **username** variable  in html.ejs it shows nothing right now...
I mean if I am calling **node ...orderCompleted.js** in teminal I have to check what I pass there as arguments in sendOrderCompletedEmail() - because general fn is:
image
and I pass this 
image
I was tryig to access albumNames at leat here and I could reach them- not sure why i can't reach them in html.ejs:
image

src/controllers/orderController.js
akosasante: If the `stripe.paymentIntents.create` part fails from your first screenshot, that will raise an error. 

Since we are in a `try/catch` block, that means Node will proceed to the `catch` block on line 69-76. 

Then, on line 71 we tell mongo to `abortTransaction` which means it will "rollback" (that is, undo) all database changes that occurred in this transaction/session (so everything after line 11 where we started the transaction, this includes undoing the order creation and the purchased album creation). 

The `setInterval` function in `Order.js` does not come into play here at all. That function is happening independently just checking periodically for any orders that have been in `'pending'` state for too long. It updates those old pending orders to `'cancelled'`. Then separately, because there is a TTL (time-to-live) index on the `orderStatus` column, any orders that are in `'cancelled'` state for too long (currently set to 7200 seconds/2 hours) are deleted completely from the database.

src/controllers/orderController.js
akosasante: Changing the status directly in MongoDB will _**not**_ trigger any mongoose middleware, because mongoose is a separate library from the MongoDB itself.

akosasante: We can either (1) use a node script to update the `orderStatus` of an order to `complete` using mongoose (Order.update) (just for dev testing purposes), or (2) add an `updateOrder` endpoint/controller function that we can use to send an update request via Postman, or (3) update to `complete` based on Stripe webhook events

src/mailing/orderCompleted.js
akosasante: No this script would be meant for doing dev/manual testing of the "send an order completed email" function. It is not meant to be unit tested nor will it be run by the application automatically. It is meant to be run by a dev manually to verify that the whole flow of sending an email works when an order status is updated to `payment_successful`. 

You would run the script by calling `node updateOrderToPaidScript paste-some-order-id-from-the-database-here` command in your terminal

src/models/Order.js
```diff
@@ -1,4 +1,6 @@
 const mongoose = require('mongoose');
+const { sendOrderCompletedEmail } = require('../mailing/sender');
+const User = require('../models/User');
```
akosasante: require `User` is not needed

src/models/Order.js
```diff
@@ -115,6 +111,31 @@ OrderSchema.pre('findOne', async function (next) {
   console.log('Pre-findOne finished');
   next();
 });
+
+// send order completion email when the order status changes to "payment_successful"
+OrderSchema.post('findOneAndUpdate', async function (doc) {
+  try {
+    if (this.getUpdate().$set.orderStatus === 'payment_successful') {
+      const user = await User.findById(doc.user);
```
akosasante: We don't use this variable, so this line (`const user = await ...`) can be removed

src/mailing/sender.js
```diff
@@ -33,6 +33,8 @@ const baseEmail = new Email({
     webResources: { relativeTo: path.resolve('src/mailing/templates') },
   },
 });
+console.log(path.resolve('../templates'));
+console.log(process.cwd());
```
akosasante: can be removed

src/mailing/sender.js
```diff
@@ -33,6 +33,8 @@ const baseEmail = new Email({
...
 
 const subject = 'First Subject of the first email to be sent';
```
akosasante: this `subject` variable is not used anywhere; we can remove it

src/mailing/templates/orderCompleted/html.ejs
akosasante, line 1: Right now, if I try to test this by running the function: `node updateOrderToPaidScript.js 64fcdc35d7349bd5eff622e2`, I get the following previewed email:



looks pretty good, the main thing to fix is the part where we loop over the `orderItems` as right now we're getting the full ordered album object.

src/mailing/templates/orderCompleted/html.ejs
```diff
@@ -0,0 +1,42 @@
...
+                        </td>
+                    </tr>
+                    <tr>
+                        <td>
+                            <h1>Hello <%= username %>,</h1>
+                            <p>Thank you for purchasing in our music store!!!</p>
+                            <p>Here's a list of albums you bought:</p>
+                            <ul>
+                                <% orderItems.forEach(function(item) { %>
+                                    <li><%= item.album%> - <%= item.quantity %> x $<%= item.album %> </li>
```
akosasante:
```suggestion
                                    <li><%= item.album.albumName%> - <%= item.quantity %> x $<%= item.album.price %> </li>
```

^ we should access the album name and album price instead of the full album object

src/controllers/orderController.js
akosasante: :+1:  remember to remove this line 61 comment to avoid future confusion

src/controllers/orderController.js
```diff
@@ -5,58 +5,78 @@ const { BadRequestError } = require('../errors');
...
+    await order.save({ session });
+
+    await session.commitTransaction();
+    //Frontend will pass the clientSecret (he gets from here)to Stripe. This is how Stripe authenticates and knows that this frontend is legitimately handling a Stripe payment for the user
+    res
+      .status(StatusCodes.CREATED)
+      .json({ clientSecret: paymentIntent.client_secret, order });
+  } catch (error) {
+    // Handle errors and rollback the transaction if it fails
```
akosasante:
```suggestion
    // Handle errors and rollback the transaction if any part of the try/catch fails
```

src/mailing/orderCompleted.js
akosasante, line 1: I think you're using this function in a similar manner as the `updateOrderToPaidScript.js` for testing out the email functionality. I think the only difference is this one works by finding the most recent successful order for a given user id and then specifically calling the email function for it, vs the `updateOrderToPaidScript.js` takes a given order id and updates that order to `payment_successful` which kicks off the mongoose middleware that will send the email. 

I don't think we need to have both (you can choose which one to keep or to keep both if you want); and it's generally better to keep manually run scripts _**outside**_ of our `src` directory to reduce confusion about what files are being run by us the developers, vs what files are expected to be run by the Node/Express app itself. So if you choose to keep this one; just make sure to move out to the root of the repo instead.

src/controllers/orderController.js
NatachaKey: @akosasante THANK YOU!!!!

src/controllers/orderController.js
NatachaKey: @akosasante I took https://codethedreamteam4-2023.atlassian.net/browse/PS-202 , a 3d option

src/mailing/orderCompleted.js
NatachaKey: image
result: script ran

NatachaKey: @akosasante Akos, I almost got it, I got errors because of mongo/models differences that now seem to be fixed , now the only issue I am having is that post middleware is not being triggered after updating order status ( by running node updateOrderToPaidScript numberOfOrder) -not sure why that happens
image

orderCompleted.js
```diff
@@ -0,0 +1,78 @@
...
+    }
+    const username = user.username;
+    console.log('User ID:', user._id);
+    // Check if the user has an order with the 'payment_successful' status
+    const order = await Order.findOne({
+      user: user._id,
+      orderStatus: 'payment_successful',
+    });
+
+    console.log('ORDER:', order._id);
```
akosasante: This can fail if `order` is undefined/null if no matching order was found in line 34. Can either move this line until after the if block lines 40-44 because at that point we _know_ we've found an order. Or we can use optional chaining here `order?._id`

src/models/Order.js
```diff
@@ -137,6 +134,30 @@ OrderSchema.pre('findOne', async function (next) {
   console.log('Pre-findOne finished');
   next();
 });
+
+//AKOS: this post middleware is not being triggered after updating order status (node updateOrderToPaidScript numberOfOrder)
+// send order completion email when the order status changes to "payment_successful"
+OrderSchema.post('findByIdAndUpdate', async function (doc) {
```
akosasante: Looks like the correct middleware name is actually `findOneAndUpdate` (it should be triggered also by calls to `findByIdAndUpdate` because under-the-hood Mongoose will actually just be calling `findOneAndUpdate` for that function anyway): `https://mongoosejs.com/docs/api/query.html#Query.prototype.findOneAndUpdate()`

src/controllers/orderController.js
```diff
@@ -5,58 +5,89 @@ const { BadRequestError } = require('../errors');
...
+    const orderItemsWithFullAlbum = await order.populate({
+      path: 'orderItems.album',
     });
+    // Send the order completion email after THE ORDER HAS STATUS COMPLETE
+    await sendOrderCompletedEmail(
+      user.email,
+      user.name,
+      orderItemsWithFullAlbum.orderItems,
+      orderItemsWithFullAlbum.total
+    );
```
akosasante: We can remove this from here because we don't want to send the email at this stage yet. At this point, the payment intent has just been submitted to Stripe; but the user hasn't actually put in the payment information and we don't know if the payment will be successful. So we likely don't want to send a "your order has completed" email. We _could_ have a separate email that says "we got your order request, please proceed to checkout" but that may get a bit too busy since the user could still abandon their order at this point of the process.

src/controllers/orderController.js
```diff
@@ -5,58 +5,89 @@ const { BadRequestError } = require('../errors');
 const PurchasedAlbum = require('../models/PurchasedAlbum');
 const CustomError = require('../errors');
 const { checkPermissions } = require('../utils');
-
+const mongoose = require('mongoose');
+const { sendOrderCompletedEmail } = require('../mailing/sender');
```
akosasante: We can remove this :+1:

src/models/Order.js
NatachaKey: In Mongoose, 
The findOneAndUpdate function is used to find a single document that matches a specified condition and update it. It returns the updated document.
findByIdAndUpdate:

The findByIdAndUpdate function, on the other hand, is used to find a document by its unique identifier (usually the _id field) and update it. It also returns the updated document.
//here we set a condition ourselves  `if (this.getUpdate().$set.orderStatus === 'payment_successful')`, correct middleware name is `findOneAndUpdate`

src/controllers/orderController.js
NatachaKey: @akosasante oh that totally makes sence!! thank you!

NatachaKey:
@akosasante Akos, when I change  orderStatus to  payment_successful, doing node updateOrderToPaidScript xxx , the terminal indicates that the order status was updated - but if I want to test the email being send automatically (now I can't do that)-I should complete first this ticket https://codethedreamteam4-2023.atlassian.net/browse/PS-202 , right? if so, can i merge this one?

src/mailing/orderCompleted.js
akosasante: Nice! And did you get an email preview from that as well?

src/models/Order.js
akosasante: Yes; to clarify, in the code we'd still call `findByIdAndUpdate` but the middleware must be called `findOneAndUpdate`: [https://stackoverflow.com/a/73215258](https://stackoverflow.com/a/73215258)  
 

> `findByIdAndUpdate()` is essentially a wrapper for `findOneAndUpdate()`. That is, it is _equivalent to_ `findOneAndUpdate({ _id: id }, ...)`.
> 
> As such, it **does not bypasses** mongoose middleware completely.
> 
> As per the [documentation](https://mongoosejs.com/docs/api.html#model_Model-findByIdAndUpdate):
> 
> > This function triggers the following middleware.
> > 
> > *   `findOneAndUpdate()`

src/controllers/orderController.js
```diff
@@ -64,7 +65,7 @@ const createOrder = async (req, res) => {
     const orderItemsWithFullAlbum = await order.populate({
       path: 'orderItems.album',
     });
```
akosasante: can be removed?

akosasante (approved):
Looks good to go :+1:

src/models/Order.js
NatachaKey: @akosasante thank you! 
link to docs: https://mongoosejs.com/docs/api/model.html#Model.findByIdAndUpdate()

src/controllers/orderController.js
NatachaKey: sure! done

PR #59, opened 2023-09-08
Ps190 PS191 notifications order cancellations

mvstoyan:


mvstoyan:
@akosasante I don't understand where I have an error in the path; I constantly receive 'Route does not exist

src/expressServer.js
```diff
@@ -140,6 +140,11 @@ const socketServer = new Server(server);
 const setupSocket = require('./live');
 const io = socketServer.of('/'); // Create an instance of Socket.io
 
+// Add this route to serve the order notifications page
+app.get('/order-notifications', (req, res) => {
+  res.render('orderNotifications'); // Render the EJS template
+});
+
```
akosasante: Express reads routes and middleware from top-to-bottom.

So when you type in `localhost:8000/order-notifications` in the browser, it's gonna check in this file each line that says `app.use` or `app.<httpMethod>`. 

*   it doesn't match anything from line 119-127
*   it doesn't find a matching file in the `public` static folder
*   next thing it gets to is `app.use(notFoundMiddleware)` on line 131. That middleware runs for anything that hasn't been matched up to that point, which is why you get the "Route does not exist" page. That's why we say that that one and the errorHandlerMiddleare "must be defined after all other routes and middleware".

mvstoyan:
@akosasante Here is PS190-PS191 tickets together. I trying test Postman. Did I do correct? 
screen

.eslintrc.json
```diff
@@ -17,5 +17,6 @@
     "node": true,
     "es2023": true,
     "jest/globals": true
-  }
+  },
+  "ignorePatterns": ["src/public/orderNotifications.js"]
```
akosasante: I think you added this because without it, there's a lint error in that file like this:

```
/Users/aasante/dev/practicum2023/backend/src/public/orderNotifications.js
  1:21  warning  'location' is defined but never used  no-unused-vars
```

and this is because we've told ESLint in this configuration file to expect that the code its inspecting will be code from `node` (so it knows to expect node global variables like "require" or "module.epxorts" without those needing to be explicitly imported) and `ES2023` (so it knows to expect modern javascript like async/await), and `jest/globals` (so it knows to expect Jest global variables like "describe" and "beforeAll" without them being explicitly imported) .

Rather than skipping linting on that file we have some other options:

1.  Add `"browser": true` to the `env` block on line 16-20. This tells ESLint that some files are browser Javascript instead of node Javascript. Which means it will know to expect global variables like "document" and "location" without those needing to be imported.
2.  Tell it in those files which variables are implicitly global. It looks like you took this approach here for example: ed54b17

src/live/emitters.js
```diff
@@ -0,0 +1,5 @@
+function sendOrderCancelledEvent(io, payload) {
+  io.to(payload.userId).emit('orders:cancelled', payload.message);
```
akosasante: Instead of asking the caller to import `io` and pass it to this function; I think it makes more sense to have this file/module import `io` itself. Since this is where all the socket interaction is happening.

```suggestion
const { io } = require("../expressServer");

function sendOrderCancelledEvent(payload) {
  io.to(payload.userId).emit('orders:cancelled', payload.message);
```

akosasante (commented):
Great start! I know the EJS implementation notes were pretty sparse; so I've left some additional direction here in comments. The end result should be an Order Notifications page that will show the message in the browser (rather than just the browser console) either in an HTML element of your choice or using a third-party library to show toast messages.

akosasante:
## Additional Implementation Notes

We now have an Order Notification page that successfully connects to the backend via Socket.IO/Websockets and is listening for messages for the `orders:cancelled` event. Some next steps:

#### Orders.js

*   Currently, when the `updateOrderStatus` runs, it updates _any_ orders that have been pending for too long. These may be for multiple different users. And a single user _could_ have multiple abandoned orders (currently not really possible via the frontend but definitely possible if you call the create order API a bunch of times). So that means we won't have a single `userId` to send the socket message to. We could have multiple. But we also don't want to send multiple notifications to one user if they have many orders that were cancelled at the same time.
*   So we can handle this by getting the orders that are being updated and grouping them by user. Unfortunately, the `updateMany` function does _not_ return the documents that were updated. Instead it returns an object with some data of how many documents matched and how many documents were updated. That means we have to fetch the orders that we're about to update _first_ before we actually do the update. 
*   This is a good use-case for `aggregate` since that lets us do grouping within the database, instead of having to do it in the JS code ourselves. It also means we don't have to worry about accidentally triggering the `find` middleware as we would if we try to just do a normal `Order.find` call ([example with raw MongoDb queries but should be able to translate to mongoose syntax](https://studio3t.com/knowledge-base/articles/mongodb-aggregation-framework/))

```javascript
const ordersToUpdate = await Order.aggregate([
  { $match: { ...describe the fields you want to match. this should be the same as the properties passed to the updateMany call } },
  { $group: { ...describe what field to group by, and use $push to create an array of orders for each group } }
]);


// After the updateMany call is successfully run, call the emitter function to send a message about these orders:
sendOrderCancelledEvent(ordersToUpdate);
```

#### live/index.js

*   One thing that I missed including in the [ticket,](https://codethedreamteam4-2023.atlassian.net/browse/PS-191) the instructions in the first ticket lead you to emit a `join:user_notifications` event from the Order Notification page. We need to actually handle that event in the server so that we create a room and make that socket instance join that room so that they can receive any future messages for that room ([Room docs](https://socket.io/docs/v3/rooms/))

```javascript
  socket.on('join:user_notifications', (userId) => {
    console.log('User joined room for notifications:', userId);
    socket.join(userId);
  });
```

*   To better keep things organized, you may want to put that in its own file instead of directly in index.js (like what we've done with `testPing`

#### emitters.js

*   the Order module is going to call the `sendOrderCancelledEvent` here with an object that has the `ordersToUpdate`

```javascript
function sendOrderCancelledEvent(ordersToUpdate) {
  console.log("ordersToUpdate: ", ordersToUpdate);
  
  // loop through the ordersToUpdate array, each element should be something like { _id: <userId>, orders: <array of orders> }
  
  // for each user in the array, send a message to the room for that user id, like this:
  global.io.to(userId.toString()).emit('orders:cancelled', `${numOrders} of your ${orderNoun} have been cancelled.`);
}
```

*   The reason we're using `global.io` here is because if we try to do `const { io } = require("../expressServer")` then we end up with a circular dependency.That's because `expressServer.js` file is first opened by node when the `app.js` file requires it. Then `expressServer.js` will import `Orders.js` (via `orderController` via `orderRoutes`), which will import `emitters.js`, which tries to import `io` from `expressServer` but `io` hasn't yet been set up in there because node hasn't gotten to that line. So in order to avoid all that complicatedness :sweat:  we can make the `io` variable a global (see next section). It is rarely good practice to do this and there are some other approaches like "dependency injection" or creating a "singleton object" but for now this is a reasonable way forward

#### express.js

As described in the previous bullet point:

```javascript
// Set up Socket.io connection event
io.on('connection', (socket) => {
  console.log('Connected');
  setupSocket(io, socket); // Call your setupSocket function

  socket.on('msg_from_client', function (from, msg) {
    console.log('Message is ' + from, msg);
  });
  socket.on('disconnect', function () {
    console.log('Disconnected');
  });
});

// add the io instance to the global object so that it can be used in other modules without circular dependency
global.io = io;
```

#### orderNotifications.js

*   Currently you have the user ID hardcoded. One way we can make it a little more dynamic is by having the user id be defined in the URL via a query parameter

```javascript
  const queryParams = new URLSearchParams(window.location.search);
  const userId = queryParams.get('userId')
  
    socket.on('connect', function () {
    console.log('CONNECTED: ', this); // should log the socket object

    // send a message to the socket server to tell it to add us to a room for `userId` notifications
    this.emit('join:user_notifications', userId);
```

*   In addition to (or instead of) logging to console when we receive an order:cancelled message, we can also append to the DOM to see the result in the browser itself:

```javascript
// subscribe to orders:cancelled messages
    this.on('orders:cancelled', (response) => {
    
    // you can expand on this by using actual html elements with classes instead of just a boring text node :'D 
    document.body.appendChild(document.createTextNode(response));
```

*   For extra fun, you could try using a third-party library like [Toastify](https://github.com/apvarun/toastify-js) to create a snackbar/toast that will pop up to show the received message

src/public/orderNotifications.js
```diff
@@ -0,0 +1,15 @@
+const io = require('socket.io-client');
```
akosasante: `require` is only available within node.js files themselves, this Javascript file is not running on our node server but rather in the client/browser. 

The socket.io server-side npm package, automatically creates a public/static path for the socket.io-client: [https://socket.io/docs/v4/client-installation/#:~:text=io%20will%20be%20registered%20as%20a%20global%20variable%3A](https://socket.io/docs/v4/client-installation/#:~:text=io%20will%20be%20registered%20as%20a%20global%20variable%3A)

That is how we were able to import it in the HTML: `<script src="/socket.io/socket.io.js"></script>`

Now that `io` variable is available globally in any script loaded by this HTML. So we don't need to try to require or import anything, we can go ahead and use it right away.

If you get a lint error because ESLint doesn't realize that `io` will be available globally (or any other third-party scripts you add to the HTML document), then you can add a comment at the top like this: `/* global io */`

.eslintrc.json
mvstoyan: Thank you, added "browser": true

mvstoyan:
Order.js 
screen3

mvstoyan:
@akosasante I'm stuck and can't figure out how to fix the problems related to ordersToUpdate when it's passed to the sendOrderCancelledEvent function. Additionally, I'm not sure if I've correctly resolved the circular dependency issue. Can you please help me understand and solve these issues?

src/expressServer.js
```diff
@@ -140,6 +145,7 @@ const server = http.createServer(app);
 const socketServer = new Server(server);
 const setupSocket = require('./live');
 const io = socketServer.of('/'); // Create an instance of Socket.io
+global.io = io; // add the io instance to the global object
```
akosasante: Only one of these is needed; can remove this line and leave 164 as is.

src/live/emitters.js
```diff
@@ -0,0 +1,15 @@
...
+function sendOrderCancelledEvent(ordersToUpdate) {
+  console.log("ordersToUpdate: ", ordersToUpdate);
+  
+  for (const userData of ordersToUpdate) {
+    const userId = userData._id.toString();
+    const numOrders = userData.orders.length;
+    const orderNoun = numOrders === 1 ? 'order' : 'orders';
+    
+    io.to(userId).emit('orders:cancelled', `${numOrders} of your ${orderNoun} have been cancelled.`);
+  }
```
akosasante: Here we get that circular dependency I warned of, when we start the app. So `io` will be undefined and this message won't get emmitted :( 

This is where we want to use the `io` variable that we stuck in the `global` object:

```suggestion
function sendOrderCancelledEvent(ordersToUpdate) {
  console.log("ordersToUpdate: ", ordersToUpdate);
  
  for (const userData of ordersToUpdate) {
    const userId = userData._id.toString();
    const numOrders = userData.orders.length;
    const orderNoun = numOrders === 1 ? 'order' : 'orders';
    
    global.io.to(userId).emit('orders:cancelled', `${numOrders} of your ${orderNoun} have been cancelled.`);
  }
```

src/models/Order.js
```diff
@@ -78,13 +79,34 @@ const updateOrderStatus = async () => {
...
+      await Order.updateMany(
+        { _id: { $in: userOrders.orders.map((order) => order._id) } },
+        { $set: { orderStatus: 'cancelled' } }
+      );
+
+      // Emit a message for the user
+      sendOrderCancelledEvent(io, {
+        userId: userOrders._id,
+        message: 'Your order has been cancelled.',
+      });
```
akosasante: `sendOrderCanceleldEvent` in `emitter.js` is only expecting 1 argument, the `userOrders` object of orders to emit info about.

```suggestion
      sendOrderCancelledEvent(userOrders);
```

src/models/Order.js
```diff
@@ -1,5 +1,6 @@
 const mongoose = require('mongoose');
-
+const { sendOrderCancelledEvent } = require('../live/emitters');
+const io = require('../expressServer');
```
akosasante: This will still be a circular dependency and this line should be removed:

1.  expressServer loaded by the app.js file
2.  Order.js loaded by order controller by the expressServer
3.  At this point on line 3 we try to import `io` but it's not been defined/set up yet in expressServer, so we'll just get undefined
4.  After that in expressServer we set up io and export it; but this doesn't update any previous places that tried to import the variable too early

src/models/Order.js
```diff
@@ -119,7 +141,7 @@ OrderSchema.pre('findOne', async function (next) {
 OrderSchema.index(
   { expiresAt: 1 },
   {
-    expireAfterSeconds: 7200,
+    expireAfterSeconds: 60, //7200
```
akosasante: Any reason to change this or just for testing?

src/views/orderNotifications.ejs
```diff
@@ -0,0 +1,15 @@
...
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Order Notifications</title>
+    <script src="/socket.io/socket.io.js"></script>
+  </head>
+  <body>
+    <h1>Order Notifications</h1>
+    <div id="notification-container"></div>
```
akosasante: Is this div used anywhere for a particular use case?

src/live/emitters.js
```diff
@@ -0,0 +1,15 @@
+const { io } = require("../expressServer");
+
+function sendOrderCancelledEvent(ordersToUpdate) {
+  console.log("ordersToUpdate: ", ordersToUpdate);
+  
+  for (const userData of ordersToUpdate) {
+    const userId = userData._id.toString();
+    const numOrders = userData.orders.length;
+    const orderNoun = numOrders === 1 ? 'order' : 'orders';
```
akosasante: In Order.js, we loop through the aggregated \`ordersToUpdate\` and pass in just the object for a given user's orders to this emitter.js function. So we expect this function to receive an object that looks like:

```javascript  
{   
 _id: <user id>,  
 orders: <list of orders that have been cancelled for this user for being abandoned too long>  
}  
```

So there is nothing to iterate/loop over here, we can just access these variables directly from the object passed in:

```suggestion  
function sendOrderCancelledEvent(userData) { 

  console.log("ordersToUpdate: ", userData);  
   
  const userId = userData.\_id.toString();  
  const numOrders = userData.orders.length;  
  const orderNoun = numOrders === 1 ? 'order' : 'orders';  
```

src/public/orderNotifications.js
```diff
@@ -0,0 +1,32 @@
...
+    // send a message to the socket server to tell it to add us to a room for `userId` notifications
+    this.emit('join:user_notifications', userId);
+
+    // subscribe to orders:cancelled messages
+    this.on('orders:cancelled', (response) => {
+    
+      // you can expand on this by using actual html elements with classes instead of just a boring text node :'D 
+      document.body.appendChild(document.createTextNode(response));
+      // Display the message using Toastify
+      Toastify({
```
akosasante, line 20: We need to import the Toastify library in order to use it. In this case, we would do that by adding a script tag to the `.ejs` file so that the browser downloads the library when the page loads.

Otherwise we get this error:



Per the instructions in the [Toastify README](https://github.com/apvarun/toastify-js#adding-toastifyjs-to-html-page-using-the-traditional-method), we can do this by importing the JS script and CSS file for Toastify from a CDN (a cloud URL):

```
<link rel="stylesheet" type="text/css" href="https://cdn.jsdelivr.net/npm/toastify-js/src/toastify.min.css">
<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/toastify-js"></script>
```

^ we would put those lines in the `orderNotifications.ejs` file. And uninstall `toastify-js` from the npm `package.json` file.

---

Another option, is to keep the version that's installed in our server, and serve that directly to the browser.

To do that we need to set up a static path to tell Express to serve the `toastify` bundle:

So in `expressServer.js` we'd need to set up some routes for the JS file and the CSS file that are in node\_modules from installing `npm i toastify`. Something along the lines of:

```javascript
app.get('/toastify/toastify.js', (req, res) => {
  res.setHeader('content-type', 'application/javascript');
  const jsFile = readFileSync('./node_modules/toastify-js/src/toastify.js');
  res.send(jsFile)
});
app.get('/toastify/toastify.css', (req, res) => {
  res.setHeader('content-type', 'text/css');
  const cssFile = readFileSync('./node_modules/toastify-js/src/toastify.css');
  res.send(cssFile)
});
```

And in `orderNotifications.ejs` we'd import them using the URLs that we set up in `expressServer.js`:

```
    <script src='/toastify/toastify.js'></script>
    <link type='text/css' rel='stylesheet' href='/toastify/toastify.css' />
```

src/models/Order.js
mvstoyan: only for testing, I will back 7200, when everything will work for me

src/public/orderNotifications.js
mvstoyan: screen

src/live/emitters.js
mvstoyan: screen1

mvstoyan: screen5

mvstoyan: Thank you very much! I've sent screenshots of what I'm getting in the browser and terminal. I don't quite understand how to test this code in Postman. I'm establishing a connection with the socket and Listening to
order:cancelled.

.eslintrc.json
```diff
@@ -16,6 +16,8 @@
   "env": {
     "node": true,
     "es2023": true,
+    "browser": true,
     "jest/globals": true
-  }
+  },
+  "ignorePatterns": ["src/public/orderNotifications.js"]
```
akosasante: this `ignorePatterns` line can be removed

package.json
```diff
@@ -44,19 +45,22 @@
     "spotify-web-api-node": "^5.0.2",
     "stripe": "^13.0.0",
     "swagger-ui-dist": "^5.4.2",
+    "toastify": "^2.0.1",
```
akosasante: This project seems to be more dead (last update in 2020) vs the `toastify-js` one (last update in 2022): [https://github.com/georgehanson/toastify](https://github.com/georgehanson/toastify)

Any reason to have both here?

package.json
```diff
@@ -30,6 +30,7 @@
     "express-favicon": "^2.0.4",
     "express-rate-limit": "^6.8.1",
     "express-session": "^1.17.3",
+    "express-static-gzip": "^2.1.7",
```
akosasante, line 33: I am curious what this library was added for?

src/expressServer.js
```diff
@@ -1,4 +1,4 @@
-const path = require('path');
+//const path = require('path');
```
akosasante: Can be removed?

src/expressServer.js
```diff
@@ -128,6 +129,28 @@ app.use('/api/v1/wishlist', wishlistRoutes /* #swagger.tags = ['Wishlist'] */);
...
+  const jsFile = readFileSync('./node_modules/toastify-js/src/toastify.js');
+  res.send(jsFile);
+});
+
+app.get('/toastify/toastify.css', (req, res) => {
+  res.setHeader('content-type', 'text/css');
+  const cssFile = readFileSync('./node_modules/toastify-js/src/toastify.css');
+  res.send(cssFile);
+});
+
```
akosasante: If `expressStaticGzip` works the way that I think it does, then it will already be serving these files from the appropriate node\_modules forlder with the appropriate `Content-Type` headers. So lines 142 to 152 can be removed

src/expressServer.js
```diff
@@ -154,10 +177,7 @@ io.on('connection', (socket) => {
   });
 });
 
-// Serving the documentation page
-app.get('/tests', (req, res) => {
-  const documentationFilePath = path.join(__dirname, 'public', 'tests.html');
-  res.sendFile(documentationFilePath);
-});
```
akosasante, line 161: Are you intentionally removing this route for the documentation page?

test/sendEmail.test.js
akosasante, line 1: are the changes in this file needed?

akosasante (approved):
:tada:

package.json
mvstoyan: Using express-static-gzip helped resolve the issue with the error message "http://localhost:8000/node_modules/toastify-js/src/toastify.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled" that was appearing in my browser's console. Now that Toastify is functioning correctly, I no longer encounter this error. Therefore, it seems that I can probably remove it.

src/expressServer.js
mvstoyan: Since I no longer have an error in the console, do you think it's better for me to remove this code or express-static-gzip? They both works well

test/sendEmail.test.js
mvstoyan: I receive a warning if I don't comment out these lines.

src/expressServer.js
mvstoyan: returned, it by mistake

src/expressServer.js
mvstoyan: I'm leaving this in because the code below was removed by mistake

package.json
mvstoyan: removed toastify

src/live/emitters.js
akosasante: Sorry for the delay in responding to this question:

You'll want to follow the same steps as what's happening to the user in the browser on the `orderNotifications.js` page:

1.  Connect to the socket on `http://localhost:8000` :white\_check\_mark: 
2.  Emit a message to the `join:user_notifications` channel with the user ID whose notifications you want to receive. This is needed because for the order cancelled notification we only send it to the specific "room" for that user id (since we don't want other users to get notifications about your cancelled orders). I think that part is missing in your screenshot. You'll need to do this in the "Message" tab of the Postman socket request. When you send this message from Postman, then in the backend terminal you should see the following text logged: `User joined room for notifications: <your user id>`
3.  Subscribe to events for **`orders`**`:cancelled` channel. Your screenshot has it mis-spelled with the singular **`order`**`:cancelled` so that may have caused some issues testing it out. 
4.  Create an order (POST request using POSTMAN to the `/api/v1/orders` endpoint) to create a 'pending' order. Allow some time to let the backend convert that 'pending' order to a 'cancelled' one. To speed things up for local testing, I set my expiry time to 15 seconds using the `DEV_ORDER_EXPIRY_DURATION=15000` environment variable in my `.env` file.

mvstoyan: Its tested for me too, thank you very much!

PR #61, opened 2023-09-09
PATCH /api/v1/users/updateCurrentUser  test

mvstoyan:
in progress...


test/userController.test.js
```diff
@@ -50,6 +50,243 @@ afterAll(async () => {
...
+
+    // Act: Attempt to delete the user using the authenticated cookie
+    const response = await request(app)
+      .delete('/api/v1/users/deleteSingleUser')
+      .set('Cookie', [signedCookie]);
+
+    // Assert: Check the response status for success
+    expect(response.status).toBe(200);
+    expect(response.body).toHaveProperty('msg', 'Success! User Deleted.');
+  });
```
akosasante, line 218: Currently, this test is failing because it's getting a 403 instead of a 200:

```

    Expected: 200
    Received: 403
```

We know that this endpoint is restricted to `admin` role users only. So the 403 makes sense because we're currently logging in as a user (Ava) whose role is just `user`. 

I would suggest creating an admin user (either within this test or in the `beforeEach` or `beforeAll` hook) and then logging in as that admin user for this success case test. 

The current test could still be useful as a separate test for verifying that we do indeed get a 403 when the logged in user is not an admin :)

test/userController.test.js
```diff
@@ -50,6 +50,243 @@ afterAll(async () => {
...
+    // Act: Attempt to delete a user without authentication
+    const response = await request(app).delete(
+      '/api/v1/users/deleteSingleUser'
+    );
+
+    // Assert: Check the response status for an error
+    expect(response.status).toBe(401);
+  });
+});
+/*
```
akosasante: All this commented out stuff can be removed?

akosasante (commented):
Sorry I reviewed and commented on the [https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/63/files](https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/63/files) PR first. But it looks like this PR contains changes from both?

test/userController.test.js
mvstoyan: not need, removed, sorry.

test/userController.test.js
mvstoyan: Two different tickets, for write and read endpoints. This is why two PR. But i probably going to merge this one with PS145. I need my admin from PS145 now))

test/userController.test.js
```diff
@@ -58,9 +58,28 @@ afterAll(async () => {
...
+    username: 'ava',
+    email: 'ava@ava.com',
+    password: 'secret',
+    role: 'user',
+  });
+});
+
+afterEach(async () => {
+  // Delete the test user after each test
+  await User.deleteMany({});
```
akosasante, line 77: Since we already do a `deleteMany` specifically for the testUser in the `beforeEach`, why not just do this `User.deleteMany({})` in the beforeEach instead? Since it's already deleting all users

test/userController.test.js
```diff
@@ -58,9 +58,28 @@ afterAll(async () => {
...
+  await User.deleteMany({ email: testUserCredentials.email });
+
+  // Create a test user before each test
+  testUser = await User.create({
+    name: 'ava',
+    username: 'ava',
+    email: 'ava@ava.com',
+    password: 'secret',
+    role: 'user',
+  });
```
akosasante, line 72: Since we are creating the test user before each test, we can probably delete the code from `main` branch on line 45 that does the same thing at the very beginning before the tests are run (`beforeAll`)

akosasante (approved):
Looks good to me; with the remaining change needed being the `admin` role login for the delete endpoint which you've said will be done in a separate PR :+1:

test/userController.test.js
mvstoyan: jest

test/userController.test.js
mvstoyan: I merged this thread with PS145. Can we close this one? I will merge the full test with the main branch from PS145, after fixing and passing

akosasante:
Per your comment, closing this one in favour of #63

PR #63, opened 2023-09-16
[PS-145] user controller tests

mvstoyan:


test/userController.test.js
```diff
@@ -94,3 +94,143 @@ describe('GET /api/v1/users/:user_id endpoint', () => {
     expect(response.status).toBe(404);
   });
 });
+
+describe('PATCH /api/v1/users/updateCurrentUser endpoint', () => {
+  it('should update the current user', async () => {
+    const signedCookie = await loginAndReturnCookie({
+      email: 'ava@ava.com',
+      password: 'secret',
+    });
```
akosasante: We can narrow things down to only run this one single test (line 99) with the following Jest command:

`npm run test -- --testPathPattern=test/userController.test.js --testNamePattern="should update the current user"`

*   The first `--` tells npm that we're passing these command options to the underlying `jest` command.
*   `--testPathPattern` lets us give Jest a path as a string or regex that tells it only to run tests in that file (https://jestjs.io/docs/cli#--testpathpatternregex)
*   `--testNamePattern` lets us give Jest a string or regex that it will search for in the file(s) its running the test in; it will only run those tests whose name (either the `decribe` block or the `it` block) matches. We can also shorten this by using `-t="should update the current user"`

Once you do this, you will see that this test fails because the user cannot log in and get a cookie. The login endpoint is returning a `401`. We know that this happens when the credentials that were given to the login endpoint don't match any user in the database. Furthermore, the error log shown in the test output shows that the response did not return any `cookie` so the step where we try to find the `token` in the cookie fails:

```
> bb-practicum-team-4-back@1.0.0 test
> jest --watchAll=false --detectOpenHandles --silent --testPathIgnorePatterns=src "--testPathPattern=test/userController.test.js" "--testNamePattern=should update the current user"

 FAIL  test/userController.test.js
  ● PATCH /api/v1/users/updateCurrentUser endpoint › should update the current user

    TypeError: Cannot read properties of undefined (reading 'find')

      26 |   // Split the cookie string by ';' and take the first part (before the first ';')
      27 |   const signedCookie = cookieHeader
    > 28 |     .find((cookie) => cookie.startsWith('token'))
         |      ^
      29 |     .split(';')[0];
      30 |   return signedCookie; // Return the signed cookie, which typically contains authentication token
      31 | };

      at find (test/userController.test.js:28:6)
      at Object.<anonymous> (test/userController.test.js:109:26)

2023-09-19 09:01:01 [ERROR]: Invalid Credentials - Error: Invalid Credentials
    at login (/Users/aasante/dev/practicum2023/backend/src/controllers/authController.js:69:11)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
2023-09-19 09:01:01 [INFO]: POST /api/v1/auth/login 401 72.801 ms - 29
Test Suites: 1 failed, 1 total
Tests:       1 failed, 9 skipped, 10 total
Snapshots:   0 total
Time:        2.364 s, estimated 4 s
```

The reason this happens, is that each test starts with a clean slate, an empty database. We need to set up ("Arrange") what we want the state of the database to be before our test. 

In this branch, you can see that for example, in the "Get user by id" test on lines 56-63, the test code first "Arranges" the state for the test by creating a test user with the matching email and password. That way, the credentials for logging in will work as expected.

If you pull in the latest changes from `main`, you can see that I've moved the `testUser` creation up into the `beforeAll` hook. This way, each test doesn't need to create that test user and can just go ahead and log in right away as the first "Arrange" step.

akosasante: Since we're always logging in with the testUserCredentials that were defined on lines 13-16, we can simplify this a bit:

```suggestion
    // Arrange: Log in and get a signed cookie with valid test user credentials
    const signedCookie = await loginAndReturnCookie(testUserCredentials);
```

test/userController.test.js
```diff
@@ -94,3 +94,143 @@ describe('GET /api/v1/users/:user_id endpoint', () => {
...
+});
+
+describe('GET /api/v1/users/getCurrentUserWithPurchasedAlbums endpoint', () => {
+  it('should return the data of the currently authenticated user with purchased albums', async () => {
+    // Act: Log in and get a signed cookie
+    const signedCookie = await loginAndReturnCookie(testUserCredentials);
+
+    // Act: Get the current user's information with purchased albums using the authenticated cookie
+    const response = await request(app)
+      .get('/api/v1/users/getCurrentUserWithPurchasedAlbums')
```
akosasante: When I tried to run this test (line 136), I got the following error:

```
 FAIL  test/userController.test.js
  ● GET /api/v1/users/getCurrentUserWithPurchasedAlbums endpoint › should return the data of the currently authenticated user with purchased albums

    expect(received).toBe(expected) // Object.is equality

    Expected: 200
    Received: 404

      146 |
      147 |     // Assert: Check the response status and user information
    > 148 |     expect(response.status).toBe(200);
          |                             ^
      149 |     expect(response.body).toHaveProperty('user');
      150 |     expect(response.body.user).toHaveProperty('name', 'ava');
      151 |     expect(response.body.user).toHaveProperty('email', 'ava@ava.com');

      at Object.toBe (test/userController.test.js:148:29)

2023-09-19 09:14:09 [INFO]: POST /api/v1/auth/login 201 65.700 ms - 101
2023-09-19 09:14:09 [ERROR]: Cast to ObjectId failed for value "getCurrentUserWithPurchasedAlbums" (type string) at path "_id" for model "User" - CastError: Cast to ObjectId failed for value "getCurrentUserWithPurchasedAlbums" (type string) at path "_id" for model "User"
    at ObjectId.Object.<anonymous>.ObjectId.cast (/Users/aasante/dev/practicum2023/backend/node_modules/mongoose/lib/schema/objectid.js:250:11)
    at ObjectId.Object.<anonymous>.SchemaType.applySetters (/Users/aasante/dev/practicum2023/backend/node_modules/mongoose/lib/schematype.js:1220:12)
    at ObjectId.Object.<anonymous>.SchemaType.castForQuery (/Users/aasante/dev/practicum2023/backend/node_modules/mongoose/lib/schematype.js:1632:15)
    at cast (/Users/aasante/dev/practicum2023/backend/node_modules/mongoose/lib/cast.js:356:32)
    at model.Query.Object.<anonymous>.Query.cast (/Users/aasante/dev/practicum2023/backend/node_modules/mongoose/lib/query.js:4911:12)
    at model.Query.Object.<anonymous>.Query._castConditions (/Users/aasante/dev/practicum2023/backend/node_modules/mongoose/lib/query.js:2232:10)
    at model.Query._findOne (/Users/aasante/dev/practicum2023/backend/node_modules/mongoose/lib/query.js:2519:8)
    at model.Query.exec (/Users/aasante/dev/practicum2023/backend/node_modules/mongoose/lib/query.js:4430:28)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at getSingleUser (/Users/aasante/dev/practicum2023/backend/src/controllers/userController.js:29:16)
2023-09-19 09:14:09 [INFO]: GET /api/v1/users/getCurrentUserWithPurchasedAlbums 404 16.441 ms - 67
```

We can see thta the test tried to make a `GET` request to the URL `/api/v1/users/getCurrentUserWithPurchasedAlbums`. And in the second last line in that error stacktrace, we can see that the error occured in the `getSingleUser` function in the `userController.js` file on line 29: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/PS145-user-test-read-endpoints/src/controllers/userController.js#L29](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/PS145-user-test-read-endpoints/src/controllers/userController.js#L29)

So my next question to myself would be, "Why did it call that controller function, instead of the `getCurrentUserWithPurchasedAlbums` handler?".

Let's take a look at the router file: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/PS145-user-test-read-endpoints/src/routes/userRoutes.js#L28C1-L30](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/PS145-user-test-read-endpoints/src/routes/userRoutes.js#L28C1-L30)

We see on line 29, that the correct path to get a user with their albums is actually: `/api/v1/showMe/withAlbums`. So what happened here is that express saw that the path didn't match any of the routes with strict string matching, so it assumed that `getCurrentUserWithPurchasedAlbums` must have been a user id and instead called the route on line 41. And the error occurs when Mongo complains that the string `getCurrentUserWithPurchasedAlbums` is not a valid ObjectId type.

akosasante:
```suggestion
      .get('/api/v1/users/showMe/withAlbums')
```

test/userController.test.js
```diff
@@ -94,3 +94,143 @@ describe('GET /api/v1/users/:user_id endpoint', () => {
...
+      // Add more checks for album properties if needed
+    });
+
+    expect(response.body.user).not.toHaveProperty('password');
+  });
+
+  it('should return a 401 status if the user is not authenticated', async () => {
+    // Act: Attempt to get the current user's information with purchased albums without authentication
+    const response = await request(app).get(
+      '/api/v1/users/getCurrentUserWithPurchasedAlbums'
```
akosasante:
```suggestion
      '/api/v1/users/showMe/withAlbums'
```

test/userController.test.js
akosasante: You can also verify this by checking the backend API docs and verifying the correct URL needed to be called in this case

test/userController.test.js
```diff
@@ -94,3 +94,143 @@ describe('GET /api/v1/users/:user_id endpoint', () => {
...
+});
+
+describe('GET /api/v1/users/showCurrentUser endpoint', () => {
+  it('should return the data of the currently authenticated user', async () => {
+    // Act: Log in and get a signed cookie
+    const signedCookie = await loginAndReturnCookie(testUserCredentials);
+
+    // Act: Get the current user's information using the authenticated cookie
+    const response = await request(app)
+      .get('/api/v1/users/showCurrentUser')
```
akosasante: This is the correct url for this endpoint:

```suggestion
      .get('/api/v1/users/showMe')
```

test/userController.test.js
```diff
@@ -94,3 +94,143 @@ describe('GET /api/v1/users/:user_id endpoint', () => {
...
+      name: 'ava',
+      email: 'ava@ava.com',
+      role: 'user',
+    });
+    expect(response.body.user).not.toHaveProperty('password');
+  });
+
+  it('should return a 401 status if the user is not authenticated', async () => {
+    // Act: Attempt to get the current user's information without authentication
+    const response = await request(app).get('/api/v1/users/showCurrentUser');
```
akosasante:
```suggestion
    const response = await request(app).get('/api/v1/users/showMe');
```

test/userController.test.js
```diff
@@ -94,3 +94,143 @@ describe('GET /api/v1/users/:user_id endpoint', () => {
...
+describe('GET /api/v1/users endpoint', () => {
+  it('should return a list of users without the password field', async () => {
+    // Act: Log in and get a signed cookie
+    const signedCookie = await loginAndReturnCookie(testUserCredentials);
+
+    // Act: Get a list of users using the authenticated cookie
+    const response = await request(app)
+      .get('/api/v1/users')
+      .set('Cookie', [signedCookie]);
+
```
akosasante, line 210: This endpoint only works for admin users. But the `testUserCredentials` are associated with a user, Ava Smith, with the role `'user'`.

We can follow similar steps as what was done to create `Ava` to create an admin user. If we want this admin user to be reusable throughout the tests, then the best place to create it is in the `beforeAll` callback/hook which runs one time when this file is loaded by Jest, before any tests have been run:

1.  In the top of the file, where other variables are declared, declare a new empty variable for `testAdmin`
2.  Declare a `const` variable for `testAdminCredentials` so that we can easily reuse the email and password for loggging in as an admin
3.  In `beforeAll` , use `await createSingleUser` to create a user with the `testAdminCredentials` email and password, select a name and username, and set the `role` to `'admin'`

Then in this test, on line 208, instead of logging in with the `testUserCredentials`, you would want to log in as an admin:

```
    // Arrange: Log in as an admin and get a signed cookie
    const signedCookie = await loginAndReturnCookie(testAdminCredentials);
```

test/userController.test.js
```diff
@@ -94,3 +94,143 @@ describe('GET /api/v1/users/:user_id endpoint', () => {
...
+    // Assert: Check the response status and the list of users
+    expect(response.status).toBe(200);
+    expect(response.body).toHaveProperty('users');
+    expect(Array.isArray(response.body.users)).toBe(true);
```
akosasante, line 214: :white\_check\_mark:  this looks good. Just sharing some other ways to do this same kind of test assertion:

`expect(response.body.users).toBeInstanceOf(Array);` (https://jestjs.io/docs/expect#tobeinstanceofclass)

`expect(response.body.users).toHaveLength(2); // There are 2 users in the database (ava and the admin)` (https://jestjs.io/docs/expect#tohavelengthnumber)

test/userController.test.js
```diff
@@ -94,3 +94,143 @@ describe('GET /api/v1/users/:user_id endpoint', () => {
...
+      expect(user).not.toHaveProperty('password');
+    });
+  });
+
+  it('should return a 401 status if the user is not authenticated', async () => {
+    // Act: Attempt to get a list of users without authentication
+    const response = await request(app).get('/api/v1/users');
+
+    expect(response.status).toBe(401);
+  });
```
akosasante, line 230: One additional test we can have is to verify that this endpoint only works for admin users and not regular user roles:

```suggestion
  });
  
  it('should return a 403 status if the user is not an admin', async () => {
    // Arrange: Log in as a regular user and get a signed cookie
    const signedCookie = await loginAndReturnCookie(testUserCredentials);

    // Act: Attempt to get a list of users using the authenticated cookie
    const response = await request(app).get('/api/v1/users').set('Cookie', [signedCookie]);

    // Assert: Check the response status for 403
    expect(response.status).toBe(403);
  });
```

test/userController.test.js
```diff
@@ -71,9 +71,28 @@ afterAll(async () => {
...
+    email: 'ava@ava.com',
+    password: 'secret',
+    role: 'user',
+  });
+});
+
+afterEach(async () => {
+  // Delete the test user after each test
+  await User.deleteMany({});
+});
```
akosasante: Same comment as in the other usercontroller test PR about simplifying this by having only one step for `deleteMany` and removing the redundant test user creation from the `beforeAll` block

akosasante (commented):
Will the admin user login for the delete endpoints be added to this PR or #61  ?

test/userController.test.js
mvstoyan: jest
1 failed, can't fix😒

test/userController.test.js
```diff
@@ -42,12 +44,19 @@ beforeAll(async () => {
...
+  //  username: 'ava123',
+  //  role: 'user',
+  //});
+  // Create the test admin user with the specified admin credentials
+  //testAdmin = await createSingleUser({
+  //  ...testAdminCredentials,
+  //  name: 'Admin',
+  //  username: 'admin123',
+  //  role: 'admin', // Set the role to 'admin'
+  //});
```
akosasante: We can delete this :+1:  Since we are creating and assigning the `testUser` variable before each test (`beforeEach`) rather than once at the beginning of the test suite here (`beforeAll`). 

The main reason we do that in this case is because we have some tests like the `updateCurrentUser` test where we modify the testUser. If we only created the `testUser` up here in the `beforeAll` then we'd need to adjust any test that happens after `updateCurrentUser` to expect that the `testUser.name` is now "new\_name" for example. But that's called having "brittle tests" where one test affects the input/output of a whole separate test. So to avoid that; we create a brand new testUser before each test.

For the admin user, we could create it here or in `beforeEach` either one would be fine since we don't have any tests that modify the `testAdmin` user. But to keep the test suite easy to follow and extend, it's probably best to just create all our users in the same place. It also makes it easier to handle our pre-test cleanup where we delete all users in the db so that we start with a fresh slate each time.

test/userController.test.js
```diff
@@ -58,9 +67,35 @@ afterAll(async () => {
...
+    email: 'admin@admin.com',
+    password: 'secret',
+    role: 'admin',
+  });
+});
+
+afterEach(async () => {
+  // Delete the test user after each test
+  await User.deleteMany({});
+});
```
akosasante: We can delete this :+1: 

Jest will run in the following order within a test block/describe block (see: [https://noriste.github.io/reactjsday-2019-testing-course/book/jest-101/jest-lifecycle.html)](https://noriste.github.io/reactjsday-2019-testing-course/book/jest-101/jest-lifecycle.html))

1.  beforeAll
2.  beforeEach
3.  the test itself
4.  afterEach
5.  repeat 2-3 until all tests in this block have completed
6.  afterAll 

So if we are already going to delete all the users in step (2) `beforeEach` it doesn't add anything to also delete them in step (4) `afterEach` . Since we have the rest of our logic in `beforeEach` we can simplify our code here by just removing `afterEach` altogether

test/userController.test.js
```diff
@@ -58,9 +67,35 @@ afterAll(async () => {
   await mongodb.stop();
 });
 
+beforeEach(async () => {
+  // Delete all users before creating the test user
```
akosasante:
```suggestion
  // Delete all users before creating the test users (one for each role)
```

test/userController.test.js
```diff
@@ -93,3 +127,240 @@ describe('GET /api/v1/users/:user_id endpoint', () => {
...
+    expect(response.status).toBe(200);
+    expect(response.body).toHaveProperty('user');
+    expect(response.body.user).toMatchObject(updatedUserData);
+  });
+
+  it('should return an error if required data is missing', async () => {
+    const signedCookie = await loginAndReturnCookie({
+      email: 'ava@ava.com',
+      password: 'secret',
+    });
```
akosasante: not a big deal; but to keep things consistent let's use the variable we've already declared with this data:

```suggestion
    const signedCookie = await loginAndReturnCookie(testUserCredentials);
```

test/userController.test.js
```diff
@@ -93,3 +127,240 @@ describe('GET /api/v1/users/:user_id endpoint', () => {
...
+    // Assert: Check the response status for an error
+    expect(response.status).toBe(404);
+  });
+
+  it('should return an error if not authenticated', async () => {
+    // Act: Attempt to delete a user without authentication
+    const response = await request(app).delete(
+      '/api/v1/users/deleteSingleUser'
+    );
+
```
akosasante: Ah I think we're mixing up the controller handler name `deleteSingleUser` with the actual URL/path that is defined in `userRoutes.js`. We can also verify by double checking what the correct path is in the generated API docs.

akosasante (commented):
Looking good; left some suggestions for cleaning things up a bit and a pointer for the failing DELETE test. After that should be good :+1:

test/userController.test.js
mvstoyan: Thank you, Akos! Very good information for me✍️

test/userController.test.js
mvstoyan: Yes, I wasn't very attentive

mvstoyan:
@akosasante All tests passed, thank you very much for all the explanations and your endless patience❤️

src/middleware/error-handler.js
akosasante, line 1: I think we can revert the change in this file :)

akosasante (approved):
👍 🎉

src/middleware/error-handler.js
mvstoyan: jest @akosasante If I go back to previous code, I return the previous error.

akosasante: That's expected, the logs we see here are exactly what our Express Server is logging/would log if these requests were really being made. In this case it is  logging the `error` from the error handler when there is a 404 or 401 that we are testing for. But as long as all the tests are passing then that means we are handling or asserting on that error case, then there is no problem. It is still useful to be able to see all the INFO and ERROR logs even when testing, to understand what steps the server has done. 

So in the logs shown in your screenshot I can follow along and see that:

*   You're running only the three tests in the `DELETE /api/v1/users/deleteSingleUser endpoint` describe block
*   Test1:
    *   INFO 201 log from `morgan` library for the request when the `should delete a user if the user exists` test logs in  :white\_check\_mark: 
    *   INFO 200 log from `morgan` library for the request when the `should delete a user if the user exists` test successfully deletes a user  :white\_check\_mark:
*   Test2:
    *   INFO 201 log from `morgan` library for the request when the `should return an error if the user does not exist` test logs in :white\_check\_mark: 
    *   ERROR log printed by our `logger.error` call in the error handler middleware (using the `winston` library) when `should return an error if the user does not exist` test tries to delete a non existent user id. This is exactly what we're testing for and we assert that with (`expect(response.status).toBe(404);`) and the test passes :white\_check\_mark: 
    *   INFO 404 log from from `morgan` library for the response returned when server fails to delete a non-existing user
*   Test3:
    *   ERROR log printed by our `logger.error` call in the error handler middleware (using the `winston` library) when `should return an error if not authenticated` test tries to login with invalid credentials. This is exactly what we're testing for and we assert that with (`expect(response.status).toBe(401);`) and the test passes :white\_check\_mark:
    *   INFO 401 log from from `morgan` library for the response returned when server fails to login

akosasante: We do this because on the other hand, if we completely don't log errors at all in the `test` environment it can make it a bit more difficult to debug things when we can't easily see the cases where the error handler is invoked.  The code in `main` just makes things a bit more streamlined in `test` env by not printing the full error stack, just the message itself

akosasante (approved):

src/middleware/error-handler.js
mvstoyan: Thank you very much for such a clear explanation.

PR #64, opened 2023-09-16
PS140-test-auth-controller

mvstoyan:


mvstoyan:
I have one error Cast to ObjectId failed for value "nonexistentUserId"

akosasante:
> 
> 
>  I have one error Cast to ObjectId failed for value "nonexistentUserId"

That's expected. Since for that in the userController.test.js file, we are intentionally passing a bad id (`"nonexistentUserId"` ) to the "GET user by id" endpoint. The test itself is expecting to get a `404` and thus passes (we can see that we have 7 passed tests and no failed tests in our test summary). 

If it makes it a bit less confusing, we can definitely turn off logging of error stacks in test mode, by adjusting the line here in the `error-handler` middlware: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/middleware/error-handler.js#L5](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/middleware/error-handler.js#L5) and making it instead only do that in other environments:

`if (process.env.NODE_ENV !== "test") logger.error(err);`

test/authController.test.js
```diff
@@ -0,0 +1,108 @@
...
+describe('Authentication API Endpoints', () => {
+  it('should register a new user and log in', async () => {
+    // Arrange: Register a new user
+    const registrationResponse = await request(app)
+      .post('/api/v1/auth/register')
+      .send(testUser);
+
+    // Assert: Check the response status and body
+    expect(registrationResponse.status).toBe(201); // Expecting a successful registration
+    expect(registrationResponse.body).toHaveProperty('user'); // Expecting a user object in the response
```
akosasante, line 71: This is great; we can also make further assertions to (1) verify that the registered user was actually stored in the db as expected, and (2) that the shape of the response from the `register` endpoint is what we expect

```suggestion
    // Assert: Check the response status and body
    expect(registrationResponse.status).toBe(201); // Expecting a successful registration
    expect(registrationResponse.body).toHaveProperty('user'); // Expecting a user object in the response
    const createdUser = await User.findOne({ email: testUser.email });
    // Expecting the user to be saved in the database (not null)
    expect(createdUser).not.toBeNull();
    // Expect the user object in the response to match the user object in the database
    expect(registrationResponse.body.user).toMatchObject({
      email: createdUser.email,
      name: createdUser.name,
      role: createdUser.role,
      userId: createdUser.id,
    });
```

test/authController.test.js
```diff
@@ -0,0 +1,108 @@
...
+
+    // Act: Log in with the newly registered user's credentials
+    const loginResponse = await request(app)
+      .post('/api/v1/auth/login')
+      .send(testUserCredentials);
+
+    // Assert: Check the response status and body
+    expect(loginResponse.status).toBe(201); // Expecting a successful login
+    expect(loginResponse.body).toHaveProperty('user'); // Expecting a user object in the response
```
akosasante, line 90: Here we could also have further assertions on the shape of the object inside `loginResponse.body.user`

akosasante (approved):
Some small suggestions, but overall looks ready to merge :+1:

test/authController.test.js
mvstoyan: jest

test/authController.test.js
mvstoyan: jest1
With each unit test, it becomes clearer and more familiar. Thank you very much!❤️

PR #65, opened 2023-09-17
utils JSDoc

mvstoyan:
As I understand it, because our main documentation is based on Swagger, and we use JSDoc only for utils, we don't need to install the dependency "npm install -g jsdoc."?

### DIFF
--- a/src/utils/createTokenUser.js
+++ b/src/utils/createTokenUser.js
@@ -1,4 +1,11 @@
 /**
+ * @typedef {Object} TokenUser - User object with data for token creation.
+ * @property {string} name - User's name.
+ * @property {string} userId - User's ObjectId from the Mongo database.
+ * @property {string} role - User's role. Can be either 'admin' or 'user'.
+ * @property {string} email - User's email.
+ */
+/**
  * Creates a user object based on the data provided by the user.
  *
  * @param {Object} user - User object with data.
@@ -6,7 +13,7 @@
  * @param {string} user._id - User's identifier.
  * @param {string} user.role - User's role.
  * @param {string} user.email - User's email.
- * @returns {Object} - User object with data for token creation.
+ * @returns {TokenUser} - User object with data for token creation.
  */
 
 const createTokenUser = (user) => {
--- a/src/utils/createTokenUser.js
+++ b/src/utils/createTokenUser.js
@@ -1,3 +1,14 @@
+/**
+ * Creates a user object based on the data provided by the user.
+ *
+ * @param {Object} user - User object with data.
+ * @param {string} user.name - User's name.
+ * @param {string} user._id - User's identifier.
+ * @param {string} user.role - User's role.
+ * @param {string} user.email - User's email.
+ * @returns {Object} - User object with data for token creation.
+ */
+
 const createTokenUser = (user) => {
   // Create a token-based user object using the provided user data
   return {
--- a/src/utils/jwt.js
+++ b/src/utils/jwt.js
@@ -17,11 +17,15 @@ const creatJWT = ({ payload }) => {
 };
 
 /**
+ * @typedef {import('./createTokenUser.js').TokenUser} TokenUser
+ */
+/**
  * Checks the validity of a JWT token.
  *
  * @param {Object} options - Options for verifying a JWT token.
  * @param {string} options.token - The JWT token to verify.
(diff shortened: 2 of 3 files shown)


src/utils/createTokenUser.js
```diff
@@ -1,3 +1,14 @@
...
+ * Creates a user object based on the data provided by the user.
+ *
+ * @param {Object} user - User object with data.
+ * @param {string} user.name - User's name.
+ * @param {string} user._id - User's identifier.
+ * @param {string} user.role - User's role.
+ * @param {string} user.email - User's email.
+ * @returns {Object} - User object with data for token creation.
+ */
+
```
akosasante: Looks great! We can further define what the shape of the return value is by using the `typedef` tag: [https://jsdoc.app/tags-typedef.html](https://jsdoc.app/tags-typedef.html)

```suggestion
/**
 * @typedef {Object} TokenUser - User object with data for token creation.
 * @property {string} name - User's name.
 * @property {string} userId - User's ObjectId from the Mongo database.
 * @property {string} role - User's role. Can be either 'admin' or 'user'.
 * @property {string} email - User's email.
 */

/**
 * Creates a user object based on the data provided by the user.
 *
 * @param {Object} user - User object with data.
 * @param {string} user.name - User's name.
 * @param {string} user._id - User's identifier.
 * @param {string} user.role - User's role.
 * @param {string} user.email - User's email.
 * @returns {TokenUser} - User object with data for token creation.
 */

```

src/utils/jwt.js
```diff
@@ -8,8 +16,24 @@ const creatJWT = ({ payload }) => {
   return token;
 };
 
+/**
+ * Checks the validity of a JWT token.
+ *
+ * @param {Object} options - Options for verifying a JWT token.
+ * @param {string} options.token - The JWT token to verify.
+ * @throws {Error} - Thrown if the token is invalid.
+ */
```
akosasante: This function also returns the token object if the JWT verification passes. We can reused the type definition declared in the `checkTokenUser` file:

```suggestion
/**
 * @typedef {import('./createTokenUser.js').TokenUser} TokenUser
 */

/**
 * Checks the validity of a JWT token.
 *
 * @param {Object} options - Options for verifying a JWT token.
 * @param {string} options.token - The JWT token to verify.
 * @throws {Error} - Thrown if the token is invalid.
 * @returns {TokenUser} - User object with data for token creation.
 */
```

akosasante (approved):
Left some small suggestions, but overall looks good to me :+1: 

As for your question:

> As I understand it, because our main documentation is based on Swagger, and we use JSDoc only for utils, we don't need to install the dependency "npm install -g jsdoc."?

Yes writing JSDoc comments does not require any additional depedencies in node. And our external API documentation is specifically using Swagger and autogenerated from comments with the `swagger-autogen` library. 

Having JSDoc comments in the file is helpful for a newcomer to the repo and also helps your IDE (eg: VSCode) to generate hints about what fields are available when using these functions elsewhere in the code.

Now if we want to also publicly host these documentations in a site similar to how we do the Swagger API docs, then yes in that case, we would want to use a library like [jsdoc,](https://www.npmjs.com/package/jsdoc) which will read these comments and generate HTML files from them. The npm project page for jsdoc also suggests various templating libraries like [BetterDocs](https://github.com/SoftwareBrothers/better-docs) to use with it to change the look and feel of the documentation. Then you can use `express.static` to serve the output folder of jsdoc where it generates all the html.

**Resources**
- https://www.youtube.com/watch?v=UBMYogTzsBk
- https://www.youtube.com/watch?v=YK-GurROGIg
- https://www.youtube.com/watch?v=Nqv6UkTROak
- https://www.youtube.com/watch?v=lTCadytiCNs

src/utils/jwt.js
mvstoyan: @akosasante We have unit tests. Is it a good practice to create documentation for them using JSDoc and integrate it into a separate HTML document for the team? For example, we currently have documentation in manual HTML format for socket tests and userById. Does it make sense to rewrite all of this using JSDoc?

src/utils/createTokenUser.js
mvstoyan: jsDoc

src/utils/jwt.js
mvstoyan: jsDoc

PR #66, opened 2023-09-18
Album test read endpoints

DYA13:
Akos could you please check ? I have Validation Error :Invalid Spotify URL. I am stuck a bit with it. Thank you


test/albumController.test.js
```diff
@@ -0,0 +1,105 @@
...
+const request = require('supertest');
+const { MongoMemoryServer } = require('mongodb-memory-server');
+const Album = require('../src/models/Album');
+
+const {
+  getAllAlbums,
+  getSingleAlbum,
+  getAlbumWithAllUsersWhoPurchasedIt,
+  getFilteredAlbums,
+} = require('../src/controllers/albumController');
```
akosasante: We are not writing unit tests that call these controller files directly so we do not need to import them here

test/albumController.test.js
```diff
@@ -0,0 +1,105 @@
...
+  await mongooseConnection.disconnect();
+  await mongodb.stop();
+});
+
+describe('AlbumController API Tests', () => {
+  const mockAlbumData = {
+    albumName: 'Test Album',
+    artistName: 'Test Artist',
+    price: 9.99,
+    spotifyUrl: 'https://api.spotify.com/v1/albums/',
```
akosasante: "Invalid Spotify URL" message comes from the mongoose validator for the album schema in the Album.js file on line 45 of that file: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/1e48ca722a1a852752cda040cae126beeffd535d/src/models/Album.js#L36-L47](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/1e48ca722a1a852752cda040cae126beeffd535d/src/models/Album.js#L36-L47)

The validator for the `spotifyUrl` field expects the following regex: `/^https:`**`\/\/`**`api\.spotify\.com`**`\/`**`v1`**`\/`**`albums`**`\/`**`[a`**`-`**`zA`**`-`**`Z0`**`-`**`9]+$/`

That regex means it's expecting a string with the shape: [`https://api.spotify.com/v1/albums/somelettersandornumbers`](https://api.spotify.com/v1/albums/somelettersandornumbers) but you are not passing in anything after the `/albums/` part of the URL so it is failing that check.

test/albumController.test.js
```diff
@@ -0,0 +1,105 @@
...
+    const response = await request(app).get('/api/v1/albums');
+    expect(response.status).toBe(StatusCodes.OK);
+    expect(response.body).toHaveProperty('albums');
+    expect(response.body.albums).toHaveLength(1);
+  });
+
+  it('should test the getAllAlbums endpoint - Error Case (Unauthorized)', async () => {
+    const response = await request(app).get('/api/v1/albums');
+    expect(response.status).toBe(StatusCodes.UNAUTHORIZED);
+  });
```
akosasante: This fails because you've written the test to expect a 401, but the endpoint is returning a 200:

```
  ● AlbumController API Tests › should test the getAllAlbums endpoint - Error Case (Unauthorized)

    expect(received).toBe(expected) // Object.is equality

    Expected: 401
    Received: 200
```

If we check our `albumRoutes` we see this is expected because we don't have any auth middlware on the `/api/v1/albums` route. Anyone can fetch the list of albums available, without logging in. This was done so that users can browse the store as a guest without being forced to log in first. 

[https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/8f5ec32895b103423ab561869b2479b76223af5d/src/routes/albumRoutes.js#L21](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/8f5ec32895b103423ab561869b2479b76223af5d/src/routes/albumRoutes.js#L21)

  
So we can just remove this particular test case. If you'd like instead, you could have a test case for verifying that the endpoint returns an empty list if there are no albums in the database (you'd first have to do `await Album.deleteMany({});` within the test to delete any albums currently in the db, and then make the GET call)

test/albumController.test.js
```diff
@@ -0,0 +1,105 @@
...
+
+  it('should test the getSingleAlbum endpoint - Success Case', async () => {
+    const albums = await Album.find({});
+    const albumId = albums[0]._id;
```
akosasante: Could shorten this a bit with the `findOne` mongoose function that returns the first matching album from the db, and some object destructuring:

```suggestion
    const { _id: albumId } = await Album.findOne({});
```

test/albumController.test.js
```diff
@@ -0,0 +1,105 @@
...
+
+    const response = await request(app).get(`/api/albums/v1/${albumId}`);
```
akosasante: This test is currently receiving a `404` instead of the expected `200` status code. That is because Express is returning a "Route not found" error because this route is invalid:

```

    Expected: 200
    Received: 404
```

The URL should not be `/api/albums/v1` it should be `/api/v1/albums/${albumId}`

test/albumController.test.js
```diff
@@ -0,0 +1,105 @@
...
+    expect(response.status).toBe(StatusCodes.OK);
+    expect(response.body).toHaveProperty('album');
+  });
+
+  it('should test the getSingleAlbum endpoint - Error Case (Not Found)', async () => {
+    const response = await request(app).get('/api/albums/v1/invalidAlbumId');
```
akosasante:
```suggestion
    const response = await request(app).get('/api/v1/albums/invalidAlbumId');
```

test/albumController.test.js
```diff
@@ -0,0 +1,105 @@
...
+    expect(response.status).toBe(StatusCodes.NOT_FOUND);
+  });
+
+  it('should test the getAlbumWithAllUsersWhoPurchasedIt endpoint - Success Case', async () => {
+    const albums = await Album.find({});
+    const albumId = albums[0]._id;
```
akosasante: suggestion for one-liner

```suggestion
    const { _id: albumId } = await Album.findOne({});
```

test/albumController.test.js
```diff
@@ -0,0 +1,105 @@
...
+
+    const response = await request(app).get(
+      `/api/albums/${albumId}/purchased-users`
```
akosasante: There is no endpoint with this URL in the app.

In our albumRoutes.js file, we see that the controller we're trying to test here is mounted on the URL: `/api/v1/albums/:id/listOfUsersWhoPurchasedThisAlbum`: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/8f5ec32895b103423ab561869b2479b76223af5d/src/routes/albumRoutes.js#L31-L37](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/8f5ec32895b103423ab561869b2479b76223af5d/src/routes/albumRoutes.js#L31-L37)

So that's what we need to use here in the test (or if we want to update that URL we need to (1) make sure anywhere the frontend is using that endpoint is updated, and (2) update the albumRoutes.js file, and (3) regenerate the swagger docs)

akosasante:
```suggestion
      `/api/albums/${albumId}/listOfUsersWhoPurchasedThisAlbum`
```

test/albumController.test.js
```diff
@@ -0,0 +1,105 @@
...
+    );
+    expect(response.status).toBe(StatusCodes.OK);
+    expect(response.body).toHaveProperty('album');
```
akosasante, line 106: Since this test is specifically about making sure this endpoint returns the album _**with**_ all the users who purchased it; it's probably worthwhile to also assert that in our expectations here:

```suggestion
    expect(response.body).toHaveProperty('album');
    expect(response.body).toHaveProperty('album');
```

test/albumController.test.js
```diff
@@ -0,0 +1,105 @@
...
+    expect(response.body).toHaveProperty('purchasingUsersCount');
+  });
+
+  it('should test the getAlbumWithAllUsersWhoPurchasedIt endpoint - Error Case (Not Found)', async () => {
+    const response = await request(app).get(
+      '/api/albums/invalidAlbumId/purchased-users'
```
akosasante:
```suggestion
      '/api/v1/albums/invalidAlbumId/listOfUsersWhoPurchasedThisAlbum'
```

^ correct URL (with invalid album id)

test/albumController.test.js
```diff
@@ -0,0 +1,105 @@
...
+    expect(response.status).toBe(StatusCodes.OK);
+    expect(response.body).toHaveProperty('album');
+  });
+
+  it('should test the getSingleAlbum endpoint - Error Case (Not Found)', async () => {
+    const response = await request(app).get('/api/albums/v1/invalidAlbumId');
+    expect(response.status).toBe(StatusCodes.NOT_FOUND);
+  });
+
+  it('should test the getAlbumWithAllUsersWhoPurchasedIt endpoint - Success Case', async () => {
```
akosasante, line 87: For `getAlbumWithAllUsersWhoPurchasedIt` endpoint, keep in mind that based on our `albumRoutes.js` file, this endpoint is only accessible by an admin user.

So we need to first log into the site, with admin user credentials before being able to get a successful response. You can follow a similar pattern as what's used in the `userController.test.js` file for logging into the server: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/album-test-read-endpoints/test/userController.test.js#L19-L31](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/album-test-read-endpoints/test/userController.test.js#L19-L31)

It might even be a good idea to extract that function into a common `test_helper.js` file so that multiple test files can import it instead of having to duplicate it in each one.

test/albumController.test.js
```diff
@@ -0,0 +1,105 @@
...
+    expect(response.status).toBe(StatusCodes.NOT_FOUND); // Expect not found status code
+  });
+
+  it('should test the getFilteredAlbums endpoint - Success Case', async () => {
+    const response = await request(app).get(
+      '/api/v1/albums?limit=10&order=asc'
+    );
+    expect(response.status).toBe(StatusCodes.OK);
+    expect(response.body).toHaveProperty('albums');
+    expect(response.body).toHaveProperty('count');
```
akosasante, line 127: We can also assert that the `count` is the number that we expect (in this case, only 1 album has been created).

Another useful test would be to create, let's say 5 albums, and then use a limit of `2` and verify that the endpoint acutally only returns `2` albums in the response.

test/albumController.test.js
```diff
@@ -0,0 +1,105 @@
...
+  });
+
+  it('should test the getFilteredAlbums endpoint - Error Case (Bad Request)', async () => {
+    const response = await request(app).get(
+      '/api/v1/albums?invalidParam=invalidValue'
+    );
+    expect(response.status).toBe(StatusCodes.BAD_REQUEST); // Expect bad request status code
+  });
```
akosasante: I don't see anything in the Swagger docs about a Bad Request status being returned when the parameters have invalid params? And the code doesn't indicate that either: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/ed2a4e7e04bf48fce393395a8e00a4f3975c6549/src/controllers/albumController.js#L162-L226](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/ed2a4e7e04bf48fce393395a8e00a4f3975c6549/src/controllers/albumController.js#L162-L226)

It seems that any invalid params would just be ignored. So thi stest case can probably be removed.

test/albumController.test.js
DYA13: deleted commit f2001037da4024490c12c8797f4d8ac72e764391

test/albumController.test.js
DYA13: Ok, should I pass 'id' in some way or write this URL in another way?

test/albumController.test.js
DYA13: I have tried to do as you explained , I have changed it for verifying if the endpoints returns  an empty list, could you please have a look? commit 23b4f0ecdf3ecb19bdc226ec04792096dfec04e0

test/albumController.test.js
DYA13: Thank you

test/albumController.test.js
DYA13: Sorry Akos , I have forgotten v1 , edited, commit 345aa941589e6a181179bb51d2233adb3398dd91

test/albumController.test.js
DYA13: Again my apologies

test/albumController.test.js
DYA13: removed,  commit 54b583a65fb7fe1f4ff0948857e2904890c9cca3

test/albumController.test.js
DYA13: There is no actually test  in ticket for it , may be I can delete it?

test/albumController.test.js
akosasante: Since we're just creating a fake album for these tests, just putting any random characters after `/albums` should work. So something like:  
 

```suggestion
    spotifyUrl: 'https://api.spotify.com/v1/albums/blah123',
```

is fine

test/albumController.test.js
akosasante: The `beforeEach` step creates an album though right; so we need to delete that album in the test:

You have:

```
  it('should return an empty list if there are no albums in the database', async () => {
    const response = await request(app).get('/api/v1/albums');
    expect(response.status).toBe(StatusCodes.UNAUTHORIZED);
    expect(response.body).toEqual([]);
  });
```

But (1) this won't be an "Unauthorized" status code, it'll be a 200 because we don't check for logged in users on this endpoint. And (2) we need to first delete the albums before we can do the test so:

```
  it('should return an empty list if there are no albums in the database', async () => {
    await Album.deleteMany({});
    const response = await request(app).get('/api/v1/albums');
    expect(response.status).toBe(StatusCodes.OK);
    expect(response.body).toEqual([]);
  });
```

test/albumController.test.js
akosasante: You have `v1` in there two times now in that commit: ``const response = await request(app).get(`/api/v1/albums/v1/${albumId}`);`` but it should only be there between "api" and "albums"

test/albumController.test.js
akosasante: The ticket asks to create a test for all the `read` endpoints (so any GET endpoints); which would include this one. And it does call out the endpoint that is handled by the controller function `getAlbumWithAllUsersWhoPurchasedIt`. So I don't think this needs to be deleted; it's just that the URL that you're hitting `/api/albums/${albumId}/purchased-users`, is not a valid URL. We should make sure that we compare to `AlbumRoutes.js` and to the Swagger docs to be using the correct API URL.

test/albumController.test.js
```diff
@@ -0,0 +1,92 @@
...
+  it('should test the getSingleAlbum endpoint - Error Case (Not Found)', async () => {
+    const response = await request(app).get('/api/v1/albums/invalidAlbumId');
+    expect(response.status).toBe(StatusCodes.NOT_FOUND);
+  });
+
+  it('should test the getAlbumWithAllUsersWhoPurchasedIt endpoint - Success Case', async () => {
+    const { _id: albumId } = await Album.findOne({});
+
+    const response = await request(app).get(
+      `/api/albums/${albumId}/listOfUsersWhoPurchasedThisAlbum`
```
akosasante: missing the v1

test/albumController.test.js
DYA13: Great , I didn't know it

test/albumController.test.js
DYA13: Thank you , ok I understood, commit 6ebe135cf632e360db1ddde5a71323b08b647acf

test/albumController.test.js
DYA13: I am sorry 😅, edited , commit 7039a0ccec34b99975bfb5ccf0ad1f285f977e56

test/albumController.test.js
DYA13: I have edited the API URL , commit  8ac2243106768be8c277e8efdd032cab37c27fb4
I just wanted to ask regarding read / write endpoints, read endpoints are all GET requests and write endpoints are PUT , POST request , is it right ?

test/albumController.test.js
```diff
@@ -0,0 +1,92 @@
...
+    expect(response.status).toBe(StatusCodes.OK);
+    expect(response.body).toHaveProperty('albums');
+    expect(response.body.albums).toHaveLength(1);
+  });
+
+  it('should return an empty list if there are no albums in the database', async () => {
+    await Album.deleteMany({});
+    const response = await request(app).get('/api/v1/albums');
+    expect(response.status).toBe(StatusCodes.OK);
+    expect(response.body).toEqual([]);
```
akosasante: The error message given by the test shows that the `body` of the response isn't directly an empty array; verify this against the swagger docs/controller code as well to see what we should be asserting here:

```
  ● AlbumController API Tests › should return an empty list if there are no albums in the database

    expect(received).toEqual(expected) // deep equality

    Expected: []
    Received: {"albums": [], "count": 0}

      47 |     const response = await request(app).get('/api/v1/albums');
      48 |     expect(response.status).toBe(StatusCodes.OK);
    > 49 |     expect(response.body).toEqual([]);
         |                           ^
      50 |   });
      51 |
      52 |   it('should test the getSingleAlbum endpoint - Success Case', async () => {

      at Object.toEqual (test/albumController.test.js:49:27)
```

test/albumController.test.js
DYA13: I have set test_helper.js as separate file ,  commit 513150c969881a957daf9b7aea8e73eb6965622f. and commit 6572af83c71ec991f18ecd30765c89b1d6690d3f

test/albumController.test.js
DYA13: added v1 
 `  const response = await request(app)
      .get(`/api/v1/albums/${albumId}/listOfUsersWhoPurchasedThisAlbum`)
      .set('Cookie', loginResponse.header['set-cookie']);`

test/albumController.test.js
DYA13: I tried to edit it in this way commit d565fd8fd84ddca98a0a6da9e84803aebe2cbf1e

test/albumController.test.js
DYA13: Regarding the test for count , I tried to make it in this way commit f195f4c3fcda6a4769d1729b012029784fffa5a8

DYA13: As for limitation of albums , I have tried to implement in this way commit 954f9f5cfb2dcb769562890f41d8a1a95108857c

DYA13:
It seems , it can't find cookies in file test_helper.js , may be I forgot to import some files 🤔
image
Actually here returns as well  401 but should 404 
image

test/test_helper.js
```diff
@@ -0,0 +1,21 @@
+const request = require('supertest');
+const { app } = require('../src/expressServer.js');
+
+const loginAsAdmin = async (credentials) => {
```
akosasante: We should be able to use this to log in as a `user` role user and as an `admin` user. I don't think there's anything in here that is specific to logging in as an `admin`. It just depends on what credentials we pass into the function with. 

```suggestion
const loginAndReturnCookie = async (credentials) => {
```

test/test_helper.js
```diff
@@ -0,0 +1,21 @@
...
+  // Find the cookie that starts with 'token' in the 'set-cookie' header
+  // Split the cookie string by ';' and take the first part (before the first ';')
+  const signedCookie = cookieHeader
+    .find((cookie) => cookie.startsWith('token'))
+    .split(';')[0];
+  return signedCookie; // Return the signed cookie, which typically contains authentication token
+};
+
+module.exports = {
+  loginAsAdmin,
```
akosasante:
```suggestion
  loginAndReturnCookie,
```

test/albumController.test.js
```diff
@@ -0,0 +1,187 @@
...
+  it('should test the getSingleAlbum endpoint - Error Case (Not Found)', async () => {
+    const response = await request(app).get('/api/v1/albums/invalidAlbumId');
+    expect(response.status).toBe(StatusCodes.NOT_FOUND);
+  });
+
+  it('should test the getAlbumWithAllUsersWhoPurchasedIt endpoint - Success Case', async () => {
+    const { _id: albumId } = await Album.findOne({});
+
+    // Login as admin
+    const loginResponse = await loginAsAdmin(app);
```
akosasante: Your `loginAsAdmin` function takes `credentials` as an argument. Taking a look at how that function is used in `userController.test.js` we can see that we are passing an object `{ email: <email to log in with>, password: <password to log in with> }`: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/album-test-read-endpoints/test/userController.test.js#L65

But here you've passed in the Express `app` instead. So that is why it is failing to log in and failing to find a cookie in the response from the login request.

If you want to log in you need to:

1.  Create a user (with mongoose's `User.create` function): [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/album-test-read-endpoints/test/userController.test.js#L56-L62](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/album-test-read-endpoints/test/userController.test.js#L56-L62)
2.  Use the email and password of that user to login by calling the test helper login function

If you want to log in as an admin user, follow the same steps as above but create the user with the role `"admin"`

akosasante:
> It seems , it can't find cookies in file test\_helper.js , may be I forgot to import some files 🤔 
> 
> image

^ that means the login request did not return a Cookie. So the login request failed (see my comments in the review for why that happened)

>  Actually here returns as well 401 but should 404 
> 
> image

^ 401 is `UNAUTHORIZED` which we return from the login endpoint if invalid credentials were used. 

If you run the test with the latest changes from `main` you will see the following logged which shows that the login request is the one that failed:  
 

```
2023-09-20 10:59:21 [ERROR]: Please provide email and password
2023-09-20 10:59:21 [INFO]: POST /api/v1/auth/login 400 63.425 ms - 43
```

When running Jest you can choose to run only a specific test which can be helpful if you don't want to get distracted by logs from all the successful tests: `npm run test -- --testPathPattern=test/albumController.test.js --testNamePattern="getAlbumWithAllUsersWhoPurchasedIt"`

test/albumController.test.js
DYA13: I have tried to change it but still have  errors lines 93 and 184 with limit of albums 🥹 ,commit f4d6bbbbbef4da708efad289e00c456c8968988c

DYA13: it can't find cookie 
image

test/test_helper.js
akosasante, line 1: We can also help simplify other existing test files (`userController.test.js` and `authController.test.js` from `main` ) by updating them to also use this test\_helper exported `loginAndReturnCookie` function instead of having it repeated across files.

test/albumController.test.js
```diff
@@ -0,0 +1,190 @@
...
+  });
+
+  it('should test the getAlbumWithAllUsersWhoPurchasedIt endpoint - Success Case', async () => {
+    // Try to find an album; if not found, create a default one
+    const album =
+      (await Album.findOne({})) ||
+      (await Album.create({
+        albumName: 'Album',
+        artistName: 'Artist',
+      }));
```
akosasante: We know that at the beginning of each test, we've created an album (in the `beforeEach` Jest callback: [https://noriste.github.io/reactjsday-2019-testing-course/book/jest-101/jest-lifecycle.html)](https://noriste.github.io/reactjsday-2019-testing-course/book/jest-101/jest-lifecycle.html)) so I don't think we need ot have this conditional here. Did you encounter a case where the created album wasn't available here in this test?

```suggestion
    // Try to find an album; if not found, create a default one
    const album = await Album.findOne({});
```

test/albumController.test.js
```diff
@@ -0,0 +1,190 @@
...
+
+    const adminCredentials = {
+      email: 'admin@admin.com',
+      password: 'adminpassword',
+    };
+
+    // Use the loginAndReturnCookie function to log in as admin and get the signed cookie
+    const signedCookie = await loginAndReturnCookie(adminCredentials);
```
akosasante, line 97: At this point, your fake Mongo database in this test has 0 users. So we try to login with these admin credentials and we can see in the test logs that it fails:

```
2023-09-22 12:51:20 [ERROR]: Invalid Credentials
2023-09-22 12:51:20 [INFO]: POST /api/v1/auth/login 401 28.167 ms - 29
```

  
First we need to create an admin user in the database, so that we can log in as that user. You can see that we do that in the userController by having a \`let\` variable for \`testUser\` and then creating and assigning a user to that variable to be used throughout the tests: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/test/userController.test.js#L45-L50

For now, a good place to put that would be in a `beforeAll` hook (we can write it anywhere in this `describe` block or right before the `describe` block. Either way, Jest will run it one time, before running any tests; so it's an appropriate place to do any setup arrangements that we don't expect to change throughout the tests). Let's put it right before the `beforeEach` lifecycle block as a best practice (line 33):

```javascript
  beforeAll(async () => {
    await User.create({
      email: 'admin@admin.com',
      password: 'adminpassword',
      name: 'Ava Smith',
      username: 'ava123',
      role: 'admin'
    })
  });
```

test/albumController.test.js
akosasante: Oops in my previous comment I didn't include what I meant by that 😮‍💨 ; this endpoint returns an `album` object with an additional field that shows all the users who purchased it. We should add at least one `expect` assertion here to verify that the response body we get actually has that field; this will help us catch any bugs or errors in the future if the shape of this response is accidentally changed in the controller:

```json
    {
      album: {
        _id: '650dc79709eee6395154e250',
        artistName: 'Test Artist',
        albumName: 'Test Album',
        price: 9.99,
        image: '/uploads/example.jpg',
        spotifyUrl: 'https://api.spotify.com/v1/albums/blah123',
        averageRating: 0,
        numOfReviews: 0,
        createdAt: '2023-09-22T16:57:59.195Z',
        updatedAt: '2023-09-22T16:57:59.195Z',
        __v: 0,
        purchasedByUsers: [],
        id: '650dc79709eee6395154e250'
      },
      purchasingUsersCount: 0
    }
```

test/test_helper.js
```diff
@@ -0,0 +1,23 @@
+const request = require('supertest');
+const { app } = require('../src/expressServer.js');
+
+const loginAndReturnCookie = async (credentials) => {
+  // Send a login request and get the signed cookie from the response
+  const resp = await request(app).post('/api/v1/auth/login').send(credentials);
+  // Log the response and cookie header for debugging
+  console.log('Response:', resp);
+  console.log('Cookie Header:', resp.headers['set-cookie']);
```
akosasante: We can remove these :+1:

test/albumController.test.js
```diff
@@ -0,0 +1,190 @@
...
+    const albumToCreate = {
+      albumName: 'Album 1',
+      artistName: 'Artist 1',
+      spotifyUrl: 'https://api.spotify.com/v1/albums/blah123',
+    };
+
+    await Album.create(albumToCreate);
+
+    const response = await request(app).get(
+      '/api/v1/albums?limit=10&order=asc'
```
akosasante: This is currently testing the `getAllAlbums` endpoint `/api/v1/albums` which doesn't handle query parameters and will just ignore the `limit` and `order` passed here.

The `getFilteredAlbums` route is defined here: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/8f5ec32895b103423ab561869b2479b76223af5d/src/routes/albumRoutes.js#L24](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/8f5ec32895b103423ab561869b2479b76223af5d/src/routes/albumRoutes.js#L24)

so this url should be updated to:

```suggestion
      '/api/v1/albums/filter?limit=10&order=asc'
```

test/albumController.test.js
```diff
@@ -0,0 +1,190 @@
...
+        averageRating: 4.2,
+        numOfReviews: 1,
+      },
+    ];
+
+    await Album.insertMany(albumsToCreate);
+
+    // Make a GET request to the /api/v1/albums endpoint with a limit parameter set to 2
+    const limit = 2;
+    const response = await request(app).get(`/api/v1/albums?limit=${limit}`);
```
akosasante: Same as above comment about the correct URL to use the filter endpoint that takes a bunch of query parameters:

```suggestion
    const response = await request(app).get(`/api/v1/albums/filter?limit=${limit}`);
```

test/albumController.test.js
```diff
@@ -0,0 +1,190 @@
...
+        category: 'rock',
+        spotifyUrl: 'https://api.spotify.com/v1/albums/blah123',
+        averageRating: 4.2,
+        numOfReviews: 1,
+      },
+    ];
+
+    await Album.insertMany(albumsToCreate);
+
+    // Make a GET request to the /api/v1/albums endpoint with a limit parameter set to 2
```
akosasante:
```suggestion
    // Make a GET request to the /api/v1/albums/filter endpoint with a limit parameter set to 2
```

test/albumController.test.js
```diff
@@ -0,0 +1,190 @@
...
+
+    // Assert that the response body is an object with "albums" and "count" keys
+    expect(response.body).toHaveProperty('albums');
+    expect(response.body).toHaveProperty('count');
+
+    // Assert that "albums" is an array with a length equal to the specified limit
+    expect(response.body.albums).toHaveLength(limit);
+
+    // Assert that the "count" field reflects the actual count of albums in the database
+    // In this case, it should be 5 because we created 5 albums
```
akosasante: We created 3 albums here and there's also the album that gets created before each test in the `beforeEach`. So I would expect that there are 4 total albums in the db.

However, this endpoint returns `count` property that's just the count of albums _returned_ in this response: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/ed2a4e7e04bf48fce393395a8e00a4f3975c6549/src/controllers/albumController.js#L188

akosasante (commented):
You'll want to do `git fetch origin && git merge origin/main` or click the "Update Branch" button here in the PR (and then `git pull` those changes to your local machine) to make sure that your tests are running against the latest version of the code.

test/albumController.test.js
DYA13: I was thinking that it is necessary to provide more detailed condition. But if it works without this part of code, we wouldn't need it

test/albumController.test.js
DYA13: Thank you, it is clear now 👍, commit e3338305ec80449d31eb5dc78f6bf4061f82a794

test/test_helper.js
DYA13: Done 👍, commit bb992665d0a17c0f4f7e5e70f0ee774a172bf987

test/albumController.test.js
DYA13: Thank you, understood 👍

test/albumController.test.js
DYA13: Thank you 👍

DYA13:
Akos when I run debugger , I see that test fail and points on this file, why does it happen like this ? 
image

test/albumController.test.js
DYA13: I have changed the limit of albums for 3 and test results successful, it seems it calculates these albums which are in block

test/test_helper.js
DYA13: loginAndReturnCookie imported in userController.test.js and authController.test.js,   commit  8bf38475e4e72c1cd846926a9b90cc2b1f007888

akosasante (approved):
Looks great Yelena!

PR #67, opened 2023-09-19
Forum room

mvstoyan:
in progress...
Nest step  - 
Implement real-time chat using sockets.
Add real-time message updates.


src/controllers/forumController.js
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    const newRoom = await Forum.create({ name });
+    res.status(StatusCodes.CREATED).json(newRoom);
+  } catch (error) {
+    console.error(error);
+    res
+      .status(error.statusCode || StatusCodes.INTERNAL_SERVER_ERROR)
+      .json({ error: error.message });
+  }
```
akosasante, line 26: try/catch shouldn't be needed here. Any errors raised in the express controllers will be handled by the `error-handler` middleware thanks to our use of the `express-async-errors` library; and that middleware already does the `console.error` and `res.status` bits that you've duplicated here.

src/controllers/forumController.js
```diff
@@ -0,0 +1,68 @@
...
+const getRooms = async (req, res) => {
+  try {
+    const rooms = await Forum.find({});
+    res.status(StatusCodes.OK).json(rooms);
+  } catch (error) {
+    console.error(error);
+    res
+      .status(StatusCodes.INTERNAL_SERVER_ERROR)
+      .json({ error: 'Internal Server Error' });
+  }
```
akosasante, line 38: same comment about try/catch

src/controllers/forumController.js
```diff
@@ -0,0 +1,68 @@
...
+      throw new CustomError.NotFoundError('Room not found');
+    }
+
+    res.status(StatusCodes.OK).json(room);
+  } catch (error) {
+    console.error(error);
+    res
+      .status(error.statusCode || StatusCodes.INTERNAL_SERVER_ERROR)
+      .json({ error: error.message });
+  }
```
akosasante, line 61: same comment about try/catch

src/live/setupSocket.js
akosasante, line 1: Is this file being called from anywhere?

src/live/chatHandlers.js
```diff
@@ -0,0 +1,7 @@
+const sendChatMessage = (io, socket, message) => {
+  console.log(`Received chat message: ${message}`);
+
+  io.emit('chatMessage', message);
```
akosasante, line 4: This will emit the chat message to _all_ connected clients. Do we want this or is this chat message event only meant for some subset of clients (eg: only for a particular user? only for a particular room? do we want the user who sent the message to also receive the message?)

akosasante (commented):
Could you describe in your PR the approach you're planning here? How will the new Forum model be used? When/where in the frontend would requests to the new ForumController be made vs using the real-time sockets?

The code so far looks fine, but I'm not clear on the goals you're aiming to accomplish with these changes.

akosasante:
Based on our chat in Slack we'll close PR for now (it can always be reopened at a later date :) )

PR #68, opened 2023-09-19
Reset password endpoint ps 164

NatachaKey:


NatachaKey:
Needs testing- after review/ question inside

src/controllers/authController.js
```diff
@@ -158,9 +158,47 @@ const forgotPassword = async (req, res) => {
   res.status(StatusCodes.OK).json({ message: 'Password reset email sent' });
 };
 
+//reset Password endpoint
+const reset_password = async (req, res) => {
```
akosasante: For consistency, we've been using `camelCase` throughout the app for function names, so let's stick to that here instead of `snake_case`

```suggestion
const resetPassword = async (req, res) => {
```

src/controllers/authController.js
```diff
@@ -158,9 +158,47 @@ const forgotPassword = async (req, res) => {
...
+  const { passwordToken, newPassword } = req.body;
+  //AKOS: How would we receive this passwordToken from the frontend?
```
akosasante: When the user receives their "forgot password" email, they were given a link to the "reset password" page that included the token as a query parameter: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/7195c0c10fe8f2cfeee15e887c094408524535e6/src/mailing/sender.js#L54](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/7195c0c10fe8f2cfeee15e887c094408524535e6/src/mailing/sender.js#L54)

The frontend should pass that query parameter back to the backend in this `resetPassword` endpoint.

src/controllers/authController.js
```diff
@@ -158,9 +158,47 @@ const forgotPassword = async (req, res) => {
...
+    passwordResetExpiresOn: { $gt: new Date() }, // Check if reset token is still valid
+  });
+
+  if (!user) {
+    throw new CustomError.NotFoundError('User not found');
+  }
+
+  // Check if password reset token is still valid
+  const now = new Date();
+  if (user.passwordResetExpiresOn < now) {
```
akosasante: This was already checked on line 168. The user would not have been returned from our mongo query if the `passwordResetExpiresOn` date was in the past. 

If you prefer to check them separately to give more granular error messages; then you can remove the `passwordResetExpiresOn` filter from the earlier query.

src/controllers/authController.js
```diff
@@ -158,9 +158,47 @@ const forgotPassword = async (req, res) => {
...
+  const { passwordToken, newPassword } = req.body;
+  //AKOS: How would we receive this passwordToken from the frontend?
+  // Find the user with the provided password token
+  const user = await User.findOne({
+    passwordResetToken: passwordToken,
+    passwordResetExpiresOn: { $gt: new Date() }, // Check if reset token is still valid
+  });
+
+  if (!user) {
+    throw new CustomError.NotFoundError('User not found');
```
akosasante, line 172:
```suggestion
    throw new CustomError.NotFoundError('User not found or reset token has expired');
```

src/controllers/authController.js
```diff
@@ -158,9 +158,47 @@ const forgotPassword = async (req, res) => {
...
+  user.passwordResetExpiresOn = undefined;
+  await user.save();
+
+  res.json({ message: 'Password reset successful' });
+
+  /* add later
+    #swagger.summary 
+    #swagger.description 
+    #swagger.responses 
+  */
```
akosasante: :+1:

src/controllers/authController.js
NatachaKey: @akosasante sure! done

src/controllers/authController.js
NatachaKey: @akosasante thank you!! just added to .env BACKEND_BASE_URL_AUTH_PROD=https://musicstore.onrender.com/api/v1/auth
BACKEND_BASE_URL_AUTH_DEV=https://localhost:8000/api/v1/auth

NatachaKey: @akosasante so this part is wrong- right? if frontend passes 'token' query parameter back to the backend in this resetPassword endpoint- it doesn't come from req.body, but from req.query.token?
image so we could modify this: 
  ```
const {  newPassword } = req.body; 
  const { passwordToken} = req.query.token ;
```
right?

File: src/controllers/authController.js (hunk 3 above)
Comment (NatachaKey) [reply src/controllers/authController.js]: @akosasante sure! wanted to do that but forgot to!

Comment (NatachaKey) [reply src/controllers/authController.js]: I was following tickets's instructions and write explicit error messages- but I will leave two options in the code(1st one is commented out)

File: src/controllers/authController.js (hunk 2 above)
Comment (akosasante) [reply src/controllers/authController.js]: Hmm, are we sure that we need separate env variables for PROD vs DEV? Or can we use a single environment variable (`BACKEND_BASE_URL`) and if it's in local dev, it will be localhost because that's what you'll have in your `.env` file; and if it's in Render.com it'll get that same variable from the Environment Variables settings.

Comment (akosasante) [reply src/controllers/authController.js]: The user will be brought to a page: `<url>?token=<resetToken>`. The page will have inputs for them to type in their new password and submit the form. The frontend can choose to pass that data to this endpoint in the request body; it doesn't necessarily have to pass it to the endpoint in the query parameter just because it received the info in a query parameter. Whichever is easier. I would suggest that the `body` is more common for a POST request like this. 

Example vanilla js code:

```js
const urlParams = new URLSearchParams(window.location.search);
const token = urlParams.get('token');

await fetch("/api/v1/auth/reset_password", {
  method: "POST",
  headers: {
    "Content-Type": "application/json",
  },
  body: JSON.stringify({
    newPassword: "newPassword",
    passwordToken: token,
  }),
});

```

src/mailing/sender.js
```diff
@@ -51,7 +51,7 @@ async function sendForgotPasswordEmail(toEmail, resetToken) {
       template: 'forgot_password',
       message: { to: toEmail },
       locals: {
-        resetLink: `${process.env.BACKEND_BASE_URL}/reset_password?token=${resetToken}`,
+        resetLink: `${process.env.BACKEND_BASE_URL_AUTH_PROD}/resetPassword?token=${resetToken}`, //or BACKEND_BASE_URL_AUTH_DEV in development
```
akosasante: Just keep as `BACKEND_BASE_URL` . Prod vs Dev won't require a code change in that case, just a matter of changing the value in the environment variable itself.

akosasante: If you're wondering if the same should be the case for Order.js where we use `DEV_ORDER_EXPIRY_DURATION` and `PROD_ORDER_EXPIRY_DURATION` ; the answer is yeah we could probably simplify that one as well by just having a single environment variable instead :shrug:

src/controllers/authController.js
NatachaKey: @akosasante thank you so much for the explanation!!!

src/mailing/sender.js
NatachaKey: @akosasante I thik I understand- so I just live it in .env as` BACKEND_BASE_URL=https://localhost:8000/api/v1/auth ` in local file, right?
BUT in render.com I should save it as `BACKEND_BASE_URL=https://musicstore.onrender.com/api/v1/auth`
, right? 

 I was trying to enter to render account to add this new variable- but the credentials are not valid, how can I acces it? https://codethedreamteam4-2023.atlassian.net/browse/PS-68  https://vault.bitwarden.com/#/send/M22jkSX9DUu0zLBEACPPgQ/7N7VgBqszBwDKanWNe4qBA

akosasante: Yes exactly :+1: 

Looks like that one expired; try this link (and feel free to save the credentials to your local password manager :smile: ): https://send.bitwarden.com/#wHuUS7Dl8kiRnLBpAQu4eQ/VAIpn8a9nfb6BICMdgJPjA

NatachaKey: @akosasante sure! thank you!  just created an account on bitwarden to store passwords/ didn't know about this service

NatachaKey: @akosasante tested it in Postman, and added variables to render.com. can I merge ?

akosasante (approved):

PR #69, opened 2023-09-22
PS184-albumChat

mvstoyan:


mvstoyan:
I’m sure I have more than one error, I keep getting 404 in postman, can you help me plz😒

akosasante (commented):
While we could do the socket join step via an HTTP request, that then sets up the socket; it is not very efficient and may cause us more complications down the line. 

Instead, let's remove the changes here in `albumRoutes` and `albumController`. 

What you have in `src/live/index.js` looks good and is sufficient for setting up a Socket room in the backend. Next we can create a page (either in EJS or in the React app) that loads an "Album Chat" page and makes a WebSocket request (rather than HTTP request) to our server to that `join:album_chat` socket endpoint to subscribe to events for that album room.

mvstoyan:
@akosasante subscribeToAlbumChat removed from albumController and albumRoutes. If everythink ok now, i will try create a page in the React app.

akosasante (approved):
:+1:

PR #70, opened 2023-09-24
Album test write endpoints

DYA13:


DYA13:
Akos , I have an error on line 199 , but I find the way to resolve it

test/albumController.test.js
```diff
@@ -173,24 +159,130 @@ describe('AlbumController API Tests', () => {
 
     await Album.insertMany(albumsToCreate);
 
-    // Make a GET request to the /api/v1/albums/filter endpoint with a limit parameter set to 4
     const limit = 3;
     const response = await request(app).get(
       `/api/v1/albums/filter?limit=${limit}`
     );
+    expect(response.status).toBe(StatusCodes.BAD_REQUEST);
```
akosasante: Why do we expect it the response to be a `BAD_REQUEST` code here and then in the next line an `OK` ?

test/albumController.test.js
```diff
@@ -173,24 +159,130 @@ describe('AlbumController API Tests', () => {
...
+  
+ 
+    expect(response.status).toBe(StatusCodes.CREATED);
+    expect(response.body).toHaveProperty('album');
+  
   });
-});
+
+ 
+    const response = await request(app)
```
akosasante, line 185: The linter is saying `test/albumController.test.js: SyntaxError: 'await' is only allowed within async functions and at the top levels of modules. (199:22)`

So we need to check which function this part of and why it doesn't have the `async` keyword. 

It currently looks like the `should create a new album - Success Case` test has an opening brace `{` on line 176 and a closing brace `})` on line 196. So this response part is not inside of any test, it's just inside the `describe` block that started on line `27`

test/albumController.test.js
```diff
@@ -173,24 +159,130 @@ describe('AlbumController API Tests', () => {
...
+      .post('/api/v1/albums')
+      .set('Cookie', signedCookie) 
+      .send(newAlbumData);
+
+    expect(response.status).toBe(StatusCodes.CREATED);
+    expect(response.body).toHaveProperty('album');
+    expect(response.body.album.albumName).toBe(newAlbumData.albumName);
+    expect(response.body.album.artistName).toBe(newAlbumData.artistName);
+
+  });
```
akosasante, line 193: Currently this set of brackets is closing the `describe` block and all the tests under this are outside of the describe block; which means they're not getting the `beforeAll` and `beforeEach` steps defined in that `describe` block

test/albumController.test.js
```diff
@@ -173,24 +159,130 @@ describe('AlbumController API Tests', () => {
...
-    // Assert that "albums" is an array with a length equal to the specified limit
     expect(response.body.albums).toHaveLength(limit);
+    expect(response.body.count).toBe(albumsToCreate.length);
+  });
 
-    // Assert that the "count" field reflects the actual count of albums in the database
+  it('should create a new album - Success Case', async () => {
 
-    expect(response.body.count).toBe(albumsToCreate.length);
+    const user = await authenticateUser(User);
```
akosasante: Where is `authenticateUser` function defined? Why is it taking just the `User` mongoose model instead of a defined user object?

test/albumController.test.js
```diff
@@ -173,24 +159,130 @@ describe('AlbumController API Tests', () => {
...
+      const newAlbumData = {
+        albumName: 'New Album',
+        artistName: 'New Artist',
+        price: 9.99,
+        spotifyUrl: 'https://api.spotify.com/v1/albums/new123',
+    };
+  
+    const response = await request(app)
+      .post('/api/v1/albums')
+      .set('Cookie', `token=${user.token}`)
```
akosasante: We should be able to get a cookie the same way you've done in the `should test the getAlbumWithAllUsersWhoPurchasedIt endpoint - Success Case` test

akosasante (commented):
Double check your brackets for the different blocks. In VS Code you can use the colours of the brackets to match them up and the vertical lines on the left-side will show where each block starts/ends

test/albumController.test.js
DYA13: Removed bad request status , commit 2cb2ecd237dc13b8d6c8b3c04eaddcac5b6345a6

test/albumController.test.js
DYA13: I have controlled the curly braces , should be ok now

test/albumController.test.js
DYA13: yes , actually there was an issue with the curly braces and the part of the code was out of the block

test/albumController.test.js
```diff
@@ -173,24 +155,151 @@ describe('AlbumController API Tests', () => {
...
+      spotifyUrl: 'https://api.spotify.com/v1/albums/new123',
+    };
+
+    const response = await request(app)
+      .post('/api/v1/albums')
+      .set('Cookie', signedCookie)
+      .send(newAlbumData);
+
+    expect(response.status).toBe(StatusCodes.CREATED);
+    expect(response.body).toHaveProperty('album');
```
akosasante, line 191: It would be good to also have an assertion to verify that the returned album is what we expect. You could test that `response.body.album.albumName` matches what we put into the request.

Or you could use the `toMatchObject` [Jest matcher](https://jestjs.io/docs/expect#tomatchobjectobject), to verify that `newAlbumData` is a subset of the returned album in the response body: `expect(response.body.album).toMatchObject(newAlbumData);`

test/albumController.test.js
```diff
@@ -173,24 +155,151 @@ describe('AlbumController API Tests', () => {
...
+      price: 14.99,
+    };
+
+    const response = await request(app)
+      .patch(`/api/v1/albums/${existingAlbum._id}`)
+      .set('Cookie', signedCookie)
+      .send(updatedData);
+
+    expect(response.status).toBe(StatusCodes.OK);
+    expect(response.body).toHaveProperty('album');
```
akosasante, line 284: Liek you've done in the `should update an existing album - Success Case` test , it would be good to also have a check here to verify that the album in the db was actually successfully updated.

akosasante (approved):
Looks good to me; you can feel free tomerge but I've included some small suggestions for buffing up some of the tests that you should familiarize yourself with. You can choose to add those here or just leave as-is, I'm more just leaving them here for learning :+1:

test/albumController.test.js
DYA13: Akos, I have added verification that the returned album matches the expected data using .toMatchObject(object), commit ae6985f501b4a99e161bed85decd056994ea1d52

test/albumController.test.js
DYA13: Done , commit 1b7dec473b617cdcff936d78a14387142cde8cac

DYA13:
I have applied your suggestions 👍, thank you

akosasante (approved):
LGTM

PR #71, opened 2023-09-26
PS186-Chat-album-message-handling

mvstoyan:


src/expressServer.js
```diff
@@ -168,6 +172,11 @@ io.on('connection', (socket) => {
   console.log('Connected');
   setupSocket(io, socket); // Call your setupSocket function
 
+  socket.on('chat:album', function (data) {
+    const { handleAlbumChat } = require('./live/handleAlbumChat');
+    handleAlbumChat(io, socket, data);
+  });
```
akosasante: We already have this a listener for this event name/channel defined in `live/index.js` we shouldn't need to have it duplicated here. Generally either place is fine to define our listeners, but to try and keep things organized, similar to how we have our HTTP routes defined in the `routes` and `controllers` folders; we'd prefer to have our socket listeners and handlers defined in the `live` folder.

src/live/handleAlbumChat.js
```diff
@@ -0,0 +1,14 @@
+function handleAlbumChat(io, socket, data) {
+  const { message, userId, albumId } = data;
+
+  // Create a chat room name for the album chat
+  const chatRoomName = `chat:album:${albumId}`;
+
+  // Join the album chat room
+  socket.join(chatRoomName);
```
akosasante: The socket _should_ already have joined this room (by sending a \`join:album\_chat\` message earlier) to be sending a chat to it now; so we probably don't need this line. But it's probably fine to leave for now as a way to ensure they're in the room I suppose :thinking:

src/expressServer.js
```diff
@@ -159,7 +159,11 @@ app.use(errorHandlerMiddleware); // Error handler middleware
 const { Server } = require('socket.io');
 const http = require('http');
 const server = http.createServer(app);
-const socketServer = new Server(server);
+const socketServer = new Server(server, {
+  cors: {
+    origin: [/localhost:3000$/, /beatbazaar\.onrender\.com$/],
+  },
+});
```
akosasante, line 166: :+1:

src/expressServer.js
mvstoyan: Removed

src/live/handleAlbumChat.js
mvstoyan: Removed

mvstoyan: postman

mvstoyan: @akosasante Look like its work

akosasante:
@mvstoyan https://streamable.com/02c9bz

src/live/handleAlbumChat.js
mvstoyan: @akosasante Thank you so much for such a detailed and clear explanation. I'd love to take any programming course from you. You explain things so wonderfully!   

I tried as you explained, and everything works. If possible, it would be good to leave the console.log for further exploration, and the girls can also watch the video and test it themselves.

mvstoyan:
> @mvstoyan https://streamable.com/02c9bz

@akosasante Thank you so much for such a detailed and clear explanation. I'd love to take any programming course from you. You explain things so wonderfully!

I tried as you explained, and everything works. If possible, it would be good to leave the console.log for further exploration, and the girls can also watch the video and test it themselves.

akosasante (approved):
:tada:

src/live/index.js
```diff
@@ -12,8 +14,18 @@ const onConnect = (io, socket) => {
...
     const chatRoomName = `chat:album:${albumId}`;
+    console.log(
+      `received message from socket: ${socket.id}. Will join room: ${chatRoomName}`
+    );
     socket.join(chatRoomName);
   });
+
+  socket.on('chat:album', (data) => {
+    console.log(
+      `received message from socket: ${socket.id}. Message=${data.message} | Album=&{data.albumId} | User=${data.userId}`
```
akosasante:
```suggestion
      `received message from socket: ${socket.id}. Message=${data.message} | Album=${data.albumId} | User=${data.userId}`
```

PR #72, opened 2023-09-26
[Ps-142] tests for the review controller - read endpoints

NatachaKey:


test/reviewController.test.js
```diff
@@ -0,0 +1,113 @@
...
+
+afterAll(async () => {
+  await server.close();
+  await mongooseConnection.disconnect();
+  await mongodb.stop();
+});
+
+describe('ReviewController API Tests', () => {
+  beforeAll(async () => {
+    await Review.deleteMany({});
```
akosasante, line 29: Since we're doing this before each test anyway; it's redundant / not needed to do it at very beginning before the tests run.

test/reviewController.test.js
```diff
@@ -0,0 +1,113 @@
...
+  beforeEach(async () => {
+    await Review.deleteMany({});
+    // Create a unique album and user for each test
+    const album = await Album.create({
+      albumName: 'Unique Album',
+      artistName: 'Unique Artist',
+      spotifyUrl: 'https://api.spotify.com/v1/albums/unique',
+    });
+
+    const user = await User.findOne({ username: 'emily123' });
```
akosasante: Instead of setting the `user` variable here by fetching from the database, we can set the variable earlier on in the `beforeAll` block when we create the user. 

On line 30; we can do `user = await User.create(...)`

---

Though to be fair, in that case, we'd need to declare the `user` variable with `let user` somewhere outside of both blocks so that both blocks have access to that variable (it's "in scope" for both blocks). This could be for example, at the top of the `describe` block (between line 27 and 28)  or outside of the `describe` block like where we declare our server, mongodb, and mongooseConnection variables.

test/reviewController.test.js
akosasante, line 33: Since we're doing this before each test anyway (line 40, `beforeEach` block); it's redundant / not needed to do it at very beginning before the tests run.

test/reviewController.test.js
```diff
@@ -0,0 +1,113 @@
...
+    };
+
+    await Review.create(mockReviewData);
+  });
+  //get all reviews
+
+  it('should test the getAllReviews endpoint - Success Case', async () => {
+    const response = await request(app).get('/api/v1/reviews');
+    expect(response.status).toBe(StatusCodes.OK);
+    expect(response.body).toHaveProperty('reviews');
```
akosasante, line 68: could we also check for the presence and correct value of the `count` property of the response body in this test?

test/reviewController.test.js
```diff
@@ -0,0 +1,113 @@
...
+    expect(response.body.count).toBe(0);
+  });
+
+  //get single review
+  it('should test the getSingleReview endpoint - Success Case', async () => {
+    const { _id: reviewId } = await Review.findOne({});
+
+    const response = await request(app).get(`/api/v1/reviews/${reviewId}`);
+    expect(response.status).toBe(StatusCodes.OK);
+    expect(response.body).toHaveProperty('review');
```
akosasante, line 98: We could also further check if the returned review is actually the one we expect. Eg; we could check if the `response.body.review._id` value matches what we expect.

Or if we got the whole review on line 89 instead of destructuring it, then we could compare the JSON value of the review to the returned value of the response body with something like:

```javascript
expect(response.body.review).toMatchObject(JSON.parse(JSON.stringify(review)))
```

[toMatchObject](https://jestjs.io/docs/expect#tomatchobjectobject) checks if the object on the right is a _subset_ (ie: has at least some of the same fields and values) as the object on the left. 

It's a bit messy with Mongoose objects that we have to do that circular JSON.parse+JSON.stringify step so that all the field values are converted to JSON types instead of Javascript native types (mostly an issue with datetime fields like `created_at`).

test/reviewController.test.js
```diff
@@ -0,0 +1,113 @@
...
+  });
+
+  it('should test the getSingleReview endpoint - Error Case (Not Found)', async () => {
+    const response = await request(app).get('/api/v1/reviews/errorReviewId');
+    expect(response.status).toBe(StatusCodes.NOT_FOUND);
+  });
+
+  //get all reviews for this particular album
+  it('should get reviews for a specific album', async () => {
```
akosasante, line 111: What happens if no reviews are found for the given album id?

test/reviewController.test.js
```diff
@@ -0,0 +1,113 @@
...
+    // Try to find an album; if not found, create a default one
+    const album = await Album.findOne({ albumName: 'Unique Album' });
+    const response = await request(app).get(
+      `/api/v1/reviews/album/${album._id}`
+    );
+
+    expect(response.statusCode).toBe(200);
+    expect(response.body).toHaveProperty('allProductReviews');
+    expect(response.body).toHaveProperty('count');
```
akosasante: We could also assert if the values of these two fields are what we expect

test/reviewController.test.js
akosasante: This also applies to the album and review we create here. The nice thing about setting them to a variable outside of this `beforeEach` block scope is that then we can refer to them in the tests without having to do another query to fetch the album or review just to be able to arrange our test case.

akosasante (approved):
Looks great! You can feel free to merge as-is as everything is passing and covers good test cases. But I've left some further suggestions to tweak or buff up some of the tests if you'd like to add a few things.

test/reviewController.test.js
NatachaKey: @akosasante Akos, I was thinking about beforeAll and beforeEach blocks- in most cases propably we can delete   beforeEach block , they are doing practically the same thing - and beforeEach is just a bit more meticulous, isn't it?

test/reviewController.test.js
NatachaKey: done

test/reviewController.test.js
NatachaKey: @akosasante thank you!!

test/reviewController.test.js
NatachaKey: @akosasante just added  
  ```
// If no reviews are found, allProductReviews should be an empty array and count should be 0
  if (response.body.count === 0) {
    expect(response.body.allProductReviews).toEqual([]);
  }
```

test/reviewController.test.js
NatachaKey: done

PR #74, opened 2023-09-29
[PS-203] Album filter endpoint modifyed to show total property

NatachaKey:


src/controllers/albumController.js
```diff
@@ -185,7 +188,7 @@ const getFilteredAlbums = async (req, res) => {
     .skip(parseInt(offset) || 0) // Skip a specified number of albums (pagination implementation)
     .limit(parseInt(limit) || 10); // Limit the number of returned albums (pagination implementation)
 
-  res.status(StatusCodes.OK).json({ albums, count: albums.length }); // Return the found albums and the count of albums
+  res.status(StatusCodes.OK).json({ albums, count: totalCount }); // Return the found albums  and the total count
```
akosasante: I think it can still be helpful to have `count` returned to the front end to indicate how many albums are on this page (of course they could just do `albums.length` themselves, but it's good practice to include) _in addition to_ a `total` field. (some examples can be seen in this [SO page](https://stackoverflow.com/questions/12168624/pagination-response-payload-from-a-restful-api))

Another thing we _could_ add (though not necessary) is a boolean field called `more`. This will be `true` if there are more results to be shown or `false` otherwise.  Alternatively, we could have `next` and `prev` page pointers that could be a string that basically reflects the same `query` that was passed in,  but with the `offset` changed that the client can use to fetch the next/previous page. And it would respectively be `null` if they are already on the first/last page.

If you add either of those additional ideas (one or the other; having both `more` _and_ `next/prev` probably doesn't make sense) then you'd have to do a bit of calculation here in the controller to determine if those fields should be populated or not. 

For example, if there's 100 albums that match the current query, and the request asked for `limit=5&offset=10`, then they would be given the 3rd page of albums (albums 11-15):

*   We can calculate the total number of pages by doing `totalCount / limit` (in this case there's 20 pages)
*   We can calculate the current page we're on by doing `1 + (offset / limit)` (we add 1, because we want the first page to be (1) not (0). In this case, that would give us page 3)
*   So based on that we know that if we're on page 3, there are `more=true` pages available
*   The query for the next page would be the same as the current one but with `newOffset = offset + limit`
*   The query for the previous page would be the same as the current query but with `newOffset = offset - limit`

NatachaKey: image
@akosasante Akos, why it's 20 here-
```
For example, if there's 100 albums that match the current query, and the request asked for limit=5&offset=10, then they would be given the 3rd page of albums (albums 11-15):

We can calculate the total number of pages by doing totalCount / limit (in this case there's 20 pages)

``` 
I thought we should divide 69 (always the same)/ 5 (we establish ourselves)= 13,8 = 14 pages. don't get where this 20 came from

NatachaKey: PS oops sorry I pushed some code from other ticket- it's not finished yet, will switch the branch now /maybe will get come conflicts later- should have done only `git add albumController.js`

akosasante: Right in my example, we had _100 results that matched_. And the user asked for a `limit` of 5 results per page. So if we divide those, 100/5 = 20, so 20 total pages.

Your screenshot looks correct in that if the total number of albums in the db that match the query (in this case, you're not filtering by any album name or artist name, so it's just the total number albums altogether in the db) is 69. The request asked for a `limit` of 5 results per page. So `totalPages = 69 / 5 = 13.8` and since we can't have a fraction of a page, we always in this case will round _up_ (`Math.ceil` in javascript) so that is `14 total pages` :white\_check\_mark: 

But to clarify, it won't _always_ be 69. What if the request was `/api/v1/albums/filter?limit=5&offset=0&artistName=me` there's only **7 total matching results**. So on the first page, we can show matching albums 1-5. And on the second page we'd show album 6-7. And then there'd be no more pages. So to calculate `totalPages` we'd do `totalPages = 7 / 5 = 1.4 ... round up to 2`.

src/controllers/albumController.js
```diff
@@ -185,7 +188,31 @@ const getFilteredAlbums = async (req, res) => {
...
+      ? `${req.originalUrl.split('&offset=')[0]}&offset=${
+          parseInt(offset) + (parseInt(limit) || 10)
+        }`
+      : null,
+    prevPage:
+      currentPage > 1
+        ? `${req.originalUrl.split('&offset=')[0]}&offset=${
+            parseInt(offset) - (parseInt(limit) || 10)
+          }`
+        : null,
```
akosasante: Overall this looks fine but I notice you're splitting on `&offset`. But what if the request was made like this: `/api/v1/albums/filter?offset=5&limit=10&artistName=b` ; I think then your code here won't reconstruct the `nextPage`/\`prevPage\` correctly (though I have not tried it myself yet here)

A preferable way to handle these may be to use the Node built-in (I think at least since version 18 and up) `URL` and `URLSearchParams` modules:

*   [https://nodejs.org/api/url.html#class-urlsearchparams](https://nodejs.org/api/url.html#class-urlsearchparams)
*   [https://www.geeksforgeeks.org/node-js-urlsearchparams-api/](https://www.geeksforgeeks.org/node-js-urlsearchparams-api/)
*   [https://www.youtube.com/watch?v=np53e9_Yilk](https://www.youtube.com/watch?v=np53e9_Yilk)

So we could do something like:

```javascript
const url = new URL(req.originalUrl); // create a "URL" object / class instance from the request
url.searchParams.set('offset': '<whatever the new offset should be>')
// then do url.toString() (or url.href) to get the final updated url string
```

test/reviewController.test.js
akosasante, line 1: unneeded change?

src/controllers/albumController.js
```diff
@@ -185,7 +188,31 @@ const getFilteredAlbums = async (req, res) => {
...
+  }); 
   /*
```
akosasante, line 233: don't forget to update the swagger doc comments with the new fields

src/controllers/albumController.js
```diff
@@ -185,7 +188,31 @@ const getFilteredAlbums = async (req, res) => {
     .skip(parseInt(offset) || 0) // Skip a specified number of albums (pagination implementation)
     .limit(parseInt(limit) || 10); // Limit the number of returned albums (pagination implementation)
 
-  res.status(StatusCodes.OK).json({ albums, count: albums.length }); // Return the found albums and the count of albums
+  const totalPages = Math.ceil(totalCount / (parseInt(limit) || 10));
+  const currentPage = Math.ceil(
+    (parseInt(offset) + 1) / (parseInt(limit) || 10)
```
```suggestion
    ((parseInt(offset) || 0) + 1) / (parseInt(limit) || 10)
```

^ in case `offset` was not given in the query params and is undefined/null

src/controllers/albumController.js
akosasante: Also keep in mind what you want to do if an `offset` was not given here.

src/controllers/albumController.js
NatachaKey: @akosasante oh that was an example!! sorry, didn't understand correctly, I thought we were talking about the actual db, my bad!

test/reviewController.test.js
NatachaKey: @akosasante I just added it to another PR to avoid the confusion

src/controllers/albumController.js
```diff
@@ -161,6 +161,15 @@ const getAlbumWithAllUsersWhoPurchasedIt = async (req, res) => {
...
 const getFilteredAlbums = async (req, res) => {
   const { limit, order, offset, albumName, artistName } = req.query;
+
+  console.log('reqoriginalUrl is', req.originalUrl);
+
+  const url = new URL('http://localhost:8000' + req.originalUrl); // create a "URL" object
+  //Akos: req.originalUrl solo gives an error
+
+  const parsedOffset = parseInt(offset, 10);
+  const parsedLimit = parseInt(limit, 10);
```
akosasante: to save having to do the comparisons `parsedOffset || 0` and `parsedLimit || 10` all throughout the code, I _think_ we can clean things up by doing that just one time here:

```suggestion
  const parsedOffset = parseInt(offset, 10) || 0;
  const parsedLimit = parseInt(limit, 10) || 10;
```

akosasante (approved):
Great job! Left one small optional suggestion

Feel free to merge after CI is passing and after adding new fields to the swagger docs comments  :+1:

PR #75, opened 2023-09-29
[PS-143] test write review endpoints

NatachaKey:


test/reviewController.test.js
```diff
@@ -124,4 +124,95 @@ describe('ReviewController API Tests', () => {
...
+  it('should create a review successfully', async () => {
+    await Review.deleteMany({});
+    const userCredentials = {
+      email: 'Emily@google.com',
+      password: 'secret',
+    };
+
+    const signedCookie = await loginAndReturnCookie(userCredentials);
+
+    const album = await Album.findOne({ albumName: 'Unique Album' });
```
akosasante: Since we store this album in a variable already (declared on line 29, and assigned in the `beforeAll` block on line 42), there's not really any need to go make a round-trip to the database to fetch it here in this test case. 

We can delete this line; and all the tests will still pass because we'll be referring to `album` that we created and stored in that variable earlier, without unnecessarily re-fetching it

test/reviewController.test.js
```diff
@@ -124,4 +124,95 @@ describe('ReviewController API Tests', () => {
...
+      .post(`/api/v1/reviews/album/${album._id}`)
+      .set('Cookie', signedCookie)
+      .send(newReviewData);
+
+    expect(response.status).toBe(StatusCodes.CREATED);
+    expect(response.body).toHaveProperty('review');
+    // If no album was found
+    if (!album) {
+      expect(response.status).toBe(StatusCodes.NOT_FOUND);
+    }
```
akosasante: Will this block ever happen to be true in this test? It doesn't seem to be possible. This should be written as a separate test case where we pass in an invalid album ID or delete the album before submitting the review or something?

test/reviewController.test.js
```diff
@@ -124,4 +124,95 @@ describe('ReviewController API Tests', () => {
...
+    // this user already submitted a review for this album
+    expect(res.status).toBe(StatusCodes.CONFLICT);
+  });
+
+  //  update a review
+  it('should update a review successfully', async () => {
+    const userCredentials = {
+      email: 'Emily@google.com',
+      password: 'secret',
+    };
```
akosasante: Since we use this in two separate tests; it may be worthwhile to move this to an upper scope so that we can refer to the same variable in both tests. That way if the password or email ever has to be changed (eg: let's say in the future we decide we only want to allow passwords longer than 6 characters), then there's just one place to look to make that change, instead of having to update multiple tests.

test/reviewController.test.js
```diff
@@ -124,4 +124,95 @@ describe('ReviewController API Tests', () => {
...
+
+    const signedCookie = await loginAndReturnCookie(userCredentials);
+
+    const existingReview = await Review.findOne({});
+    // If no existingReview was found
+    if (!existingReview) {
+      expect(response.status).toBe(StatusCodes.NOT_FOUND);
+    }
```
akosasante: I generally don't expect to see much conditional logic in a test. Each test should be testing one, _maybe two_ specific things. It's better to have many small tests, than one long test that is trying to check too many diverging things at once. 

In this case, we just fetched a review from the db at line 182. We _know_ that a review exists since we created it in the `beforeEach` block and we haven't deleted it in this test. So we _know_ that there will always be an existing review. So there's no reason to check _if_ there's no existing review. In test code, we for the most part can control everything about the state of the "world" that we're testing. So we shouldn't need to check multiple branches/optional code paths, instead we should be "Arranging" that things go down a certain path and then "Asserting" that the result of that path is what we expect it to be.

In this case, it would be better instead to have a separate test that asserts the `NOT_FOUND` result.

test/reviewController.test.js
```diff
@@ -124,4 +124,95 @@ describe('ReviewController API Tests', () => {
...
+
+    const updatedReviewData = {
+      rating: 1,
+      title: 'Don not buy this album',
+      comment: 'Awful"',
+    };
+    if (!updatedReviewData) {
+      expect(response.status).toBe(StatusCodes.BAD_REQUEST);
+    }
```
akosasante: separate test please

test/reviewController.test.js
```diff
@@ -124,4 +124,95 @@ describe('ReviewController API Tests', () => {
...
+
+    // Check if the user is the author of the review
+    const authenticatedUserId = 'mockUserId'; // Replace with actual authenticated user's ID
+
+    if (existingReview.user.toString() !== authenticatedUserId) {
+      // User is not the author of the review, respond with unauthorized error
+      expect(StatusCodes.UNAUTHORIZED).toBe(StatusCodes.UNAUTHORIZED); // //AKOS: this one works but I feel it's wrong, couldn't write here 'respone', the terminal was complaining
+    }
```
akosasante: Separate test please.

To test if we get `UNAUTHORIZED` when a user tries to update a review that doesn't belong to them, we need to first create a second user that can legitimately log in and receive a signed cookie. Then user that second user's signed cookie to make a request to the endpoint, using the `existingReview._id` that belongs to `Emily`. That should correctly simulate a user trying to update a review that doesn't belong to them.

test/reviewController.test.js
NatachaKey: @akosasante done!!

test/reviewController.test.js
NatachaKey: @akosasante 
done, thank you so much!!

test/reviewController.test.js
NatachaKey: @akosasante done=)

test/reviewController.test.js
NatachaKey: @akosasante thank you, Akos, done!

test/reviewController.test.js
```diff
@@ -28,7 +28,8 @@ describe('ReviewController API Tests', () => {
   let user;
   let album;
   let review;
-
+  let userCredentials;
+  let response;
```
akosasante: I have a bit of concern with pulling `response` up into this outer scope. While it is a variable that we re-use in multiple tests so I see with my previous comments the rationale for adding it here. But the concern stems from the fact that it could lead to hard to debug tests where we miss that `response` was set in an earlier test and not re-set in the later test before trying to do assertions on it.

test/reviewController.test.js
```diff
@@ -102,17 +107,15 @@ describe('ReviewController API Tests', () => {
...
   });
 
   //get all reviews for this particular album
   it('should get reviews for a specific album', async () => {
     // Try to find an album; if not found, create a default one
-    const album = await Album.findOne({ albumName: 'Unique Album' });
-    const response = await request(app).get(
-      `/api/v1/reviews/album/${album._id}`
-    );
+    album = await Album.findOne({ albumName: 'Unique Album' });
```
akosasante: We can remove this line; since `album` already refers to the album from the db that we try to refetch here.

akosasante: As well as the comment in the line preceding it

test/reviewController.test.js
```diff
@@ -124,4 +127,144 @@ describe('ReviewController API Tests', () => {
...
+    response = await request(app)
+      .post(`/api/v1/reviews/album/${album._id}`)
+      .set('Cookie', signedCookie)
+      .send(secondReviewData);
+    // this user already submitted a review for this album
+    expect(response.status).toBe(StatusCodes.CONFLICT);
+  });
+
+  //when album was not found
+  it('should create a review successfully-Error case -album not found ', async () => {
```
akosasante: I find it a bit confusing to say "it should create a review successfully" in this test name, since we _don't_ expect it to create a review. 

We could call this `it('should not create a review and return a 404 if the album is not found')` . Or something along those lines.

---

Another thing we _could_ do is nest `describe` blocks. So while all these test cases (`it` blocks) are currently nested under a single `describe('ReviewController API Tests', () => ...)` block, we could also have `describe` blocks inside that for each of the endpoints. As test files get larger, this can be an effective way to organize the test code to make it easier for someone new to navigate it.

It would look something like:  
 

```javascript
describe('ReviewController API Tests', () => {
  <declare any `let` variables that multiple tests in this describe block need access to>
  
  <write any beforeAll/beforeEach/afterAll/afterEach lifecycle callbacks needed for the tests in this describe block>
  
  describe('test the getAllReviews endpoint', () => {
    <any variables or lifecycle callbacks needed for _just_ the getAllReviews tests. In this case, there may not be anything additional needed here>
    
    it('should return 200 and list of all reviews in the database', async () => {...});
    it('should return 200 and an empty array if there are no reviews in the database', async () => {...});
  });
  
  ..etc for other endpoints
});
```

test/reviewController.test.js
```diff
@@ -124,4 +127,144 @@ describe('ReviewController API Tests', () => {
...
+    };
+
+    response = await request(app)
+      .patch(`/api/v1/reviews/${nonExistingReview._id}`)
+      .set('Cookie', signedCookie)
+      .send(updatedReviewData);
+    expect(response.status).toBe(StatusCodes.NOT_FOUND);
+  });
+
+  it('should update a review successfully-Error case - some updated review data provided is missing/ incorect', async () => {
```
akosasante:
```suggestion
  it('should update a review - Error case - updating the review of another user will result in a 403 forbidden error', async () => {
```

test/reviewController.test.js
```diff
@@ -124,4 +127,144 @@ describe('ReviewController API Tests', () => {
...
+      rating: 5,
+      title: 'Buy this album',
+      comment: 'Excellent"',
+    };
+
+    response = await request(app)
+      .patch(`/api/v1/reviews/${existingReview._id}`)
+      .set('Cookie', signedCookie)
+      .send(updatedReviewData);
+      expect(response.status).toBe(StatusCodes.UNAUTHORIZED);
```
akosasante: Looks like the controller throws an `CustomError.UnauthorizedError` if the user ids don't match here: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/e9bfbdac6499460dfbd35c7e23c2164f5776b0db/src/controllers/reviewController.js#L48-L52](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/e9bfbdac6499460dfbd35c7e23c2164f5776b0db/src/controllers/reviewController.js#L48-L52)

Which in turn returns a `403 Forbidden` status code here: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/0666c4aea5c4054c06840efc984c4cabec5059be/src/errors/unauthorized.js#L10](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/0666c4aea5c4054c06840efc984c4cabec5059be/src/errors/unauthorized.js#L10)

Rather than the `401 Unauthorized` that we've asserted here. 

The code seems to be reasonable, ususally `401` means "you haven't logged in or you're not allowed to do this" and `403` means "you're logged in but don't have access to this particular resource". Sometimes theyr'e used a bit interchangeably but in this case, our test should be updated to match what our code is doing.

test/reviewController.test.js
NatachaKey: @akosasante thank you, will change it back

test/reviewController.test.js
NatachaKey: thank you, Akos!!

test/reviewController.test.js
NatachaKey: @akosasante thank you!! done

test/reviewController.test.js
```diff
@@ -58,70 +59,219 @@ describe('ReviewController API Tests', () => {
...
-    expect(response.status).toBe(StatusCodes.NOT_FOUND);
+      expect(response.statusCode).toBe(200);
+      expect(response.body).toHaveProperty('allProductReviews');
+      expect(response.body.allProductReviews).toHaveLength(1);
+      expect(response.body).toHaveProperty('count');
+      expect(response.body.count).toBe(1);
+      // If no reviews are found, allProductReviews should be an empty array and count should be 0
+      if (response.body.count === 0) {
+        expect(response.body.allProductReviews).toEqual([]);
+      }
```
akosasante: Generally for a test it's a "[code smell](https://en.wikipedia.org/wiki/Code_smell)" for me if I see an `if` statement. For example, in this case, we know that the request we made on line 119 will always have exactly 1 review. So the contents of this if statement will _never_ be executed which means it's not really testing anything.

If we want to test it, we should write test code (probably as a separate test for good "separation of concerns") that creates an Album with no reviews, then does a `/api/v1/reviews/album/${secondAlbum._id}` request and verifies the shape of the response.

test/reviewController.test.js
```diff
@@ -58,70 +59,219 @@ describe('ReviewController API Tests', () => {
...
+      };
+      const response = await request(app)
+        .post(`/api/v1/reviews/album/${album._id}`)
+        .set('Cookie', signedCookie)
+        .send(newReviewData);
+
+      expect(response.status).toBe(StatusCodes.CREATED);
+      expect(response.body).toHaveProperty('review');
+
+      //create second review for the same album
```
akosasante:
```suggestion
      //create second review by the same user for the same album
```

test/reviewController.test.js
```diff
@@ -58,70 +59,219 @@ describe('ReviewController API Tests', () => {
...
+        comment: 'Awful"',
+      };
+
+      // Check if the user is the author of the review
+      const authenticatedUserId = 'mockUserId'; // Replace with actual authenticated user's ID
+
+      if (existingReview.user.toString() !== authenticatedUserId) {
+        // User is not the author of the review, respond with unauthorized error
+        expect(StatusCodes.UNAUTHORIZED).toBe(StatusCodes.UNAUTHORIZED); // //AKOS: this one works but I feel it's wrong, couldn't write here 'respone', the terminal was complaining
+      }
```
akosasante: We can remove these lines; they're not checking anything within the test specifically; and I think you have already updated the code to test the case where a non-authorized user tries to update a review :+1:

test/reviewController.test.js
```diff
@@ -58,70 +59,219 @@ describe('ReviewController API Tests', () => {
...
+      expect(response.body).toHaveProperty('review');
+
+      const updatedReview = await Review.findById(existingReview._id);
+      expect(updatedReview).not.toBeNull();
+      expect(updatedReview.rating).toBe(updatedReviewData.rating);
+      expect(updatedReview.title).toBe(updatedReviewData.title);
+      expect(updatedReview.comment).toBe(updatedReviewData.comment);
+    });
+
+    it('should update a review successfully-Error case - no existing  review was found', async () => {
```
akosasante:
```suggestion
    it('should NOT sucessfully update a review - Error case - no existing  review was found', async () => {
```

akosasante (approved):
Small cleanup comments, and double check CI. Otherwise, LGTM :rocket:

akosasante:
P.S. If the formatting that Github Actions is complaining about is from outside of your PR; you may need to do `git merge main` (or Update Branch here in Github) to get the latest changes where I've run the format/lint commands on `main` branch.

test/reviewController.test.js
NatachaKey: thank you, done!

NatachaKey:
> P.S. If the formatting that Github Actions is complaining about is from outside of your PR; you may need to do `git merge main` (or Update Branch here in Github) to get the latest changes where I've run the format/lint commands on `main` branch.

just  curious @akosasante - when we do in our branch npm run format- everytime before creating PR- aren't we applying these changes to main when we merge this branch with main- just wonderind how could main be without format if we apply it before  making PR

test/reviewController.test.js
NatachaKey: @akosasante thank you! I felt it was not that much appropriate to use `if `there...  just added this test case, please take a look when you have a moment

akosasante:
> when we do in our branch npm run format- everytime before creating PR- aren't we applying these changes to main when we merge this branch with main- just wonderind how could main be without format if we apply it before making PR

It _shouldn't_ be happening, because `husky` library should be running every time y'all make a commit. I'm not sure if there's a bug somewhere in the setup or if `husky` is not running correctly. 

Do you find that you see the following `lint-staged` message when you make your commits? Do you usually make your commit in the command line or using VS Code's "Source Control" tab? Is your teminal the Windows cmd terminal or is it PowerShell (does it show `PS` at the beginning of each line?)

https://streamable.com/g10bef

NatachaKey:
> > when we do in our branch npm run format- everytime before creating PR- aren't we applying these changes to main when we merge this branch with main- just wonderind how could main be without format if we apply it before making PR
> 
> It _shouldn't_ be happening, because `husky` library should be running every time y'all make a commit. I'm not sure if there's a bug somewhere in the setup or if `husky` is not running correctly. 
> 
> Do you find that you see the following `lint-staged` message when you make your commits? Do you usually make your commit in the command line or using VS Code's "Source Control" tab? Is your teminal the Windows cmd terminal or is it PowerShell (does it show `PS` at the beginning of each line?)
> 
> https://streamable.com/g10bef

no, I don't get lint message- when I do what you are doing in video- to run CI i do manually npm run format and lint:fix - they are not beuin executed automatically like in the video
image

akosasante:
Hmm, that's odd 🤔 

Do you have a folder in your repo named `.husky`? It's in `main` [branch](https://github.com/Code-the-Dream-School/dd-prac-team4-back/tree/main/.husky) so it should also be in your repo. Or it should have been automatically installed by the husky library when you've ran `npm install` in the past when the library was added to `package.json`. On Windows you can either check in the Windows Explorer UI (just make sure that 'show hidden folders' is [enabled](https://support.microsoft.com/en-us/windows/view-hidden-files-and-folders-in-windows-97fbc472-c603-9d90-91d0-1166d1d9f4b5)). Or in PowerShell terminal you can run the command `ls -Force`:

```
PS C:\Users\Cam\Documents\Coding\ctd\practicum2023\dd-prac-team4-back> ls -Force


    Directory: C:\Users\Cam\Documents\Coding\ctd\practicum2023\dd-prac-team4-back


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d--h--        2023-10-12   5:10 PM                .git
d-----        2023-10-12   5:10 PM                .github
d-----        2023-10-12   5:10 PM                .husky
d-----        2023-10-12   5:10 PM                images
d-----        2023-10-12   5:10 PM                logs
d-----        2023-10-12   5:10 PM                src
d-----        2023-10-12   5:10 PM                test
-a----        2023-10-12   5:10 PM           6148 .DS_Store
-a----        2023-10-12   5:10 PM            628 .eslintrc.json
-a----        2023-10-12   5:10 PM            226 .gitignore
-a----        2023-10-12   5:10 PM              7 .node-version
-a----        2023-10-12   5:10 PM            112 .prettierrc
-a----        2023-10-12   5:10 PM         306144 api-documentation.png
-a----        2023-10-12   5:10 PM            638 deleteCollectionMongo.js
-a----        2023-10-12   5:10 PM           2269 fetchAlbumSpotify.js
-a----        2023-10-12   5:10 PM           2888 orderCompleted.js
-a----        2023-10-12   5:10 PM         448289 package-lock.json
-a----        2023-10-12   5:10 PM           2250 package.json
-a----        2023-10-12   5:10 PM          12552 README.md
-a----        2023-10-12   5:10 PM          51486 swagger-autogen-success.png
-a----        2023-10-12   5:10 PM          61610 swagger-output.json
-a----        2023-10-12   5:10 PM           6577 swagger.js
-a----        2023-10-12   5:10 PM           1502 updateOrderToPaidScript.js
```

If it is there then you can check that it has the correct pre-commit file by opening `.\.husky\pre-commit` in notepad (either do this from Windows Explorer or you can run the command `notepad .\.husky\pre-commit` from your repo root in Powershell). 

The contents of the file should be:

```
#!/usr/bin/env sh
. "$(dirname -- "$0")/_/husky.sh"

npx lint-staged
```

Also, does your `node_modules` file have `husky` in there as a library and an executable? Maybe you haven't run `npm install` in a while and are missing these files?

```
PS C:\Users\Cam\Documents\Coding\ctd\practicum2023\dd-prac-team4-back> ls .\node_modules\husky*     


    Directory: C:\Users\Cam\Documents\Coding\ctd\practicum2023\dd-prac-team4-back\node_modules


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----        2023-10-13  11:45 AM                husky
```

and
```
PS C:\Users\Cam\Documents\Coding\ctd\practicum2023\dd-prac-team4-back> ls .\node_modules\.bin\husky*


    Directory: C:\Users\Cam\Documents\Coding\ctd\practicum2023\dd-prac-team4-back\node_modules\.bin


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----        2023-10-13  11:45 AM            318 husky
-a----        2023-10-13  11:45 AM            276 husky.cmd
-a----        2023-10-13  11:45 AM            488 husky.ps1
```

If you don't see those files; try running `npm install` and checking again. 
Now try making a random change to a JS file that you know will fail linting. Try to commit it and see if husky + lint-staged run now.

When I tried this on my Windows computer I got the following error:

```
PS C:\Users\Cam\Documents\Coding\ctd\practicum2023\dd-prac-team4-back> git add .\fetchAlbumSpotify.js
husky (debug) - starting pre-commit...
(node:23940) ExperimentalWarning: The ESM module loader is experimental.
file:///C:/Users/Cam/Documents/Coding/ctd/practicum2023/dd-prac-team4-back/node_modules/lint-staged/bin/lint-staged.js:21
const packageJson = JSON.parse(await fs.readFile(new URL('../package.json', import.meta.url)))
                               ^^^^^

SyntaxError: Unexpected reserved word
    at Loader.moduleStrategy (internal/modules/esm/translators.js:117:18)
    at async link (internal/modules/esm/module_job.js:42:21)
husky - pre-commit hook exited with code 1 (error)

```

After some [searching](https://github.com/conventional-changelog/commitlint/issues/2256) it seems that this is because lint-staged and its dependencies do not support node versions lower than 14. So make sure (`node -v`) that you're on a high enough version of node (you can redownload and reinstall the latest version from the [node website](https://nodejs.org/en/download) (generally it's best to download the latest Long Term Supported (LTS) version, which currently is v18; though you can also feel free to download any newer one); or if you're using a package manager you can install it your preferred way).

Let me know if it's still not working and sorry about just getting to this now; I know you had mentioned husky not working for you before 😅 . I just kept forgetting to test things out on my Windows computer to get a better idea of how to help.

NatachaKey:
@akosasante Akos, thank you so much for reviewing my work and questions, don't worry about the time- I will wait for you forever :)
I followed all the steps, my node version is 18 and I have husky istalled, and it allows me to commit without any problem-even  with a severe error in code - line here word ' hello' on line 26 
image

NatachaKey:
image

akosasante:
Hmm and the  `.\node_modules\.bin\husky.ps1` file exists there too? :thinking: 

What happens if you try to run `lint-staged` (the library that does the fixing part; husky just handles connecting to git pre-commit hooks) directly? `npx lint-staged` ?

NatachaKey:
> npx lint-staged

image
I get that there are no staged files

akosasante:
> image
> 
>  I get that there are no staged files

Does it say that even if you do have files that are changed and staged (`git add <file>` ) for a commit? 

Eg:

*   make a change in a file that you know should fail the linting step
*   `git add` that file as if you were going to commit it
*   run `npx lint-staged` and let me know what it says

NatachaKey:
> * npx lint-staged

image
 that works fine, thank you, Akos!! I thing the thing is that I've never run manually `npx lint-staged ` -till now I only ran manually `npm run lint:fix` and `npm run format` , I thought these checks run automatically when we commit

akosasante:
> > * npx lint-staged
> that works fine, thank you, Akos!! I thing the thing is that I've never run manually `npx lint-staged ` -till now I only ran manually `npm run lint:fix` and `npm run format` , I thought these checks run automatically when we commit

To clarify, this is me trying to debug your issue. Running `npx lint-staged` manually **should not** be needed. My goal was to verify whether there was something broken with that command/library, or with husky, or with how your `git` is setup. Husky, will try torun `npx lint-staged` as seen in this `pre-commit` file: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/.husky/pre-commit#L4

Next things we can try:
- Make sure Husky correctly set the git hooks path to its folder: `git config core.hooksPath` should return `.husky`. If it doesn't, then try setting it by running: `git config core.hooksPath .husky`
- Looking at the Husky repo it seems that on Windows sometimes the permissions aren't set properly, so we can try forcing that by doing `git update-index --add --chmod=+x .husky/pre-commit` . And then if it changed that file, you'll need to add+commit it.

NatachaKey:
> > > * npx lint-staged
> > >   that works fine, thank you, Akos!! I thing the thing is that I've never run manually `npx lint-staged ` -till now I only ran manually `npm run lint:fix` and `npm run format` , I thought these checks run automatically when we commit
> 
> To clarify, this is me trying to debug your issue. Running `npx lint-staged` manually **should not** be needed. My goal was to verify whether there was something broken with that command/library, or with husky, or with how your `git` is setup. Husky, will try torun `npx lint-staged` as seen in this `pre-commit` file: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/.husky/pre-commit#L4
> 
> Next things we can try:
> 
> * Make sure Husky correctly set the git hooks path to its folder: `git config core.hooksPath` should return `.husky`. If it doesn't, then try setting it by running: `git config core.hooksPath .husky`
> * Looking at the Husky repo it seems that on Windows sometimes the permissions aren't set properly, so we can try forcing that by doing `git update-index --add --chmod=+x .husky/pre-commit` . And then if it changed that file, you'll need to add+commit it.

it says  .husky/pre-commit: line 2: .husky/_/husky.sh: No such file or directory
image

PR #76, opened 2023-10-02
Update the register endpoint to send an email when successful

DYA13:
Hi Akos,I would like to ask for more details regarding transactions in Mongoose. I'm not sure whether I should implement it in the sendWelcomeEmail function or in the authController within the register function. Also, could you please explain how I can test all of this stuff?  Thank you very much!❤️

### DIFF
--- a/src/controllers/authController.js
+++ b/src/controllers/authController.js
@@ -55,6 +55,7 @@ const register = async (req, res) => {
       description: 'Bad request, missing email, name, password, or username'
     }
   */
+  emailSender.sendWelcomeEmail(user.email, user);
 };
 
 const login = async (req, res) => {
--- a/src/controllers/authController.js
+++ b/src/controllers/authController.js
@@ -184,7 +184,7 @@ const resetPassword = async (req, res) => {
   const { passwordToken, newPassword } = req.body;
 
   // Find the user with the provided password token
-  const [user] = await User.findOne({
+  const user = await User.findOne({
     passwordResetToken: passwordToken,
     // passwordResetExpiresOn: { $gt: new Date() }, // Check if reset token is still valid -option 1
   });
--- a/src/controllers/authController.js
+++ b/src/controllers/authController.js
@@ -28,10 +28,15 @@ const register = async (req, res) => {
     password,
     role,
   }); // Create a new user in the database
+
+  // Send the welcome email
+  await emailSender.sendWelcomeEmail(user.email, user);
+
   const tokenUser = createTokenUser(user); // Create a token based on user data
   attachCookiesToResponse({ res, user: tokenUser }); // Attach the token to cookies and send in the response
 
   res.status(StatusCodes.CREATED).json({ user: tokenUser }); // Send a successful response with user data
+
   /* 
     #swagger.summary = 'Register a new user'
     #swagger.parameters['user'] = {
@@ -55,7 +60,6 @@ const register = async (req, res) => {
       description: 'Bad request, missing email, name, password, or username'
     }
   */
-  emailSender.sendWelcomeEmail(user.email, user);
 };
 
 const login = async (req, res) => {
--- a/src/controllers/authController.js
+++ b/src/controllers/authController.js
@@ -53,8 +53,8 @@ const register = async (req, res) => {
   } finally {
     session.endSession();
   }
-};
-/* 
+
+  /* 
     #swagger.summary = 'Register a new user'
     #swagger.parameters['user'] = {
       in: 'body',
@@ -77,7 +77,7 @@ const register = async (req, res) => {
       description: 'Bad request, missing email, name, password, or username'
     }
   */
-
+};
 const login = async (req, res) => {
   const { email, password } = req.body;
 
--- a/src/controllers/authController.js
+++ b/src/controllers/authController.js
@@ -5,6 +5,7 @@ const CustomError = require('../errors');
 const { attachCookiesToResponse, createTokenUser } = require('../utils');
 const crypto = require('crypto');
 const emailSender = require('../mailing/sender');
+const mongoose = require('mongoose');
 
 const register = async (req, res) => {
   const { name, email, password, username } = req.body;
@@ -13,31 +14,47 @@ const register = async (req, res) => {
     throw new CustomError.BadRequestError('Please provide all required fields');
   }
 
-  const emailAlreadyExists = await User.findOne({ email }); // Check if a user with the email already exists
-  if (emailAlreadyExists) {
-    throw new CustomError.BadRequestError('Email already exists'); // If user exists, throw an error
+  const session = await mongoose.startSession();
+  session.startTransaction();
+
+  try {
+    const emailAlreadyExists = await User.findOne({ email }); // Check if a user with the email already exists
+
+    if (emailAlreadyExists) {
+      throw new CustomError.BadRequestError('Email already exists'); // If user exists, throw an error
+    }
+
+    const isFirstAccount = (await User.countDocuments({})) === 0; // Check if it's the first account
+    const role = isFirstAccount ? 'admin' : 'user'; // Assign a role based on first account or not
+
+    const user = await User.create(
+      [
+        {
+          name,
+          username,
+          email,
+          password,
+          role,
+        },
+      ],
+      { session }
+    ); // Create a new user in the database
+
+    // Send the welcome email
+    await emailSender.sendWelcomeEmail(user[0].email, user[0]);
+
+    const tokenUser = createTokenUser(user[0]); // Create a token based on user data
+    attachCookiesToResponse({ res, user: tokenUser }); // Attach the token to cookies and send in the response
+    await session.commitTransaction();
+    res.status(StatusCodes.CREATED).json({ user: tokenUser }); // Send a successful response with user data
+  } catch (error) {
+    await session.abortTransaction();
+    throw error;
+  } finally {
+    session.endSession();
   }
-
-  const isFirstAccount = (await User.countDocuments({})) === 0; // Check if it's the first account
-  const role = isFirstAccount ? 'admin' : 'user'; // Assign a role based on first account or not
-
-  const user = await User.create({
-    name,
-    username,
-    email,
-    password,
-    role,
-  }); // Create a new user in the database
-
-  // Send the welcome email
-  await emailSender.sendWelcomeEmail(user.email, user);
-
-  const tokenUser = createTokenUser(user); // Create a token based on user data
-  attachCookiesToResponse({ res, user: tokenUser }); // Attach the token to cookies and send in the response
-
-  res.status(StatusCodes.CREATED).json({ user: tokenUser }); // Send a successful response with user data
-
-  /* 
+};
+/* 
     #swagger.summary = 'Register a new user'
     #swagger.parameters['user'] = {
       in: 'body',
@@ -60,7 +77,6 @@ const register = async (req, res) => {
       description: 'Bad request, missing email, name, password, or username'
     }
   */
-};
 
 const login = async (req, res) => {
   const { email, password } = req.body;
--- a/src/controllers/authController.js
+++ b/src/controllers/authController.js
@@ -15,9 +15,9 @@ const register = async (req, res) => {
   }
 
   const session = await mongoose.startSession();
-  session.startTransaction();
 
   try {
+    session.startTransaction();
     const emailAlreadyExists = await User.findOne({ email }); // Check if a user with the email already exists
 
     if (emailAlreadyExists) {
--- a/src/controllers/authController.js
+++ b/src/controllers/authController.js
@@ -27,7 +27,7 @@ const register = async (req, res) => {
     const isFirstAccount = (await User.countDocuments({})) === 0; // Check if it's the first account
     const role = isFirstAccount ? 'admin' : 'user'; // Assign a role based on first account or not
 
-    const user = await User.create(
+    const [user] = await User.create(
       [
         {
           name,
@@ -41,9 +41,9 @@ const register = async (req, res) => {
     ); // Create a new user in the database
 
     // Send the welcome email
-    await emailSender.sendWelcomeEmail(user[0].email, user[0]);
+    await emailSender.sendWelcomeEmail(user.email, user);
 
-    const tokenUser = createTokenUser(user[0]); // Create a token based on user data
+    const tokenUser = createTokenUser(user); // Create a token based on user data
     attachCookiesToResponse({ res, user: tokenUser }); // Attach the token to cookies and send in the response
     await session.commitTransaction();
     res.status(StatusCodes.CREATED).json({ user: tokenUser }); // Send a successful response with user data
@@ -184,7 +184,7 @@ const resetPassword = async (req, res) => {
   const { passwordToken, newPassword } = req.body;
 
   // Find the user with the provided password token
-  const user = await User.findOne({
+  const [user] = await User.findOne({
     passwordResetToken: passwordToken,
     // passwordResetExpiresOn: { $gt: new Date() }, // Check if reset token is still valid -option 1
   });
--- a/src/mailing/sender.js
+++ b/src/mailing/sender.js
@@ -52,6 +52,29 @@ async function sendOrderCompletedEmail(to, username, orderItems, total) {
   });
 }
 
+// Function to send the welcome email
+async function sendWelcomeEmail(recipient, user) {
+  const template = 'welcome';
+  const subject = 'Welcome to Our Website';
+  const locals = { user };
+  await baseEmail
+    .send({
+      template,
+      message: {
+        to: recipient,
+        subject,
+      },
+      locals,
+    })
+    .then(() => {
+      console.log('Welcome email sent successfully.');
+    })
+    .catch((error) => {
+      console.error('Error sending welcome email:', error);
+      throw error;
+    });
+}
+
 async function sendForgotPasswordEmail(toEmail, resetToken) {
   try {
     await baseEmail.send({
@@ -71,4 +94,5 @@ module.exports = {
   sendTestEmail,
   sendForgotPasswordEmail,
   sendOrderCompletedEmail,
+  sendWelcomeEmail,
 };
--- a/src/mailing/sender.js
+++ b/src/mailing/sender.js
@@ -57,7 +57,7 @@ async function sendWelcomeEmail(recipient, user) {
   const template = 'welcome';
   const locals = { user };
   try {
-    await baseEmail.send({
+    const result await baseEmail.send({
       template,
       message: {
         to: recipient,
@@ -65,6 +65,7 @@ async function sendWelcomeEmail(recipient, user) {
       locals,
     });
     console.log('Welcome email sent successfully.');
+    return result
   } catch (error) {
     console.error('Error sending welcome email:', error);
     throw error;
--- a/src/mailing/sender.js
+++ b/src/mailing/sender.js
@@ -64,7 +64,6 @@ async function sendWelcomeEmail(recipient, user) {
       },
       locals,
     });
-    console.log(result);
 
     return result;
   } catch (error) {
--- a/src/mailing/sender.js
+++ b/src/mailing/sender.js
@@ -64,7 +64,8 @@ async function sendWelcomeEmail(recipient, user) {
       },
       locals,
     });
-    console.log('Welcome email sent successfully.');
+    console.log(result);
+
     return result;
   } catch (error) {
     console.error('Error sending welcome email:', error);
--- a/src/mailing/sender.js
+++ b/src/mailing/sender.js
@@ -55,14 +55,13 @@ async function sendOrderCompletedEmail(to, username, orderItems, total) {
 // Function to send the welcome email
 async function sendWelcomeEmail(recipient, user) {
   const template = 'welcome';
-  const subject = 'Welcome to Our Website';
+
   const locals = { user };
   await baseEmail
     .send({
       template,
       message: {
         to: recipient,
-        subject,
       },
       locals,
     })
--- a/src/mailing/sender.js
+++ b/src/mailing/sender.js
@@ -55,23 +55,20 @@ async function sendOrderCompletedEmail(to, username, orderItems, total) {
 // Function to send the welcome email
 async function sendWelcomeEmail(recipient, user) {
   const template = 'welcome';
-
   const locals = { user };
-  await baseEmail
-    .send({
+  try {
+    await baseEmail.send({
       template,
       message: {
         to: recipient,
       },
       locals,
-    })
-    .then(() => {
-      console.log('Welcome email sent successfully.');
-    })
-    .catch((error) => {
-      console.error('Error sending welcome email:', error);
-      throw error;
     });
+    console.log('Welcome email sent successfully.');
+  } catch (error) {
+    console.error('Error sending welcome email:', error);
+    throw error;
+  }
 }
 
 async function sendForgotPasswordEmail(toEmail, resetToken) {
--- a/src/mailing/sender.js
+++ b/src/mailing/sender.js
@@ -57,7 +57,7 @@ async function sendWelcomeEmail(recipient, user) {
   const template = 'welcome';
   const locals = { user };
   try {
-    const result await baseEmail.send({
+    const result = await baseEmail.send({
       template,
       message: {
         to: recipient,
@@ -65,7 +65,7 @@ async function sendWelcomeEmail(recipient, user) {
       locals,
     });
     console.log('Welcome email sent successfully.');
-    return result
+    return result;
   } catch (error) {
     console.error('Error sending welcome email:', error);
     throw error;
--- a/src/mailing/templates/welcome/subject.ejs
+++ b/src/mailing/templates/welcome/subject.ejs
@@ -1,1 +1,1 @@
-Welcome to Our Website
+Welcome to Our Website <%= user.name %>!
--- a/src/mailing/templates/welcome/subject.ejs
+++ b/src/mailing/templates/welcome/subject.ejs
@@ -1,0 +1,1 @@
+Welcome to Our Website
--- a/test/authController.test.js
+++ b/test/authController.test.js
@@ -4,6 +4,7 @@ const request = require('supertest');
 const { intervalId: orderUpdateInterval } = require('../src/models/Order');
 const User = require('../src/models/User');
 const { loginAndReturnCookie } = require('./test_helper');
+const sender = require('../src/mailing/sender');
 
 // Declare variables for the server, database connection, and in-memory MongoDB instance
 let server;
@@ -44,9 +45,13 @@ afterAll(async () => {
   // turn off the order update interval so that jest can cleanly shutdown
   clearInterval(orderUpdateInterval);
 });
+afterEach(() => {
+  jest.restoreAllMocks();
+});
 
 describe('Authentication API Endpoints', () => {
   it('should register a new user and log in', async () => {
+    const emailSpy = jest.spyOn(sender, 'sendWelcomeEmail');
     // Arrange: Register a new user
     const registrationResponse = await request(app)
       .post('/api/v1/auth/register')
@@ -65,6 +70,11 @@ describe('Authentication API Endpoints', () => {
       role: createdUser.role,
       userId: createdUser.id,
     });
+    expect(emailSpy).toHaveBeenCalledTimes(1); // Expecting the sendWelcomeEmail function to be called once
+    expect(emailSpy).toHaveBeenCalledWith(
+      testUser.email,
+      expect.objectContaining({ name: createdUser.name })
+    ); // Expecting the sendWelcomeEmail function to be called with the correct email address and a user object that has the expected name
 
     // Act: Log in with the newly registered user's credentials
     const loginResponse = await request(app)
--- a/test/authController.test.js
+++ b/test/authController.test.js
@@ -6,18 +6,15 @@ const User = require('../src/models/User');
 const { loginAndReturnCookie } = require('./test_helper');
 const sender = require('../src/mailing/sender');
 
-// Declare variables for the server, database connection, and in-memory MongoDB instance
 let server;
 let mongooseConnection;
 let mongodb;
 
-// Credentials for test user
 const testUserCredentials = {
   email: 'ava@ava.com',
   password: 'secret',
 };
 
-// Test user data
 const testUserData = {
   name: 'ava',
   username: 'ava',
@@ -29,13 +26,11 @@ console.log('Test User Data:', testUserData);
 console.log('Test User Credentials:', testUserCredentials);
 
 let testUser;
-// set up the mongodb and the express server before starting the tests
 beforeAll(async () => {
   mongodb = await MongoMemoryReplSet.create({
     replSet: { storageEngine: 'wiredTiger' },
   });
   const url = mongodb.getUri();
-  // set the url so that our server's mongoose connects to the in-memory mongodb and not our real one
   process.env.MONGO_URL = url;
   mongooseConnection = await connectDB(url);
   server = await app.listen(8001);
@@ -44,11 +39,9 @@ beforeEach(async () => {
   testUser = await User.create(testUserData);
 });
 afterAll(async () => {
-  // turn off the server and mongo connections once all the tests are done
   await server.close();
   await mongooseConnection.disconnect();
   await mongodb.stop();
-  // turn off the order update interval so that jest can cleanly shutdown
   clearInterval(orderUpdateInterval);
 });
 afterEach(async () => {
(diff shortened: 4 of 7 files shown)


src/mailing/templates/welcome/subject.ejs
akosasante, line 1: If you want, you can also use EJS variables here; if for example you want to personalize the subject as well with `Welcome to Our Website <%= user.name %>!` here

src/mailing/sender.js
```diff
@@ -52,6 +52,29 @@ async function sendOrderCompletedEmail(to, username, orderItems, total) {
   });
 }
 
+// Function to send the welcome email
+async function sendWelcomeEmail(recipient, user) {
+  const template = 'welcome';
+  const subject = 'Welcome to Our Website';
```
akosasante: I believe we can remove this line and line 65. The `email-templates` library will already automatically look for and use the `subject.ejs` file that's found in the template folder that you've declared (in this case it will look in `templates/welcome`.

src/mailing/sender.js
```diff
@@ -52,6 +52,29 @@ async function sendOrderCompletedEmail(to, username, orderItems, total) {
...
+      },
+      locals,
+    })
+    .then(() => {
+      console.log('Welcome email sent successfully.');
+    })
+    .catch((error) => {
+      console.error('Error sending welcome email:', error);
+      throw error;
+    });
```
akosasante: It's not necessarily bad; but it's a little uncommon to mix the `.then`/`.catch` syntax from Promise-based asynchronous code, with the `await + try/catch` syntax.

```javascript
# Promise syntax:

return baseEmail
  .send({
    template,
    message: { to: recipient },
    locals
  })
  .then((response) => {
    logger.info('Welcome email sent successfully.');
    // console.debug("GOT RESPONSE: ", response); // uncomment if we want to debug and see the return value
  })
  .catch((error) => {
    logger.error('Error sending welcome email: ', error);
    throw error;
  });
```

vs.

```javascript
try {
  const response = await baseEmail.send({
      template,
      message: { to: recipient },
      locals
  });
  logger.info('Welcome email sent successfully.');
    // console.debug("GOT RESPONSE: ", response); // uncomment if we want to debug and see the return value
} catch (error) {
  logger.error('Error sending welcome email: ', error);
  throw error;
});
```

akosasante: In this particular case, the code you've written will work totally well! 

I raise this comment because if in a different case you were wanting to get the return value of the resolved/awaited Promise, then it could lead to bugs. Eg:

```javascript
const response = await baseEmail.send(...).then(() => console.log("HI"));

console.debug("GOT RESPONSE: ", response);
```

^ In the above example, the debug log would print `GOT RESPONSE: undefined` even if the `baseEmail.send` actually happened successfully and returned something. That's because the return value of the `await` would get overwritten by the subsequent `then` that is chained there. And in this example, my `.then` isn't returning anything, it's just doing a `console.log`, and `console.log` by itself has a "void" return which means it will return `undefined`.

src/controllers/authController.js
```diff
@@ -55,6 +55,7 @@ const register = async (req, res) => {
       description: 'Bad request, missing email, name, password, or username'
     }
   */
+  emailSender.sendWelcomeEmail(user.email, user);
```
akosasante: To make it easier for someone unfamiliar with the code base to read this controller function, let's move this line up above the swagger comments.

akosasante: As well, we'll want it to happen _before_ we generate an HTTP response on line 34 with `res.status(...).json(...)`. Otherwise, right now, we'll return a 201 to the client before we've tried to send the email. And if the email fails to send, we throw an error but at that point, it's too late because we've already sent the HTTP response.

src/controllers/authController.js
akosasante, line 1: > I would like to ask for more details regarding transactions in Mongoose. I'm not sure whether I should implement it in the sendWelcomeEmail function or in the authController within the register function

Great question! Yes this is a good usecase for transactions. Currently, the other place where we use them in this app is in the `createOrder` controller function to wrap the creation of the order object, purchased albums objects, and calling stripe all under one transaction: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/67cb52b6279196f8fea68ba34dd27db8680ff0d8/src/controllers/orderController.js#L10](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/67cb52b6279196f8fea68ba34dd27db8680ff0d8/src/controllers/orderController.js#L10)

In this case, we would want to do the transaction in `authController` to wrap the creation of a User object, and the sending of an email all under one transaction. This ensures that if sending the email fails (_**by throwing an error**_) then the transaction will be _rolled back_ and the User object creation will be undone. 

---

Essentially we would want to follow a similar pattern as in the OrderController, referring to the Mongoose documentation here: https://mongoosejs.com/docs/transactions.html#:~:text=In%20practice%2C%20you,transaction%20error.

1.  Start a Mongo "session" (a special type of connection to the db where we can do things like start transactions) with mongoose's functionality: `const session = await mongoose.startSession();`
2.  Wrap all the existing code in this controller (except for the swagger docs) in a `try/catch`
3.  At the beginning of the `try {...}` block, start the transaction with `session.startTransaction();`
4.  After the user has been successfully created _**and**_ the email has been sucessfully sent and all that, then save/commit the transaction with `await session.commitTransaction();`
5.  Return the HTTP response with `res.status(...).json(...)` because at this point, we _**know**_ everything has happened successfully and been committed to the db
6.  In the `catch` we must call `await session.abortTransaction();` to tell mongoose/Mongo to undo _any_ creates/updates to the db that happened in this session's transaction. Then we can throw an error to let the `error-handler` middleware handle returning a response to the user, or we can just directly return a response to the user here with `res.status(...).json(...)`. 
7.  `try/catch` blocks also allow a third state, `finally { ... }` . This is where you put any code that you want to happen at the end of the block both in the success case and the error/catch case. In this case, what we want to put there is `session.endSession();` to close the special session connection that we had opened with the db. [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#:~:text=its%20surrounding%20parentheses.-,finallyStatements,statements%20execute%20regardless%20of%20whether%20an%20exception%20was%20thrown%20or%20caught.,-Description](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#:~:text=its%20surrounding%20parentheses.-,finallyStatements,statements%20execute%20regardless%20of%20whether%20an%20exception%20was%20thrown%20or%20caught.,-Description)

The reason that we _do_ want to wrap things in a `try/catch` here when we usually don't for these controller functions is because we have custom functionality that we want to happen with the session/transaction; rather than just letting the error-handler middleware simply handle any thrown errors by itself.

akosasante (commented):
> Also, could you please explain how I can test all of this stuff? Thank you very much!

Absolutely :smile: 

For the new `sendWelcomeEmail` we can add a new test case to the `sendEmail.test.js` file, following the same pattern as the existing `sendTestEmail` test block in there: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/0f1919578247451b82e91113aa9564acaab4d17c/test/sendEmail.test.js](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/0f1919578247451b82e91113aa9564acaab4d17c/test/sendEmail.test.js)

---

For the code in the controller, we can augment the existing registration test in `authController.test.js`. To test that an email is sent, since we can't test the return value of something that happens in a middle of a function, we'll linstead use a Jest "spy": [https://jestjs.io/docs/jest-object#jestspyonobject-methodname](https://jestjs.io/docs/jest-object#jestspyonobject-methodname) (more info on spies and mocks here: [https://spin.atomicobject.com/2021/06/28/mocks-stubs-or-spies/)](https://spin.atomicobject.com/2021/06/28/mocks-stubs-or-spies/))

First, we need to import the email module as an _object_ : `const` _**`sender`**_ `=` _**`require`**_`('../src/mailing/sender');`

Then, we need to setup an `afterEach` block to make sure we clean up the jest mocks after each test (good practice to do this even in this case where we are only using the mock/spy in a single test case anyway): 

```javascript
afterEach(() => {
  // restore the spy created with spyOn
  jest.restoreAllMocks();
});
```

  
Next we will set up the spy at the _beginning_ of the "should register a new user and log in" test block, before we make the http call: 

```javascript
// Arrange: spy on calls to the sendWelcomeEmail function
const emailSpy = jest.spyOn(sender, 'sendWelcomeEmail');
```

Lastly, when we do our assertions for the registration result, we can add some assertions that the email method was called and with the expected parameters:

```javascript
    expect(emailSpy).toHaveBeenCalledTimes(1); // Expecting the sendWelcomeEmail function to be called once
    expect(emailSpy).toHaveBeenCalledWith(testUser.email, expect.objectContaining({ name: createdUser.name })); // Expecting the sendWelcomeEmail function to be called with the correct email address and a user object that has the expected name
```

src/mailing/sender.js
DYA13: Thank you , removed subject 👍, commit b2f1a395bbe24ff6e67981b75d0f6cbc3f42595e

src/mailing/sender.js
DYA13: Thank you , I changed for await +try/catch block , commit 52d379713c431bff7bddb7d4c6f471fd591715ac

src/controllers/authController.js
DYA13: done , placed  before the generation of HTTP response , commit 1fead19fb685a576888961c62be709baa96fc11c

src/mailing/templates/welcome/subject.ejs
DYA13: I tried to make test as you explained but it needs your check, thank you

test/authController.test.js
akosasante, line 1: :tada:

test/sendEmail.test.js
```diff
@@ -23,4 +28,28 @@ describe('sendTestEmail function', () => {
     expect(result.envelope.from).toEqual(process.env.EMAIL_USERNAME); // Verify the 'from' address
     expect(result.envelope.to).toEqual([recipientEmail]); // Verify the 'to' address
   });
+
+  it('sends a welcome email with expected content', async () => {
+    let testAccount = await createTestAccount();
+
+    process.env.EMAIL_SERVICE = 'Ethereal';
+    process.env.EMAIL_USERNAME = testAccount.user;
+    process.env.EMAIL_PASSWORD = testAccount.pass;
```
akosasante: We can probably move this up top near the beginnign of the file (or at the beginning of the `description` block) that way we only need to do it once, and it will be the case for all the tests that run here.

:thinking:  Another thing we could do is have `beforeAll` and `afterAll` blocks, which might be good because that way we can reset it at the end just in case some other test across all our tests was expecting unchanged values (I don't think any of our other tests care right now, but it's always good to try to make sure your tests clean up after themselves to avoid future headaches)

```javascript
  let originalEmailService;
  let originalEmailUsername;
  let originalEmailPassword;
  let testAccount

  beforeAll(async () => {
    originalEmailService = process.env.EMAIL_SERVICE;
    originalEmailUsername = process.env.EMAIL_USERNAME;
    originalEmailPassword = process.env.EMAIL_PASSWORD;

    testAccount = await createTestAccount();
    
    process.env.EMAIL_SERVICE = 'Ethereal';
    process.env.EMAIL_USERNAME = testAccount.user;
    process.env.EMAIL_PASSWORD = testAccount.pass;
  });

  afterAll(() => {
    process.env.EMAIL_SERVICE = originalEmailService;
    process.env.EMAIL_USERNAME = originalEmailUsername;
    process.env.EMAIL_PASSWORD = originalEmailPassword;
  });
```

akosasante: Then we can remove the parts in this test and the other test in this file that are setting the environment variables to use the Ethereal test accounts since that happens already.

test/sendEmail.test.js
```diff
@@ -23,4 +28,28 @@ describe('sendTestEmail function', () => {
...
+    const mockUser = { name: 'John Doe' };
+    // Spy on the sendWelcomeEmail function to track if it's called
+    const sendWelcomeEmailSpy = jest.spyOn(sender, 'sendWelcomeEmail');
+
+    await sendWelcomeEmail(recipientEmail, mockUser);
+
+    expect(sendWelcomeEmailSpy).toHaveBeenCalledWith(recipientEmail, mockUser);
+
+    // Ensure there are no errors thrown during the email sending process
+    expect(sendWelcomeEmailSpy).not.toThrow();
```
akosasante: In the `authController.js` tests we needed to use a `spy` because there isn't really another way to access the email functionality that's happening in the middle of the auth code flow. 

But here, we're _directly_ testing the email flow; we are calling `sendWelcomeEmail` and can inspect the response of it right in this test. So we do not need to inject a jest spy to see what was passed into it our what it returned. For example, it doesn't provide us much info to know that `sendWelcomeEmail` was called with a recipient email and mock user, because we were the ones who did that on line 48 ourselves. Versus in `authController` we like to know that the code that gets run by our test is eventually doing the right thing by calling the email function correctly.

Instead, I would suggest modelling the test in this file, similarly to the `sends a test email with expected content` test above it. It should get a result from the `await sendWelcomeEmail` call, and it should use `expect` to assert that the email was (1) sent to the right person, (2) from the right person, (3) `originalMessage.text` in the result contains the `mockUser`'s name (you can use the [includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) function) .

akosasante: You may need to make a change in `sender.js` to make sure that the result of `await baseEmail.send(...)` is being stored in a variable and returned (after the console.log message) so that we can get that result here in the test and check it :eye:

test/sendEmail.test.js
```diff
@@ -1,5 +1,10 @@
 require('dotenv').config();
 const { createTestAccount } = require('nodemailer');
+const sender = require('../src/mailing/sender');
+
+afterEach(() => {
+  jest.restoreAllMocks();
+});
```
akosasante: See other comments here; we should not need to mock anything in this test file.

src/controllers/authController.js
DYA13: I tried to implement  as per your guide , here is my commit 52f117e042625700135aec062b22883912f3b0c1, but now jest tests fail in sendEmail.test.js

src/controllers/authController.js
```diff
@@ -13,31 +14,47 @@ const register = async (req, res) => {
...
-    password,
-    role,
-  }); // Create a new user in the database
+  const session = await mongoose.startSession();
+  session.startTransaction();
 
-  // Send the welcome email
-  await emailSender.sendWelcomeEmail(user.email, user);
+  try {
+    const emailAlreadyExists = await User.findOne({ email }); // Check if a user with the email already exists
```
akosasante: `session.startTransaction` should be inside our try block so that errors that happen within that transaction will be properly caught by the `try/catch` block: https://www.ultimateakash.com/blog-details/IiwzQGAKYAo=/How-to-implement-Transactions-in-Mongoose-&-Node.Js-(Express)

```js
  const session = await mongoose.startSession();

  try {
    session.startTransaction();
    const emailAlreadyExists = await User.findOne({ email }); // Check if a user with the email already exists
```

src/controllers/authController.js
```diff
@@ -13,31 +14,47 @@ const register = async (req, res) => {
...
+          role,
+        },
+      ],
+      { session }
+    ); // Create a new user in the database
+
+    // Send the welcome email
+    await emailSender.sendWelcomeEmail(user[0].email, user[0]);
+
+    const tokenUser = createTokenUser(user[0]); // Create a token based on user data
```
akosasante: Did you run into some issue that led you to change  the input of the `User.create` function from an object to an array? It seems like it makes things a bit more unclear because now `user` is a _list_ containing one user and we have to access it the user that we created by index `[0]` everywhere below.

akosasante: Oh never mind, I see now this message when I try it without the array:

```
(node:50922) [MONGOOSE] Warning: WARNING: to pass a `session` to `Model.create()` in Mongoose, you **must** pass an array as the first argument. See: https://mongoosejs.com/docs/api/model.html#Model.create()
```

TIL (today I learned) !

akosasante: In that case, I might suggest using array destructuring to assign the first value in the list returned (which is the user that we've just created) to the `user` variable like this: [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)

`const [user] = await User.create([{...}])`

akosasante:
@DYA13  ~I'm running into some quirks with Mongoose/Mongo transactions/Mongo in-mem server; I'll reply here once I figure out what's the best way forward. Sorry about that!~

[update] Okay nvm, I think whatever issue I was seeing is just something specific to my local machine. My review comments above should still stand 👍

test/sendEmail.test.js
DYA13: I have made some changes, but I messed a bit as it is new to me , could you check please?

test/sendEmail.test.js
DYA13: deleted the block with mock

src/controllers/authController.js
DYA13: done , commit ce459d315bd8980eecb2111ab0228f431716547e

src/controllers/authController.js
DYA13: distructuring done commit b9f612dfa6ee96f11f60b9556720699d2eebd226

test/sendEmail.test.js
DYA13: Akos I am a bit messed up now , according to this      `process.env.EMAIL_SERVICE = 'Ethereal'.` we use Ethereal to test credentials but in env email service I have Outlook. I tried to make jest test to see if we have connection Ethereal successful but  obviously it didn't work. Now I don't understand if we should use Outlook or Ethereal both or one of them?  jest test are not successful for api endpoints line 61 ,97 and 115 , could you please help to understand what I am making in the wrong way ?

test/sendEmail.test.js
```diff
@@ -9,7 +32,7 @@ describe('sendTestEmail function', () => {
     process.env.EMAIL_SERVICE = 'Ethereal';
     process.env.EMAIL_USERNAME = testAccount.user;
     process.env.EMAIL_PASSWORD = testAccount.pass;
-    // Import the sendTestEmail function from the sender.js file; this will now use the above env varas
+
```
akosasante, line 29: Lines 29-34 can be deleted since we are now doing this in the `beforeAll` block

test/sendEmail.test.js
```diff
@@ -1,5 +1,28 @@
 require('dotenv').config();
 const { createTestAccount } = require('nodemailer');
+const sender = require('../src/mailing/sender');
```
akosasante: This must be deleted. Not just for linting purposes, but it currently also makes the tests fail because this means that `sender.js` gets imported, and the code to create the `baseEmail` object in there runs first _before_ we have changed the `EMAIL_SERVICE`, `EMAIL_USERNAME`, and `EMAIL_PASSWORD` environment variables below.

test/sendEmail.test.js
```diff
@@ -20,7 +43,23 @@ describe('sendTestEmail function', () => {
...
+      'codethedream.practicum.team4+testrecipient@outlook.com';
+    const mockUser = {
+      name: 'Test User',
+    };
+
+    const result = await sendWelcomeEmail(recipientEmail, mockUser);
+
+    expect(result.envelope.from).toEqual(process.env.EMAIL_USERNAME);
+    expect(result.envelope.to).toEqual([recipientEmail]);
+    expect(result.message.text).toContain(mockUser.name);
```
akosasante: If we do a `console.log` (or `console.dir` since that formats objects a little nicer) of the `result` variable, we'll see that it looks like this:

```
    {
      envelope: {
        from: 'hmz2bs63d2tzxgvw@ethereal.email',
        to: [ 'codethedream.practicum.team4+testrecipient@outlook.com' ]
      },
      messageId: '<9a4c08d6-dc61-afc2-f7c0-426e42fc774a@ethereal.email>',
      message: '{"to":[{"address":"codethedream.practicum.team4+testrecipient@outlook.com","name":""}],"from":{"address":"hmz2bs63d2tzxgvw@ethereal.email","name":""},"attachments":[],"subject":"(TEST) Welcome to Our Website Test User!","html":"<!doctype html>\\n<html>\\n  <head>\\n    <title>Welcome to Our Website</title>\\n    <link rel=\\"stylesheet\\" type=\\"text/css\\" href=\\"style.css\\">\\n  </head>\\n  <body>\\n    <h1>Welcome to Our Website, Test User!</h1>\\n    <p>Thank you for registering with us.</p>\\n  </body>\\n</html>\\n","text":"WELCOME TO OUR WEBSITE, TEST USER!\\n\\nThank you for registering with us.","headers":{},"messageId":"<9a4c08d6-dc61-afc2-f7c0-426e42fc774a@ethereal.email>"}',
      originalMessage: {
        to: 'codethedream.practicum.team4+testrecipient@outlook.com',
        from: 'hmz2bs63d2tzxgvw@ethereal.email',
        attachments: [],
        subject: '(TEST) Welcome to Our Website Test User!',
        html: '<!doctype html>\n' +
          '<html>\n' +
          '  <head>\n' +
          '    <title>Welcome to Our Website</title>\n' +
          '    <link rel="stylesheet" type="text/css" href="style.css">\n' +
          '  </head>\n' +
          '  <body>\n' +
          '    <h1>Welcome to Our Website, Test User!</h1>\n' +
          '    <p>Thank you for registering with us.</p>\n' +
          '  </body>\n' +
          '</html>\n',
        text: 'WELCOME TO OUR WEBSITE, TEST USER!\n\nThank you for registering with us.'
      }
    }
```

It's easy to miss, but `message` is actually a json **string** (note the single quotation mark at the beginning and end), which represents the payload that gets sent to the email server. So we could convert that into a Javascript object and _then_ get the `text` field of that object by doing: `JSON.parse(result.message).text`.

Alternatively, we can use the result's `originalMessage` field which is already decoded into a javascript object.

In addition, we can see that anything that in our `html.ejs` file was a header (`<h1>Welcome to our Website</h1>`) is rendered in all capital letters in the plain text version of our email. So for our test to pass, we'll need to expect that as well:

```suggestion
    expect(result.originalMessage.text).toContain(mockUser.name.toUpperCase());
```

src/controllers/authController.js
```diff
@@ -163,7 +184,7 @@ const resetPassword = async (req, res) => {
   const { passwordToken, newPassword } = req.body;
 
   // Find the user with the provided password token
-  const user = await User.findOne({
+  const [user] = await User.findOne({
```
akosasante: I'm not sure I understand why this was changed to be wrapped in an array? `User.findOne` should be returning exactly one user, as an object.

test/sendEmail.test.js
akosasante: For tests, we don't want to send the email for real with Outlook. This would cause problems like (1) using up our email quotas, (2) risk of sending an email to a real custom from our testing/automated tests. So instead, we set manually reset the environment variable during our tests only, to use the [Ethereal email service](https://ethereal.email/) which is a fake email service specifically designed to be used for testing.

In development and in production, we use Outlook, and allow our server to use the environment variable set by dotenv, read from the `.env` file.

src/controllers/authController.js
akosasante: To clarify, this destructuring is only needed where we do `User.create([ {...}])` because we are passing in an _array_ to the `User.create` function which means that mongoose will return an _array_ of results. But we only need the first user in the array, so we destrucutre. 

But in the other place that you changed in that commit in the `login` function, this is not needed because `findOne` always returns just a single object and never an array.

test/authController.test.js
akosasante, line 1: One issue we have currently (this is from before your PR to be fair) is that the `logout` and `password reset` tests are _**dependent**_ on the `register` test.

Because the `logout` test tries to log out Ava; who is only created in the `register` test. So if `register` fails, then `logout` cannot succeed.

In order to create a better separation between the tests, I would suggest having a `beforeEach` block that creates (using `User.create` function) a user with Ava's info. Then `logout` and `password reset` can use that info successfuly even if the `register` test is failing.

As for `register` , of course the same user cannot register if they have already been created, so in that test we would need to delete Ava from the db at the beginning of the test before proceeding.

akosasante (commented):
I think I lost the message I had sent before but we should update `authController.test.js` to use a mongo server that supports transactions. This will be similar to the code change that Natalia is doing here: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/78#discussion_r1345808339](https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/78#discussion_r1345808339)

So our require at the top of this file changes to `const { MongoMemoryReplSet } = require('mongodb-memory-server');`

And in our `beforeAll` we would change that part to `mongodb = await MongoMemoryReplSet.create({ replSet: { storageEngine: 'wiredTiger' }});`

akosasante:
If I've added too much confusion with my previous comments, you can feel free to take a look at my debugging branch `aa/debug` specifically just the `authController.test.js` file there: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/aa/debug/test/authController.test.js

src/controllers/authController.js
DYA13: I removed destructing from resetPassword , commit f51cc7762ff368891056f8ddc6687842101129cb ,in this part of the code 
image

But in login function I didn't apply destructuring 🤔
image

test/sendEmail.test.js
DYA13: MongoMemoryReplSet added in authController.test.js , commit  dc4bc4f46e7994f26ccb00e61eb01d4a16a30c70
lines 29-34 deleted ,commit 53da59258d93532b895fd32c24dc827c0093082b

test/sendEmail.test.js
DYA13: Thank you 👍

DYA13:
Akos, thank you for your comments and explanation, there is a lot of new things for me and probably I messed up a bit 🫣

I still have the problem with the test 'should register a new user and log in'
 Authentication API Endpoints > should register a new user and log in

Error: expect(received).toBe(expected) // Object.is equality

Expected: 201
Received: 400

Actually when I register in Postman, I receive the email about successful registration immediately, but I don't see my registration in mongodb 

image

In Postman there is request without ending in result  

image

I suppose there is a problem in authentication, at the moment I haven't found the problem 🤔
By the way , registration works in main , I have checked with girls, it means definitely the problem in my branch

src/controllers/authController.js
```diff
@@ -13,26 +14,47 @@ const register = async (req, res) => {
...
+    await session.commitTransaction();
+    res.status(StatusCodes.CREATED).json({ user: tokenUser }); // Send a successful response with user data
+  } catch (error) {
+    await session.abortTransaction();
+    throw error;
+  } finally {
+    session.endSession();
+  }
+};
+/*
```
akosasante: The swagger code is now _outside_ of the function; so our swagger-autogen tool will no longer know which controller/endpoint it belongs to. Please move it up inside the function; at the end of the function (so between the brackets on line 55 and 56)

src/mailing/sender.js
```diff
@@ -52,6 +52,27 @@ async function sendOrderCompletedEmail(to, username, orderItems, total) {
...
+  const locals = { user };
+  try {
+    const result = await baseEmail.send({
+      template,
+      message: {
+        to: recipient,
+      },
+      locals,
+    });
+    console.log(result);
```
akosasante: We can probably remove this before merging

test/sendEmail.test.js
```diff
@@ -20,7 +37,39 @@ describe('sendTestEmail function', () => {
...
+  //       user: 'irma.cronin26@ethereal.email',
+  //       pass: 'mDHwkxB1yhxaEn3yCk',
+  //     },
+  //   });
+
+  //   const info = await transporter.verify();
+
+  //   expect(info).toBeDefined();
+  //   expect(info.accepted.length).toBeGreaterThan(0);
+  // });
```
akosasante: Cool test! We can keep or if it's not doing what you expect, I'd be fine with removing this before we merge.

src/controllers/authController.js
akosasante: Ah that's fine I probably just mixed up the two places when writing the comment; everything looks good now :+1:

test/authController.test.js
```diff
@@ -44,18 +51,27 @@ afterAll(async () => {
...
 });
+afterEach(async () => {
+  await User.deleteMany({});
+  jest.restoreAllMocks();
+}, 15000);
 
 describe('Authentication API Endpoints', () => {
+  beforeEach(async () => {
+    await User.create(testUser);
+  });
```
akosasante: Now we have **two** `beforeEach` blocks that are **both** trying to create the same test user, before each test. One on line 43-45 and one on line 60-62.

The one on line 43-45 is _outside_ any `describe` block, so runs _before every test in the file._  
The one on line 60-62 is _inside_ a `describe` block so it runs _before every test in that describe block._

You'll see in the test logs (if you run the test locally or if you check in the github run that right before the `POST /api/v1/auth/register` it logs the error "Email already exists": [https://github.com/Code-the-Dream-School/dd-prac-team4-back/actions/runs/6508303940/job/17677395210](https://github.com/Code-the-Dream-School/dd-prac-team4-back/actions/runs/6508303940/job/17677395210)

This is caused by trying to create a user with the same email twice. 

We can remove one of these `beforeEach` blocks; in this case it doesn't really matter which one we remove since we only have one `describe` block. But likely it's better to remove the one on line 60-62 so that if someone does eventually add more \`describe \`blocks to this test file, the test user will continue to be created for those separate `describe` blocks as well.

akosasante (commented):
One fix needs to be made in the authController test (also some small cleanup comments) and then you should be good to ship!

As for you comment about the register endpoint not working with Postman; what OS are you using (Mac, Windows, or Linux?). I ran into a similar issue when testing this on my Mac and it seemed to be due to the `email-templates` library trying to also open an ios simulator application whenever it previews an email. I've pushed a fix to that issue to `main`. Please try merging in `main` and see if you still have the issue. 

If you're on a different OS, please reach out to me on Slack and I will try to help debug with you later today or over the weekend. Thanks!

src/controllers/authController.js
DYA13: Swagger code has been moved inside the function register commit f0cf82f1f4446eaffe3143c03fc372f28d9b8648

src/mailing/sender.js
DYA13: console.log removed, commit 335d0c2b8c8a1d490beab585ecd607a7a65997a9

test/sendEmail.test.js
DYA13: I have deleted this test for now, anyway if we need later, we can add,  commit 299ecf3bf470778ca6c1c2880f1d126182f6e002

DYA13:
> issue

Akos, regarding the issue with register endpoint testing in Postman, yes I have Mac and I have IOS simulator application, I tried to register in Postman from main branch and it works perfectly , from my update-register-endpoint-email-successful branch it doesn't work. I suppose we have the same problem

test/authController.test.js
DYA13: I have removed the second beforeEach block but it doesn't work, it gives me the same error 
image

DYA13:
I also have this problem as well not only for socket but for test controllers as well🤔
image

akosasante:
> Akos, regarding the issue with register endpoint testing in Postman, yes I have Mac and I have IOS simulator application, I tried to register in Postman from main branch and it works perfectly , from my update-register-endpoint-email-successful branch it doesn't work. I suppose we have the same problem

Nice! Okay if the fix from `main` worked for you as well then you can try to merge that into this branch (be on this branch locally, and then run the command `git merge main` which will bring the latest changes from main in; or if you prefer you can click the "Update branch" button here in the PR



That way you can test out your changes locally without being blocked by this issue :smile:

akosasante:
> I also have this problem as well not only for socket but for test controllers as well🤔 
> 
> image

Oh that's odd; I wouldn't expect that `setInterval` to be called during tests because of the if block. Do you have the environment variable `START_ORDER_INTERVAL=true` in your `.env` file? If you do, could you try commenting out before running tests? You should need to have that value set.

test/authController.test.js
akosasante: Yep, keep in mind that when the user in the test tries to register, it's again going to try to make a user with that email when they call the `/register` endpoint: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/76#discussion_r1353487977](https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/76#discussion_r1353487977)

> As for `register` , of course the same user cannot register if they have already been created, so in that test we would need to delete Ava from the db at the beginning of the test before proceeding.

So at the beginning of the register test only, you'll need to delete the existing Ava user: `await User.deleteMany({})` so that they can register via the endpoint successfully.

DYA13: Ok now I understood, thank you 👍

DYA13:
> > I also have this problem as well not only for socket but for test controllers as well🤔 
> > image
> 
> Oh that's odd; I wouldn't expect that `setInterval` to be called during tests because of the if block. Do you have the environment variable `START_ORDER_INTERVAL=true` in your `.env` file? If you do, could you try commenting out before running tests? You should need to have that value set.

Yes I have this line `START_ORDER_INTERVAL=true` in env and if I comment out, I don't have this problem with setInterval. Should I leave it commented out or uncomment anyway? Actually it doesn't give me the problems in terminal, only if running test and debugger

akosasante:
> Yes I have this line `START_ORDER_INTERVAL=true` in env and if I comment out, I don't have this problem with setInterval. Should I leave it commented out or uncomment anyway? Actually it doesn't give me the problems in terminal, only if running test and debugger

Yes for now I would suggest leaving that variable commented out in your `.env` file.

The if statement is checking the condition such that it will initiate running the update order on recurring intervals only IF:

1. the server is currently running in any mode **except** for test mode

OR

2. the environment variable is enabled.

---

The idea here is that we **don't** want setInterval running in test mode because it could interfere with code we are trying to teat but updating orders in the middle of a test, _and_ it adda an extra cleanup step to all our test files where we need to shut down the setInterval recurrence so that Jest can cleanly close the test block.

So by having this conditional, we tell the app to not start that interval in test mode. In development mode and production mode, everything should continue to work as normal, with setInterval running.

But we wanted to provide an override in case for a particular test someone _did_ want that setInterval running (for example to test that it updates the orders on a schedule as expected).

I'll take a look at if theres an approach we can take to make this more straightforward. But for now, I wouldn't expect that anyone needs to have that env variable enabled at all.

akosasante (approved):
Ready to merge :rocket:

PR #77, opened 2023-10-03
PS188-return-messages-in-mongo

mvstoyan:
in progress...

### DIFF
--- a/src/live/handleAlbumChat.js
+++ b/src/live/handleAlbumChat.js
@@ -2,22 +2,16 @@ const ChatMessage = require('../models/ChatMessage');
 
 async function handleAlbumChat(io, socket, data) {
   try {
+    // console.log("data sent by frontend: ", data);
     const { albumId } = data;
 
     // Create a chat room name for the album chat
     const chatRoomName = `chat:album:${albumId}`;
 
-    const msg = await ChatMessage.create({ ...data, name: data.user });
+    const msg = await ChatMessage.create(data);
     await msg.populate('user', 'name');
     console.log('Chat message saved:', msg);
 
-    const savedMessage = {
-      ...data,
-      _id: msg._id, // Include _id in the message
-      name: data.name,
-    };
-
-    // Check and delete old messages if needed
     const messagesCount = await ChatMessage.countDocuments({ albumId });
     if (messagesCount > 50) {
       await ChatMessage.findOneAndDelete(
@@ -27,7 +21,7 @@ async function handleAlbumChat(io, socket, data) {
     }
 
     // Broadcast the message to all connected users in the album chat room
-    io.to(chatRoomName).emit('chat:album', savedMessage);
+    io.to(chatRoomName).emit('chat:album', msg);
   } catch (error) {
     console.error('Error handling chat message:', error);
   }
--- a/src/live/handleAlbumChat.js
+++ b/src/live/handleAlbumChat.js
@@ -7,16 +7,9 @@ async function handleAlbumChat(io, socket, data) {
     // Create a chat room name for the album chat
     const chatRoomName = `chat:album:${albumId}`;
 
-    const msg = await ChatMessage.create({ ...data, username: data.user });
-    await msg.populate('user', 'username');
+    const msg = await ChatMessage.create(data);
+    await msg.populate('user', 'name');
     console.log('Chat message saved:', msg);
-
-    // Return the _id of the saved message
-    const savedMessage = {
-      ...data,
-      _id: msg._id, // Include _id in the message
-      username: data.username,
-    };
 
     // Check and delete old messages if needed
     const messagesCount = await ChatMessage.countDocuments({ albumId });
@@ -28,7 +21,7 @@ async function handleAlbumChat(io, socket, data) {
     }
 
     // Broadcast the message to all connected users in the album chat room
-    io.to(chatRoomName).emit('chat:album', savedMessage); 
+    io.to(chatRoomName).emit('chat:album', data);
   } catch (error) {
     console.error('Error handling chat message:', error);
   }
--- a/src/live/handleAlbumChat.js
+++ b/src/live/handleAlbumChat.js
@@ -7,9 +7,15 @@ async function handleAlbumChat(io, socket, data) {
     // Create a chat room name for the album chat
     const chatRoomName = `chat:album:${albumId}`;
 
-    const msg = await ChatMessage.create(data);
+    const msg = await ChatMessage.create({ ...data, name: data.user });
     await msg.populate('user', 'name');
     console.log('Chat message saved:', msg);
+
+    const savedMessage = {
+      ...data,
+      _id: msg._id, // Include _id in the message
+      name: data.name,
+    };
 
     // Check and delete old messages if needed
     const messagesCount = await ChatMessage.countDocuments({ albumId });
@@ -21,7 +27,7 @@ async function handleAlbumChat(io, socket, data) {
     }
 
     // Broadcast the message to all connected users in the album chat room
-    io.to(chatRoomName).emit('chat:album', data);
+    io.to(chatRoomName).emit('chat:album', savedMessage);
   } catch (error) {
     console.error('Error handling chat message:', error);
   }
--- a/src/live/handleAlbumChat.js
+++ b/src/live/handleAlbumChat.js
@@ -7,9 +7,16 @@ async function handleAlbumChat(io, socket, data) {
     // Create a chat room name for the album chat
     const chatRoomName = `chat:album:${albumId}`;
 
-    const msg = await ChatMessage.create({ ...data, user: data.user });
-    await msg.populate('user', 'name');
-    console.log('Chat message saved:', data);
+    const msg = await ChatMessage.create({ ...data, username: data.user });
+    await msg.populate('user', 'username');
+    console.log('Chat message saved:', msg);
+
+    // Return the _id of the saved message
+    const savedMessage = {
+      ...data,
+      _id: msg._id, // Include _id in the message
+      username: data.username,
+    };
 
     // Check and delete old messages if needed
     const messagesCount = await ChatMessage.countDocuments({ albumId });
@@ -21,7 +28,7 @@ async function handleAlbumChat(io, socket, data) {
     }
 
     // Broadcast the message to all connected users in the album chat room
-    io.to(chatRoomName).emit('chat:album', data);
+    io.to(chatRoomName).emit('chat:album', savedMessage); 
   } catch (error) {
     console.error('Error handling chat message:', error);
   }
--- a/src/live/handleAlbumChat.js
+++ b/src/live/handleAlbumChat.js
@@ -1,11 +1,30 @@
-function handleAlbumChat(io, socket, data) {
-  const { message, userId, albumId } = data;
+const ChatMessage = require('../models/ChatMessage');
 
-  // Create a chat room name for the album chat
-  const chatRoomName = `chat:album:${albumId}`;
+async function handleAlbumChat(io, socket, data) {
+  try {
+    const { albumId } = data;
 
-  // Send the message to the room
-  socket.to(chatRoomName).emit('chat:album', { message, userId, albumId });
+    // Create a chat room name for the album chat
+    const chatRoomName = `chat:album:${albumId}`;
+
+    const msg = await ChatMessage.create({ ...data, user: data.user });
+    await msg.populate('user', 'name');
+    console.log('Chat message saved:', data);
+
+    // Check and delete old messages if needed
+    const messagesCount = await ChatMessage.countDocuments({ albumId });
+    if (messagesCount > 50) {
+      await ChatMessage.findOneAndDelete(
+        { albumId },
+        { sort: { timestamp: 1 } }
+      );
+    }
+
+    // Broadcast the message to all connected users in the album chat room
+    io.to(chatRoomName).emit('chat:album', data);
+  } catch (error) {
+    console.error('Error handling chat message:', error);
+  }
 }
 
 module.exports = { handleAlbumChat };
--- a/src/live/index.js
+++ b/src/live/index.js
@@ -1,8 +1,6 @@
 const { testPing } = require('./testHandlers');
 const { handleUserNotificationsJoin } = require('./notificationHandlers');
-const ChatMessage = require('../models/ChatMessage');
 const { handleAlbumChat } = require('./handleAlbumChat');
-const User = require('../models/User');
 
 const onConnect = (io, socket) => {
   console.log('a user connected');
@@ -10,9 +8,9 @@ const onConnect = (io, socket) => {
   /* Add event handlers below */
   socket.on('test:ping', testPing);
 
-  socket.on('join:user_notifications', (userId) => {
-    handleUserNotificationsJoin(io, socket, userId);
-    console.log(`User joined notifications room for user: ${userId}`);
+  socket.on('join:user_notifications', (user) => {
+    handleUserNotificationsJoin(io, socket, user);
+    console.log(`User joined notifications room for user: ${user}`);
   });
 
   socket.on('join:album_chat', (albumId) => {
@@ -24,45 +22,13 @@ const onConnect = (io, socket) => {
     //console.log(`User joined album chat room for album: ${albumId}`);
   });
 
-  // Event handler for 'chat:album' that performs the following actions:
-  // - Creates a new message and saves it to the database
-  // - Checks the message count and deletes old messages if it exceeds 50
-  // - Broadcasts the message to all connected users in the album chat room
-  // - Calls the handleAlbumChat function for additional message processing in the album chat room
   socket.on('chat:album', async (data) => {
     try {
-      if (!data.userId || !data.message || !data.albumId) {
-        console.error('Invalid data format:', data);
-        return;
-      }
-      await ChatMessage.create(data);
-      console.log('Chat message saved:', data);
-      const messagesCount = await ChatMessage.countDocuments({
-        albumId: data.albumId,
-      });
-      if (messagesCount > 50) {
-        await ChatMessage.findOneAndDelete(
-          { albumId: data.albumId },
-          { sort: { timestamp: 1 } }
-        );
-      }
-
-      // user name
-      const user = await User.findById(data.userId);
-      if (!user) {
-        console.error('User not found:', data.userId);
-        return;
-      }
-      data.userName = user.name;
-
-      io.to(`chat:album:${data.albumId}`).emit('chat:album', data);
-
       handleAlbumChat(io, socket, data);
     } catch (error) {
       console.error('Error saving chat message:', error);
     }
   });
-
   /* End event handlers */
 
   socket.on('disconnect', () => {
--- a/src/live/index.js
+++ b/src/live/index.js
@@ -2,15 +2,17 @@ const { testPing } = require('./testHandlers');
 const { handleUserNotificationsJoin } = require('./notificationHandlers');
 const ChatMessage = require('../models/ChatMessage');
 const { handleAlbumChat } = require('./handleAlbumChat');
+const User = require('../models/User');
 
 const onConnect = (io, socket) => {
   console.log('a user connected');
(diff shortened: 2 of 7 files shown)


mvstoyan:
In Postman, the request to the backend at http://localhost:8000/api/v1/chat/6r1lh7fHMB499vGKtIyJLy is executed successfully. However, I receive an empty array as the response: {"messages": []}. On the frontend, I'm still encountering a CORS issue.

src/routes/chatRoutes.js
```diff
@@ -0,0 +1,20 @@
...
+    const messages = await ChatMessage.find({ albumId })
+      .sort({ timestamp: -1 }) // Sort messages in reverse chronological order
+      .limit(50) // Limit 50 messages
+      .exec(); // Execute the database query and retrieve the results
+
+    res.json({ messages });
+  } catch (error) {
+    console.error('Error fetching chat messages:', error);
+    res.status(500).json({ error: 'Internal server error' });
+  }
```
akosasante: Small nitpick; we can simplify here by not wrapping the code in a `try/catch`, since our `catch` isn't adding much special logic that the `error-handler` middleware wouldn't already do for uncaught errors, apart from an additional log line.

src/models/ChatMessage.js
```diff
@@ -0,0 +1,31 @@
+const mongoose = require('mongoose');
+
+const chatMessageSchema = new mongoose.Schema(
+  {
+    album: {
+      type: mongoose.Schema.ObjectId,
```
akosasante: Right now, we're using the album's spotify API id, rather than the mongoose object id.

I see a couple options here; (1) we can make sure our other code (both in the frontend and backend) is using the album's object id when creating a chat room and messages or we could (2) make this field just `album_spotify_id` or `chat_room_id` or something like that and designate it as a `String` type and then have a separate 'virtual' getter/setter: [https://mongoosejs.com/docs/tutorials/virtuals.html#virtual-setters](https://mongoosejs.com/docs/tutorials/virtuals.html#virtual-setters) (similar to what we do to generate the `purchasedByUsers` field in `Album.js`). 

Let me know if you'd like me to explain either or both options more in-depth

src/models/ChatMessage.js
```diff
@@ -0,0 +1,31 @@
...
+      ref: 'User',
+      required: true,
+    },
+    timestamp: {
+      type: Date,
+      default: Date.now,
+    },
+  },
+  {
+    capped: { size: 102400, max: 50 }, // max 50 messages
```
akosasante: It looks like we're checking for the count of messages in a single chat room in the `live/index.js` file, so we can either completely remove this overall collection cap, or at least make it larger (say 1000 messages) to allow for multiple rooms to have 50 messages stored in the db at the same time.

src/live/index.js
```diff
@@ -14,6 +16,30 @@ const onConnect = (io, socket) => {
     const chatRoomName = `chat:album:${albumId}`;
     socket.join(chatRoomName);
   });
+
+  socket.on('chat:album', async (data) => {
+    try {
+      // Create and save a new message to the database
+      await ChatMessage.create(data);
```
akosasante: Currently, for me this is returning some validation errors: 

```
Error saving chat message: Error: ChatMessage validation failed: sender: Path `sender` is required., text: Path `text` is required., album: Path `album` is required.
```

We should check to see if the `data` we're getting from the frontend matches the shape that the `ChatMessage` schema expects or if there's some transformation that we need to do here in this socket handler first before trying to call `create`.

akosasante (commented):
Regarding the CORS issues that you mentioned,

Are you getting those for the websocket connection, or when trying to do a `fetch` or `axios` call to the `/api/v1/chat` endpoint from the frontend?

If you're getting them from the websocket connection, check which backend branch you're testing against, does it have the CORS changes that you merged to `main` in there?

If you're getting them from calling the `/api/v1/chat` endpoint, make sure that the frontend is using the `axiosInstance` object from the `src/apis/axiosClient.js` file because that will have the CORS headers, cookies and all that already set up for you: https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/c280585383d7d4aa4fe0f28f12e6d2b55c0b83bc/src/components/Purchase/CheckoutComplete.js#L36

src/routes/chatRoutes.js
mvstoyan: router

src/live/index.js
```diff
@@ -18,14 +21,48 @@ const onConnect = (io, socket) => {
...
+      const user = await User.findById(data.userId);
+      if (!user) {
+        console.error('User not found:', data.userId);
+        return;
+      }
+      data.userName = user.name;
+
+      io.to(`chat:album:${data.albumId}`).emit('chat:album', data);
+
+      handleAlbumChat(io, socket, data);
```
akosasante: Currently this code is sending 2 messages to the socket. One from line 58 here `io.to(...)` and then again here where we call `handleAlbumChat` where there is also a `io.to(...)` message sent.

Instead, we should move all this code to `handleAlbumChat`. Remember, that file is kind of like our "controller", and this file is like our "router". This file shouldn't have too much logic inside if it apart from saying "events for this channel should be handled by that function"

So we should end up with:

```javascript
// here in live/index.js
socket.on('chat:album', handleAlbumChat);

// in handleAlbumChat.js
async function handleAlbumChat(io, socket, data) {
	try {
	/* ... all the code you've written here should replace what's currently in that file */
	} catch (e) {
	// ...
	}
}
```

src/live/index.js
```diff
@@ -18,14 +21,48 @@ const onConnect = (io, socket) => {
...
+      // user name
+      const user = await User.findById(data.userId);
+      if (!user) {
+        console.error('User not found:', data.userId);
+        return;
+      }
+      data.userName = user.name;
+
+      io.to(`chat:album:${data.albumId}`).emit('chat:album', data);
+
```
akosasante: Rather than editing the data and sending that back; it may save us some trouble on the frontend (see my comment in your frontend PR) to send back the actual `ChatMessage` object that we've just created on line 38. I'll leave some comments in `ChatMessage.js` about how we can also get the user name from that. Instead of needing to import `User` model here and do a second query to the db to fetch the user info on lines 51-55 for just one field of data.

src/models/ChatMessage.js
```diff
@@ -0,0 +1,44 @@
...
+  {
+    albumId: {
+      type: String,
+      required: true,
+    },
+    message: {
+      type: String,
+      required: true,
+    },
+    userId: {
```
akosasante: Let's call this field `user` . Usually we call `ref` fields by the model we're referring to, rather than the `id` even though we can save it with just the id. See for example the `ref` properties that we have on the `Order` and `OrderItem` models: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/ef73cbab610ba7d19ae351e6366edd3082a0ecb2/src/models/Order.js#L11](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/ef73cbab610ba7d19ae351e6366edd3082a0ecb2/src/models/Order.js#L11)

```suggestion
    user: {
```

It also makes it a bit clearer then that we can use the [populate](https://mongoosejs.com/docs/populate.html) method in the album chat socket handler to get the user name so that we can return that data in the socket.

```javascript
// in `handleAlbumChat.js`

		const msg = await ChatMessage.create({ ...data, user: data.userId });
		await msg.populate('user', 'name'); // this populates the "user" property and because of the `ref` property on the model, mongoose knows  that it should lookup this id in the User schema. The second argument 'name', tells it to return _only_  that field.
```

src/models/ChatMessage.js
```diff
@@ -0,0 +1,44 @@
...
+
+// Creating a virtual field "album" that references the Album model
+chatMessageSchema.virtual('album', {
+  ref: 'Album', // Reference to the Album model
+  localField: 'albumId', // Local field for the relationship
+  foreignField: 'spotify_id', // Foreign field for the relationship
+  justOne: true, // Refers to a single album
+});
+
+//console.log('Virtual Field "album" Created');
```
akosasante: Honestly, I think I somehow totally confused myself when I gave the suggestion to add this virtual before, I'm so sorry :sweat: :sweat\_smile: . We can remove this code for now.

So far in our code, we don't actually care about fetching a full Album object from a ChatMessage. The only thing we need in our `chatRoutes` is the ability to search by the value that we got from the API url, which we can do without any additional modification. 

If there's a use case in the future where we need the full album from a `ChatMessage` object, then we can revisit this. Apologies for any confusion.

src/models/ChatMessage.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+);
+
+//console.log('Schema Created');
+
+// Creating a virtual field "album" that references the Album model
+chatMessageSchema.virtual('album', {
+  ref: 'Album', // Reference to the Album model
+  localField: 'albumId', // Local field for the relationship
+  foreignField: 'spotify_id', // Foreign field for the relationship
```
akosasante: Part of the reason that I want to remove this for now, is that there is no `spotify_id` on the Album schema. We _could_ add one and do a backfill to update all the existing documents with a valid value there but I don't think it's worth the effort right now since we don't have a use case fo rit.

akosasante (commented):
Sorry for a bit of delay in reviews this week!

src/live/index.js
mvstoyan: screen

src/live/index.js
```diff
@@ -18,13 +19,15 @@ const onConnect = (io, socket) => {
...
+    //console.log(`User joined album chat room for album: ${albumId}`);
   });
 
-  socket.on('chat:album', (data) => {
-    console.log(
-      `received message from socket: ${socket.id}. Message=${data.message} | Album=${data.albumId} | User=${data.userId}`
-    );
-    handleAlbumChat(io, socket, data);
+  socket.on('chat:album', async (data) => {
+    try {
```
akosasante: Since we already have a try/catch inside `handleAlbumChat` function that is catching any errors that occur and logging them; the additional try/catch here is redundant and can be removed.

src/live/handleAlbumChat.js
```diff
@@ -1,11 +1,37 @@
...
+async function handleAlbumChat(io, socket, data) {
+  try {
+    const { albumId } = data;
 
-  // Send the message to the room
-  socket.to(chatRoomName).emit('chat:album', { message, userId, albumId });
+    // Create a chat room name for the album chat
+    const chatRoomName = `chat:album:${albumId}`;
+
+    const msg = await ChatMessage.create({ ...data, username: data.user });
```
akosasante: `data` is coming from the browser and looks like:

```json
{
    "message": "hi",
    "albumId": "6r1lh7fHMB499vGKtIyJLy",
    "user": "64ef5e90086e1c19b888c72c",
    "isOwnMessage": true
}
```

You can see here that `isOwnMessage` will be ignored by mongo because it is not part of the `ChatMessage` schema. And `username` will also be ignored because it is not part of the schema.

The `ChatMessage` schema expects three fields to be passed in: `message` (a string), `albumId` (a string), and `user` (a Mongo ObjectID for a particular User model). 

In this case, we don't care if `isOwnMessage` is ignored (see my review comments in your frontend repo). And all the other fields that the frontend is passing in already have the correct field names that match the `ChatMessage` schema, so we can pass the `data` object directly to create a new `ChatMessage`:  
 

```suggestion
    const msg = await ChatMessage.create(data);
```

src/live/handleAlbumChat.js
```diff
@@ -1,11 +1,37 @@
...
+    await msg.populate('user', 'username');
+    console.log('Chat message saved:', msg);
```
akosasante: Nice! You should see that when this is logged you get something like:

```javascript
Chat message saved: {
  albumId: '6r1lh7fHMB499vGKtIyJLy',
  message: 'hi',
  user: { _id: new ObjectId("64ef5e90086e1c19b888c72c"), username: 'akos' },
  _id: new ObjectId("6525be96a91bfb376e6e86d8"),
  timestamp: 2023-10-10T21:13:58.157Z,
  __v: 0
}
```

where `msg.user` is an object, that has a `msg.user.username` field that is the username of the person who sent this message :tada:

src/live/handleAlbumChat.js
```diff
@@ -1,11 +1,37 @@
...
+
+    // Return the _id of the saved message
+    const savedMessage = {
+      ...data,
+      _id: msg._id, // Include _id in the message
+      username: data.username,
```
akosasante: `data` does not have a `username` field based on what the frontend is currently sending us

src/live/handleAlbumChat.js
```diff
@@ -1,11 +1,37 @@
...
+    };
```
akosasante: I think instead of creating a variable that combines the incoming `data` and the username from `msg.user.username`, we can instead emit `msg` as the payload on line 31 directly. That should have all the information that the frontend needs to render the messages.

mvstoyan: name 
Displays in terminal 'name'.

mvstoyan: > name 

Displays in terminal 'name' only, the 'username' does not work.

akosasante: Yes because on line 11 you've populated only `'name'` . If you want username to be populated onto the `ChatMessage.user` object from the `User` model, that line should be `await msg.populate('user', 'username')` which I think you did have there before this commit: https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/77/commits/48203fe795647116eb10f45cfe8b94ca1aca7b30

Or if you want both of those fields you can pass in an array: `await msg.populate('user', ['name', 'username'])`. Or if you really really need the _full_ user object, then you can omit the second argument: `await msg.populate('user')`. But generally it's best (especially over websocket which we want to keep speedy) to just send the fields that the client needs.

akosasante: Ah I see in the frontend you're using `user.name` now so I think you intentionally changed this to 'name' since the last review; all good then as long as frontend and backend match in what they're expecting :+1: https://github.com/Code-the-Dream-School/dd-prac-team4-front/pull/52/files#diff-dc044c6ce14501135532d751cb2c59b15079b67646f1634abe6dd9c739c899b4R125

src/live/handleAlbumChat.js
```diff
@@ -1,11 +1,30 @@
...
+    const messagesCount = await ChatMessage.countDocuments({ albumId });
+    if (messagesCount > 50) {
+      await ChatMessage.findOneAndDelete(
+        { albumId },
+        { sort: { timestamp: 1 } }
+      );
+    }
+
+    // Broadcast the message to all connected users in the album chat room
+    io.to(chatRoomName).emit('chat:album', data);
```
akosasante: You're sending back the original `data` object that the frontend sent:

```

{
  "message": "Hi",
  "albumId": "6r1lh7fHMB499vGKtIyJLy",
  "user": "65297a45ccf10113ea16cf7e"
}
```

This won't have the populated user name or any other info that we could get from the `ChatMessage` that we created on line 10. It looks like you changed this from returning the `msg` variable to returning `data` in the last commit: 48203fe

Could you help me understand if there was something missing from `msg` that led you to change that?

src/live/handleAlbumChat.js
mvstoyan: its work with "name". Frontend: <strong>{message.user.name}</strong>: {message.message}
name
but the 'username' does not work. Frontend: <strong>{message.user.username}</strong>: {message.message}
username

mvstoyan: > I think instead of creating a variable that combines the incoming `data` and the username from `msg.user.username`, we can instead emit `msg` as the payload on line 31 directly. That should have all the information that the frontend needs to render the messages.
I probably didn’t understand correctly earlier, I thought I needed to make these changes.

src/live/handleAlbumChat.js
mvstoyan: I probably didn’t understand you correctly 3 days ago, I thought I needed to make these changes, I’m returning it back.

src/live/handleAlbumChat.js
akosasante: It seems that particular user does not have a username in the database; so it makes sense that nothing was returned when that field was populated. We added `username` as a required field somewhere in the middle of the project, so users that were created before that may not have a user name unless we go back and fix them. :upside\_down\_face:  Sorry that caused some confusion!

src/live/handleAlbumChat.js
```diff
@@ -1,11 +1,36 @@
...
+async function handleAlbumChat(io, socket, data) {
+  try {
+    const { albumId } = data;
 
-  // Send the message to the room
-  socket.to(chatRoomName).emit('chat:album', { message, userId, albumId });
+    // Create a chat room name for the album chat
+    const chatRoomName = `chat:album:${albumId}`;
+
+    const msg = await ChatMessage.create({ ...data, name: data.user });
```
akosasante: `data.user` that the frontend is passing us is a user id as far as I see in the frontend code: [https://github.com/Code-the-Dream-School/dd-prac-team4-front/pull/52/files#diff-dc044c6ce14501135532d751cb2c59b15079b67646f1634abe6dd9c739c899b4R26](https://github.com/Code-the-Dream-School/dd-prac-team4-front/pull/52/files#diff-dc044c6ce14501135532d751cb2c59b15079b67646f1634abe6dd9c739c899b4R26)

Also the `ChatMessage` schema does not have a `name` attribute so this field will just be ignored. We can keep it as you had before:

```suggestion
    const msg = await ChatMessage.create(data);
```

src/live/handleAlbumChat.js
```diff
@@ -1,11 +1,36 @@
...
+    await msg.populate('user', 'name');
+    console.log('Chat message saved:', msg);
+
+    const savedMessage = {
+      ...data,
+      _id: msg._id, // Include _id in the message
+      name: data.name,
```
akosasante: The  `data` object sent to us by the frontend does not have a `name`  field, so this will always be undefined: https://github.com/Code-the-Dream-School/dd-prac-team4-front/pull/52/files#diff-dc044c6ce14501135532d751cb2c59b15079b67646f1634abe6dd9c739c899b4R26

src/live/handleAlbumChat.js
```diff
@@ -1,11 +1,36 @@
...
+    const messagesCount = await ChatMessage.countDocuments({ albumId });
+    if (messagesCount > 50) {
+      await ChatMessage.findOneAndDelete(
+        { albumId },
+        { sort: { timestamp: 1 } }
+      );
+    }
+
+    // Broadcast the message to all connected users in the album chat room
+    io.to(chatRoomName).emit('chat:album', savedMessage);
```
akosasante: My initial suggestion was to return `msg` since that will have all the info we need and any future data we add to `ChatMessage` schema. But this works as well, where you're sending a mix of info from the original socket payload (`data`) and your chosen subset of fields from the database `msg` (in your case, just `_id`). :+1:

src/live/handleAlbumChat.js
mvstoyan: I have already merged the frontend. This means I still need to change the frontend. I really can't understand the logic. As I understand it, I don’t need to add a name to the message scheme, since the name data is already in the user schema. I created a new user with Name and Username. But I don’t understand the logic at what point I should enter the name for it to work. I will try think about this tomorrow more, im sorry😒

akosasante: :wave: I don't believe you'll need to change the frontend. You're correct that name is in the user schema of the message but you're returning a mixed object rather than the `ChatMessage` itself here. Recorded a video I think I'll be able to show what I mean better that way, no need to apologize, sometimes over text things get mixed up or are not as clear! Let me know if you have any questions:

https://drive.google.com/file/d/1hONQ_OrdKTAytdUvjyVdHyv2PtzG5naT/view?usp=sharing

mvstoyan: @akosasante 🤦‍♀️i dont know what to say😅... I apologize for my confusion, it was such straightforward logic. Thank you so much for the explanation.

akosasante (approved):
Awesome work! 🎉

PR #78, opened 2023-10-04
created write endpoint for orderController

NatachaKey:


test/orderController.test.js
```diff
@@ -0,0 +1,132 @@
...
+      tax: 10,
+      total: 110,
+      user: user._id,
+    };
+
+    response = await request(app).post(`/api/v1/orders`).send(orderData);
+    expect(response.status).toBe(StatusCodes.UNAUTHORIZED);
+  });
+  //create an order -Error case - some data is missing
+  //get 401 instead of 400.. not sure why
```
akosasante: 401 = Unauthorized error. In this case, from making a request to the "create order" endpoint without logging in

test/orderController.test.js
```diff
@@ -0,0 +1,132 @@
...
+    //data for order
+    let orderData = {
+      orderItems: [
+        {
+          album: album._id,
+          quantity: 2,
+        },
+      ],
+      subtotal: 100,
+      tax: 10,
```
akosasante: The `tax` field of the `Order` schema expects a fractional number between 0 and 1. So to indicate a tax of 10% we would pass in `0.10` : [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/ef73cbab610ba7d19ae351e6366edd3082a0ecb2/src/models/Order.js#L38-L44](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/ef73cbab610ba7d19ae351e6366edd3082a0ecb2/src/models/Order.js#L38-L44)

Otherwise a validation error will occur. This applies to all the tests in this file where we're passing in a `tax` property

test/orderController.test.js
```diff
@@ -0,0 +1,132 @@
...
+      orderItems: [],
+      subtotal: 100,
+      tax: 10,
+      total: 110,
+      user: user._id,
+    };
+
+    const response = await request(app).post('/api/v1/orders').send(orderData);
+
+    expect(response.status).toBe(StatusCodes.BAD_REQUEST);
```
akosasante, line 136: This test right now is doing a great job of catching an issue in our server code :star\_struck: 

Here in the test we're expecting a `400` bad request response, which would make sense for what to return if the user input is missing some expected values. But instead, the endpoint returns a `500`. This is happening because instead of letting our `error-handler` middleware handle errors that occur in this endpoint, we've wrapped all the code in the controller in a `try/catch` (needed to do our transaction correctly) and then for _any_ error caught, we're manually returning a 500: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/testOrdersEndpoints/src/controllers/orderController.js#L75](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/testOrdersEndpoints/src/controllers/orderController.js#L75)

The suggested fix here would be to update the `orderController.js` code to re-raise the error (`throw error;` ) instead of of returning a `res.status(500)` in the `catch` block. Then, that means that the `error-handler` middleware will get to handle the error as usual and it can decide based on the type of error, what status code and message to return.

test/orderController.test.js
```diff
@@ -0,0 +1,132 @@
...
+    expect(response.status).toBe(StatusCodes.CREATED);
+    expect(response.body).toHaveProperty(
+      'clientSecret',
+      mockPaymentIntent.client_secret
+    );
+    expect(response.body).toHaveProperty('order');
+  });
+
+  //create an order -Error case- user tries to create an order without logging in
+  it('should create an order successfully- Error case- user was not logged in', async () => {
```
akosasante:
```suggestion
  it('should NOT create an order successfully if the user was not logged in', async () => {
```

I find it a bit confusing to say "successfully" in an "error case"

test/orderController.test.js
```diff
@@ -0,0 +1,132 @@
...
+        },
+      ],
+      subtotal: 100,
+      tax: 10,
+      total: 110,
+      user: user._id,
+      paymentIntentId: mockPaymentIntent.id,
+    };
+
+    //AKOS: this one gives 500 status code
```
akosasante: Whenever I need to debug a test to see why it's failing, I usually want to see more logs, so I will either run the test-watch script that continually re-runs tests whenever there's a change in the file: `npm run test:watch` or I'll run the tests just in this file with the `silent` option turned _off_ so that I can see all the logs: `npm run test -- --testPathPattern=test/orderController.test.js --silent=false`.

With those settings I saw that this test was failing with the following validation error:  
`Error while processing the order: Error: Order validation failed: tax: Tax percentage must be at most 1`

So then I changed the `tax` value to `0.10` and re-ran the tests. 

Now instead I got the following error: `Error while processing the order: MongoServerError: Transaction numbers are only allowed on a replica set member or mongos`

Seems like we need some particular settings to let transactions work with our in-memory mongo server. Taking a look at that libraries docs led me to this page: [https://github.com/nodkz/mongodb-memory-server/blob/master/docs/guides/quick-start-guide.md#replicaset](https://github.com/nodkz/mongodb-memory-server/blob/master/docs/guides/quick-start-guide.md#replicaset). Specifically, it says:

> // To use Transactions, the "storageEngine" needs to be changed to \`wiredTiger\`

So we need to import `const { MongoMemoryReplSet } =` _**`require`**_`('mongodb-memory-server');` instead of `MongoMemoryServer`. And then we want to pass in the storage engine when creating the mongo server with: `mongodb = await MongoMemoryReplSet.create({ replSet: { storageEngine: 'wiredTiger' } });` .

test/orderController.test.js
```diff
@@ -0,0 +1,132 @@
...
+        {
+          album: album._id,
+          quantity: 2,
+        },
+      ],
+      subtotal: 100,
+      tax: 10,
+      total: 110,
+      user: user._id,
+      paymentIntentId: mockPaymentIntent.id,
```
akosasante: When the client calls this "Create Order" endpoint, they won't have a payment intent id yet. It's only in this `createOrder` controller that the backend _first_ creates a payment intent instance for this order. 

Trying to pass one in here, will just be ignored by the controller since there's not code in there to parse that field out of the body: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/67cb52b6279196f8fea68ba34dd27db8680ff0d8/src/controllers/orderController.js#L15](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/67cb52b6279196f8fea68ba34dd27db8680ff0d8/src/controllers/orderController.js#L15)

Currently, our orderController uses a test stripe API key, so it will just be making the real request to Stripe as part of this test, using a test key. If we had a real live production key, we of course would not want our tests to be creating real payement intents with that key; so we would set the API key to the test key in the `beforeAll` block, similar to what we do to set the `MONGO_URL` to point to the test database.

Another approach, if we don't want to be actually making requests to Stripe (eg: if we feel it adds to much latency/time to the tests, or if we're concerned that our Github Action CI or somewhere else that we're running the tests won't have access to the network/internet) then we could just [mock](https://spin.atomicobject.com/2021/06/28/mocks-stubs-or-spies/) the stripe call altogether:

```javascript
# This goes either at the very top of your test file (if you want to use the same mock for all tests) or at the top of particular test case where you want this mock
jest.mock('stripe', () => {
  // When `require('stripe')` is called, the below jest mock function will be returned instead
  return jest.fn(() => ({
    // When the `require('stripe')(API_KEY)` is called, the below object will be returned instead
    paymentIntents: {
      // The `stripe.paymentIntents.create` will be replaced with the below jest mock function that returns our mock data
      create: jest.fn(() => ({
        id: 'pi_mock123',
        client_secret: 'mock_secret',
      }))
    }
  }))
});
```

test/orderController.test.js
NatachaKey: @akosasante wow !! thank you!!

src/controllers/orderController.js
```diff
@@ -23,6 +23,10 @@ const createOrder = async (req, res) => {
     const user = req.user; //contains user object
     console.log(user);
 
+    if (!user) {
+      throw new CustomError.UnauthenticatedError('User is not logged in');
+    }
```
akosasante: I believe the `authenticateUser` middleware on the route should already be forcing it to be the case that only logged in users will have their requests handled by this controller function?

akosasante: All the tests pass without this line so I think we can remove this :+1:

test/orderController.test.js
```diff
@@ -0,0 +1,137 @@
...
+
+beforeAll(async () => {
+  mongodb = await MongoMemoryReplSet.create({
+    replSet: { storageEngine: 'wiredTiger' },
+  });
+  const url = mongodb.getUri();
+  process.env.MONGO_URL = url;
+  mongooseConnection = await connectDB(url);
+  server = await app.listen(8001);
+  // process.env.STRIPE_SECRET_KEY = stripeSecretKey; -Akos: I didn't understand what else to do using this approach
```
akosasante: This line can be removed :+1:  Let me know if you have any remaining questions about the approach used here to mock the stripe module :slightly\_smiling\_face:

test/orderController.test.js
```diff
@@ -0,0 +1,137 @@
...
+        client_secret: 'mock_secret',
+      })),
+    },
+  }));
+});
+
+beforeAll(async () => {
+  mongodb = await MongoMemoryReplSet.create({
+    replSet: { storageEngine: 'wiredTiger' },
+  });
```
akosasante, line 32: Might be worth adding a comment here for future readers for why we had to use this different mongo server class and storage engine. Something about "A version of the mongo server that supports replica sets must be used for the order controller test cases because Mongo transactions are used"

test/orderController.test.js
```diff
@@ -0,0 +1,137 @@
...
+      spotifyUrl: 'https://api.spotify.com/v1/albums/unique',
+    });
+  });
+
+  beforeEach(async () => {
+    userCredentials = {
+      email: 'Emily@google.com',
+      password: 'secret',
+    };
+  });
```
akosasante: Do the user credentials need to be recreated before each test? Do we ever modify the values of this variable? If not, I think it's best to put this variable instead in the `beforeAll` block

test/orderController.test.js
```diff
@@ -0,0 +1,137 @@
...
+      user: user._id,
+    };
+
+    const response = await request(app)
+      .post(`/api/v1/orders`)
+      .set('Cookie', signedCookie)
+      .send(orderData);
+
+    expect(response.status).toBe(StatusCodes.CREATED);
+    expect(response.body).toHaveProperty('order');
```
akosasante, line 96: I would like to also verify that the body returns the stripe client secret since that is an important part of this response used by the frontend

test/orderController.test.js
```diff
@@ -0,0 +1,137 @@
...
+      tax: 0.1,
+      total: 110,
+      user: user._id,
+    };
+
+    const response = await request(app).post(`/api/v1/orders`).send(orderData);
+    expect(response.status).toBe(StatusCodes.UNAUTHORIZED);
+  });
+  //create an order -Error case - some data is missing
+  //get 401 instead of 400.. not sure why
```
akosasante: We can remove this comment line now :+1:

akosasante (approved):
Apologies for the delay. Looks good to go with some small cleanup (and run formatting so that CI passes). :rocket:

akosasante:
Oh one last thing, I realized that there's now another 'write' order endpoint, "deleteOrder". You can feel free to implement that in this PR or in a separate one if you prefer.

akosasante:
P.S. If the formatting that Github Actions is complaining about is from outside of your PR; you may need to do `git merge main` (or Update Branch here in Github) to get the latest changes where I've run the format/lint commands on `main` branch.

NatachaKey:
@akosasante Akos, could you please take a look- I am importing Order model to implement tests but get ewwor - create/deleteMany is not a fn

test/orderController.test.js
```diff
@@ -0,0 +1,215 @@
+const { app, connectDB } = require('../src/expressServer'); // Import your Express app here
+const { StatusCodes } = require('http-status-codes');
+const request = require('supertest');
+const { MongoMemoryReplSet } = require('mongodb-memory-server');
+const User = require('../src/models/User');
+const Order = require('../src/models/Order');
```
akosasante: Unlike most of our other model files, `src/models/Order.js` returns an _object_ containing multiple variables, rather than just a single variable: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/testOrdersEndpoints/src/models/Order.js#L171](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/testOrdersEndpoints/src/models/Order.js#L171) 

So when we import it with _require_ we either have to access those objects with the dot syntax: `Order.Order` or destructure it as soon as we require it `const { Order } = require('../src/models/Order');`.

test/orderController.test.js
```diff
@@ -0,0 +1,215 @@
...
+//       email: 'Holly@google.com',
+//       password: 'secret',
+//     };
+//     const signedCookie = await loginAndReturnCookie(adminCredentials);
+//      //Akos: get   TypeError: Order.create is not a function //maybe the error is in export in Order.js? but I've checked that it's exported correctly
+//         // Mock ID
+// const mockOrderId = new mongoose.Types.ObjectId();
+
+//      const orderData = {
+//       _id: mockOrderId,  // Assigning the mock ID
```
akosasante: When we pass this data to `await Order.create(...)` mongoose will save this order in the db and generate an id; why do we need to bypass that by creating our own mock id here first?

test/orderController.test.js
```diff
@@ -0,0 +1,215 @@
...
+
+    const response = await request(app)
+      .delete(`/api/v1/orders/${nonExistingOrderId}`)
+      .set('Cookie', signedCookie);
+
+    expect(response.status).toBe(StatusCodes.NOT_FOUND);
+  });
+
+  // Delete order - Error case - Unauthorized user
+  it('should return a 403 status if user is not authorized to delete the order', async () => {
```
akosasante: Actually, this test is surfacing something a bit strange (which is completely the purpose of writing tests, so great job!):

*   According to `orderRoutes.js` we only allow admins to delete orders using this endpoint: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/testOrdersEndpoints/src/routes/orderRoutes.js#L22](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/testOrdersEndpoints/src/routes/orderRoutes.js#L22)
*   But in the controller function for this endpoint, we're calling `checkPermissions` which is usually used to ensure that only an admin or the matching user can do this action: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/testOrdersEndpoints/src/controllers/orderController.js#L145](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/testOrdersEndpoints/src/controllers/orderController.js#L145)

So in this case, the test is passing because the request is indeed returning a 403 Forbidden. But it's not because it's a user trying to delete anothe user's credentials. If we add some log lines, we can see that the code doesn't even get that far, because the request is rejected first by the `authorizePermissions` middleware and doesn't execute the controller code at all. 

We can likely just reword this test to be about non-admin users not being authorized to delete the order. And in that case, we don't need to create `anotherUser`; we can just log in as `userCredentials` who is already not an admin and thus should receive a 403 Forbidden.

DYA13:
Akos ,thank you for your comments, there is a lot of new information for me and probably I messed up a bit 🙈 
No I still expiriencing the problem with it 
 Authentication API Endpoints > should register a new user and log in
-----
Error: expect(received).toBe(expected) // Object.is equality

Expected: 201
Received: 400
Actually when I try to register in Postman , I receive email with successful registration immediately 
image
But I dont see my registration mongodb and in Postman it is loading wiithout resulting in nothing 
image
I haven't found the problem at the moment , but should be something wrong with registration , I suppose 🫣

test/orderController.test.js
NatachaKey: @akosasante thank you so much!! we should expont and import it in the same way, got it

test/orderController.test.js
NatachaKey: @akosasante oh I was trying everything  to solve the bug thas was caused by not requiring correctly `const { Order } = require('../src/models/Order');.`
will delete it

test/orderController.test.js
NatachaKey: @akosasante I deletes this checkPermissions(req.user, order.user);  from orderController.js 
it seems to have no sence since weonly let admin to delete the order?
image

akosasante: Yep that makes sense to me :+1:

test/orderController.test.js
```diff
@@ -0,0 +1,200 @@
...
+    };
+
+    const response = await request(app)
+      .post(`/api/v1/orders`)
+      .set('Cookie', signedCookie)
+      .send(orderData);
+
+    expect(response.status).toBe(StatusCodes.CREATED);
+    expect(response.body).toHaveProperty('order');
+    expect(response.body).toHaveProperty('clientSecret');
```
akosasante, line 97: Since we did create a mock to handle the Stripe request and told the mock Stripe to return a particular `client_secret`  (on lines 14- 26), we actually have the ability here to make sure that our controller code did the right thing and pases that client secret back to the frontend in the response. So while this is good to check just that the body has the property `clientSecret` , we can go one step further and ensure that the value sent by stripe to the backend, matches the value the backend returns in its response:

```suggestion
    expect(response.body).toHaveProperty('clientSecret');
    expect(response.body.clientSecret).toEqual('mock_secret');
```

akosasante (approved):
One small comment but overall looks good to me, :ship: !

test/orderController.test.js
NatachaKey: @akosasante everything seems to be fine after npm run test , all tests pass, but jest complains about updateOrderStatus function- I guess I am resolving it in a stripe webhook ticket 
image

PR #79, opened 2023-10-05
basic setup without testing -problem on the frontend

NatachaKey:
Akos: has some issues with testings/ was not able to make a payment on the website//maybe this ticket is not merged yet
//also noticed that orders in mongo with status 'pending' are not being deleted as we set it p in the code 
image

### DIFF
--- a/src/controllers/orderController.js
+++ b/src/controllers/orderController.js
@@ -162,4 +162,51 @@ const deleteOrder = async (req, res) => {
   */
 };
 
-module.exports = { createOrder, getAllOrders, getSingleOrder, deleteOrder };
+const handleStripePayment = async (req) => {
+  try {
+    // Receive Stripe Signature
+    const stripeSignature = req.headers['stripe-signature'];
+
+    //  Construct the Webhook Event
+    const event = stripe.webhooks.constructEvent(
+      req.rawBody, // Raw text body payload received from Stripe
+      stripeSignature, // Value of the `stripe-signature` header from Stripe
+      process.env.STRIPE_CLI_WEBHOOK_SECRET // Webhook Signing Secret
+    );
+
+    //  Retrieve Payment Intent and Order Information
+    const paymentIntent = event.data.object; // Payment intent information
+    const orderId = paymentIntent.metadata.orderId; // Order ID passed in metadata
+    const amountPaid = paymentIntent.amount; // Amount paid
+    // Additional order-related info if needed
+
+    //  Update Order Status
+    const order = await Order.findById(orderId);
+
+    if (event.type === 'payment_intent.succeeded') {
+      // Update order status to successful
+      order.status = 'payment_successful';
+    } else if (event.type === 'payment_intent.payment_failed') {
+      // Update order status to failed
+      order.status = 'payment_failed';
+    }
+
+    // Save the updated order
+    await order.save();
+
+    // Handle other actions based on the event type if needed
+
+    return { success: true };
+  } catch (error) {
+    console.error('Error handling Stripe payment:', error);
+    return { success: false, error: error.message };
+  }
+};
+
+module.exports = {
+  createOrder,
+  getAllOrders,
+  getSingleOrder,
+  deleteOrder,
+  handleStripePayment,
+};
--- a/src/controllers/orderController.js
+++ b/src/controllers/orderController.js
@@ -169,7 +169,7 @@ const handleStripePayment = async (req, res) => {
   const stripeSignature = req.headers['stripe-signature'];
 
   //  Construct the Webhook Event
-  const event = stripe.webhooks.constructEvent(
+  const event =  stripe.webhooks.constructEvent(
     req.body, // Raw text body payload received from Stripe
     stripeSignature, // Value of the `stripe-signature` header from Stripe
     process.env.STRIPE_CLI_WEBHOOK_SECRET // Webhook Signing Secret
@@ -196,7 +196,7 @@ const handleStripePayment = async (req, res) => {
 
   // Handle other actions based on the event type if needed
 
-  return res.status(200);
+  return res.status(200).send()
 };
 
 module.exports = {
--- a/src/controllers/orderController.js
+++ b/src/controllers/orderController.js
@@ -164,7 +164,7 @@ const deleteOrder = async (req, res) => {
   */
 };
 
-const handleStripePayment = async (req) => {
+const handleStripePayment = async (req, res) => {
   // Receive Stripe Signature
   const stripeSignature = req.headers['stripe-signature'];
 
--- a/src/controllers/orderController.js
+++ b/src/controllers/orderController.js
@@ -41,6 +41,8 @@ const createOrder = async (req, res) => {
 
       await purchasedAlbum.save({ session });
     }
+    // console.log('this our  order id', order._id.toString()); //see line 53
+
     //Server  tells Stripe that an order is being made (creating a "paymentIntent" object in Stripe)
     const paymentIntent = await stripe.paymentIntents.create({
       amount: total * 100,
@@ -48,7 +50,7 @@ const createOrder = async (req, res) => {
       description: `Order #${order._id}`,
       metadata: {
         userId: req.user.userId,
-        orderId: order._id,
+        orderId: order._id.toString(),
         totalQuantity: orderItems.reduce(
           (total, item) => total + item.quantity,
           0
@@ -169,7 +171,7 @@ const handleStripePayment = async (req) => {
 
     //  Construct the Webhook Event
     const event = stripe.webhooks.constructEvent(
-      req.rawBody, // Raw text body payload received from Stripe
+      req.body, // Raw text body payload received from Stripe
       stripeSignature, // Value of the `stripe-signature` header from Stripe
       process.env.STRIPE_CLI_WEBHOOK_SECRET // Webhook Signing Secret
     );
@@ -177,7 +179,6 @@ const handleStripePayment = async (req) => {
     //  Retrieve Payment Intent and Order Information
     const paymentIntent = event.data.object; // Payment intent information
     const orderId = paymentIntent.metadata.orderId; // Order ID passed in metadata
-    const amountPaid = paymentIntent.amount; // Amount paid
     // Additional order-related info if needed
 
     //  Update Order Status
@@ -185,10 +186,10 @@ const handleStripePayment = async (req) => {
 
     if (event.type === 'payment_intent.succeeded') {
       // Update order status to successful
-      order.status = 'payment_successful';
+      order.orderStatus = 'payment_successful';
     } else if (event.type === 'payment_intent.payment_failed') {
       // Update order status to failed
-      order.status = 'payment_failed';
+      order.orderStatus = 'payment_failed';
     }
 
     // Save the updated order
--- a/src/controllers/orderController.js
+++ b/src/controllers/orderController.js
@@ -196,7 +196,7 @@ const handleStripePayment = async (req, res) => {
 
   // Handle other actions based on the event type if needed
 
-  return res.status(200).send()
+  return res.status(200).end()
 };
 
 module.exports = {
--- a/src/controllers/orderController.js
+++ b/src/controllers/orderController.js
@@ -73,10 +73,9 @@ const createOrder = async (req, res) => {
     await session.abortTransaction();
 
     console.error('Error while processing the order:', error);
-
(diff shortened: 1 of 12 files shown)


src/expressServer.js
```diff
@@ -185,4 +191,43 @@ global.io = io;
 // export the database connection function
 const connectDB = require('./db/connect');
 
+// ======STRIPE WEBHOOK SETUP ======
```
akosasante: Is this section (lines 194-232) for testing purposes or something else? I'm unsure about having this route here in this file rather than a controller, and about having a separate server start up here listening on  port 4242 :thinking:

src/expressServer.js
```diff
@@ -86,7 +89,10 @@ app.get('*swagger-initializer.js', (req, res) =>
 
 // Serve Swagger UI API documentation from the static files path
 app.use('/api-docs', express.static(pathToSwaggerUi));
-
+app.use(
+  '/api/v1/orders/payment_status',
+  express.raw({ type: 'application/json' })
+);
```
akosasante, line 92: :+1:

src/expressServer.js
akosasante: Along with line 20, and 26-28

src/controllers/orderController.js
```diff
@@ -162,4 +162,51 @@ const deleteOrder = async (req, res) => {
...
 
-module.exports = { createOrder, getAllOrders, getSingleOrder, deleteOrder };
+const handleStripePayment = async (req) => {
+  try {
+    // Receive Stripe Signature
+    const stripeSignature = req.headers['stripe-signature'];
+
+    //  Construct the Webhook Event
+    const event = stripe.webhooks.constructEvent(
+      req.rawBody, // Raw text body payload received from Stripe
```
akosasante: Since we run the `express.raw` middleware in `expressServer.js` for this endpoint, that middleware puts the raw bytes body into the `req.body`  property, so we can access it there (I think here `req.rawBody` may just be undefined, but I haven't checked it myself with console logging)

```suggestion
      req.body, // Raw text body payload received from Stripe
```

src/controllers/orderController.js
akosasante, line 1: I noticed while trying this out that currently, the payment intents that were going to Stripe in the stripe dashboard, did not have an `orderId` in the metadata, even though we have this in our `createOrder` controller:

```
    //Server  tells Stripe that an order is being made (creating a "paymentIntent" object in Stripe)
    const paymentIntent = await stripe.paymentIntents.create({
      amount: total * 100,
      currency: 'usd',
      description: `Order #${order._id}`,
      metadata: {
        userId: req.user.userId,
        orderId: order._id,
        totalQuantity: orderItems.reduce(
          (total, item) => total + item.quantity,
          0
        ),
        totalAlbums: orderItems.length,
      },
    });
```

I added some console logging and found that `console.log(order._id)` in that function returns an _object_ (`new ObjectId(....)`) rather than a _string_. And then I guess Stripe just ignores it and throws it away since it expects only string values in the metadata object.

We can fix this by just changing that line to `order._id.toString()`.

That way in the `handleStripePayment` we can have access to metadata.orderId :tada:

src/controllers/orderController.js
```diff
@@ -162,4 +162,51 @@ const deleteOrder = async (req, res) => {
...
+
+    //  Update Order Status
+    const order = await Order.findById(orderId);
+
+    if (event.type === 'payment_intent.succeeded') {
+      // Update order status to successful
+      order.status = 'payment_successful';
+    } else if (event.type === 'payment_intent.payment_failed') {
+      // Update order status to failed
+      order.status = 'payment_failed';
```
akosasante: The Order object in our app does not have a `status` field :wink:

akosasante (commented):
@NatachaKey  Sorry about that! I think you'll need to run @JafariM 's branch `PS-81-create-sidebar-for-order-details` on the frontend; that should have all the fixes that let the full checkout flow work properly.

src/expressServer.js
NatachaKey: for sure it's useless- inicially thought it will help to run stripe webhook

src/controllers/orderController.js
NatachaKey: @akosasante oops! it's orderStatus! I swear i was checking it lol

src/controllers/orderController.js
```diff
@@ -162,4 +164,50 @@ const deleteOrder = async (req, res) => {
...
+
+    // Save the updated order
+    await order.save();
+
+    // Handle other actions based on the event type if needed
+
+    return { success: true };
+  } catch (error) {
+    console.error('Error handling Stripe payment:', error);
+    return { success: false, error: error.message };
```
akosasante: In a controller handler `return` returns a value to the calling backend function but it _does not send an HTTP response_. So you'll see in your terminal when running the Stripe CLI that Stripe never gets a response and after waiting around a while it will print:

```
[ERROR] Failed to POST: Post "http://localhost:8000/api/v1/orders/payment_status": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

https://support.stripe.com/questions/webhooks-how-to-investigate-and-fix-timed-out#:~:text=Stripe%20expects%20your%20server%20to,the%20Event%20was%20successfully%20received.

We need to instead be returning a status OK (200-299 status code; in our case just a plain 200 will suffice). If you want you can return the json {success: true} but I don't think Stripe cares/checks the body anyway. 

In the error case, we could check the error more to figure out if this is a problem with what Stripe sent us (BAD\_REQUEST, or something else in the 400s) or a problem with our code (500s error). But most likely, since Stripe is always sending a specific payload, it's unlikely that they will suddenly start sending us some bad data. So we can probably just remove this try/catch; and let the error be handled by the error-handler middleware which will log the error for us already and return a 500.

akosasante (commented):
One last thing to fix and then you should be good to go; once you have stripe cli working locally!

src/controllers/orderController.js
```diff
@@ -162,4 +164,45 @@ const deleteOrder = async (req, res) => {
...
+    // Update order status to failed
+    order.orderStatus = 'payment_failed';
+  }
+
+  // Save the updated order
+  await order.save();
+
+  // Handle other actions based on the event type if needed
+
+  return res.status(200);
```
akosasante: Note the `linter` message here; our function right now only defined `req` as an argument; we should have `res` up on line 167  as well like our other controller functions.

src/models/Order.js
```diff
@@ -103,6 +103,7 @@ const updateOrderStatus = async () => {
     console.error('Error updating orders:', error);
   }
 };
+console.log('meaning', process.env.START_ORDER_INTERVAL);
```
akosasante, line 106: :eyes:

src/controllers/orderController.js
```diff
@@ -162,4 +164,45 @@ const deleteOrder = async (req, res) => {
...
```
akosasante: Another part I missed here earlier; `res.status(...)` is not enough to tell Express to send the response. It waits to know what _body_ to send if any.

Since Stripe doesn't care about the response body, just the status, we can send an empty response like this: `res.status(200).end()`

(`res.status(200).json({})` or `res.status(200).send()` also would work though those are technically sending a non-empty response body, with either just a blank json object in the first case, or an empty string response body in the second case).

NatachaKey: @akosasante , Akos,  why- if our `event.type === **'payment_intent.succeeded'` - and this is what we do receive after making a payment + the `orderStatus `in mongo is being changed to `payment successful`- but  why in stripe dashboard we  still see 
that webhook delivery failed- once it returned timed out error- which i was trying to resolve with adding await to 
```
 const event =  stripe.webhooks.constructEvent(
    req.body, // Raw text body payload received from Stripe
    stripeSignature, // Value of the `stripe-signature` header from Stripe
    process.env.STRIPE_CLI_WEBHOOK_SECRET // Webhook Signing Secret
  );
```
in handleSptipePayment controller  but it didn't help- and now I get 404 - besides I've added empty body to  `res.status(200).end()`
image
image

akosasante: Hi @NatachaKey I just tried with your latest changes and things seem to be working well.

One thing I didn't realize, as I'm a bit new to Stripe development as well, is that the Webhooks table you've screenshotted, will only tell us things from the perspective of the live website. And since we haven't merged this PR yet, of course, the live website does not yet have a `/api/v1/orders/payment_status` endpoint, which is what is causing the "Route not found" 404 errors seen there. I thought that forwarding to the CLI meant that it would _only_ call the localhost version, but it looks like it actually copies the request so it's calling both the production music app server AND your local one. 

The main thing we are concerned about here at the moment before merging is that the code works as expected when we forward the call to our local servers.

Here are the steps I took to test:

- Frontend running `PS-81-create-sidebar-for-order-details` branch
- Backend running this `stripeWebhook` branch
- Another terminal with Stripe CLI running and forwarding any calls made to the live webhook endpoint: `D:\stripe\stripe.exe listen --load-from-webhooks-api --forward-to localhost:8000`
- Created an order by opening `localhost:3000`, adding an album to cart and clicking "Proceed to Checkout"
  - At this point, verified from backend logs that a POST request was made to `/api/v1/orders` ✔️ 
image
  - Verified that the Mongo database had an order created with `orderStatus` being `pending` and a `paymentId` populated ✔️ 
  - Verified in Stripe developer dashboard that an incomplete payment was started for the order: https://dashboard.stripe.com/test/payments ✔️ 
  - In my terminal that's running the Stripe CLI, I could see that the Stripe servers had received a `payment_intent.created`  call ✔️ 
image
- Submitted a valid payment for the order on the checkout page using a Stripe test credit card number
  - Verified the backend logs that a POST request was made to` /api/v1/orders/payment_status` and that 200 response was returned ✔️
image 
  - Verified in Mongo database that the same order had its `orderStatus` updated to `payment_successful` ✔️
  - Verified in Stripe CLI that it tool had made a request to the correct endpoint and received a 200 ✔️
image
  - Verified in Stripe dashboard in the **EVENTS** section of the Developers page, that the response had been successfully received in the CLI: https://dashboard.stripe.com/test/events/evt_3O1u3CKnb0YvaiB30lebEI7X ✔️ 
image


If the above steps work for you as well, then you can feel free to merge!

akosasante (approved):
🚀  (after any lint/format issues are fixed)

src/controllers/orderController.js
NatachaKey: @akosasante got it!! thank you so much!!! You are the BEST!!!😍
image
just applied formatting and linting -not sure why the CI jobs fail here - I've checked formatting several times locally and when I push it says that there's nothing to commit..
image

NatachaKey: image

akosasante: :tada: 

---

That's odd, when I pull this branch and run the prettier check I get:

```
aasante@aasante-C02F75HHMD6R backend % git switch stripeWebhook
Switched to branch 'stripeWebhook'
Your branch is behind 'origin/stripeWebhook' by 2 commits, and can be fast-forwarded.
  (use "git pull" to update your local branch)
aasante@aasante-C02F75HHMD6R backend % git pull
Updating 62806f8..b75d636
Fast-forward
 src/controllers/orderController.js | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

aasante@aasante-C02F75HHMD6R backend % npm run format:check

> bb-practicum-team-4-back@1.0.0 format:check
> prettier --check .

Checking formatting...
[warn] src/controllers/orderController.js
[warn] Code style issues found in the above file. Run Prettier to fix.
```

I wonder if updating your branch (either in Github using the button, and then do `git pull` OR in your terminal by doing `git fetch origin main && git merge origin/main`) . And then running the format + lint fix commands maybe will work?

---

Ohhh interesting that last screenshot might help us fix husky issue!

typicode/husky#944

Try the following:

*   copy the content of `.husky/pre-commit` into another file for safekeeping
*   run `npx husky-init` (unfortunately this will overwrite the `pre-commit` file)
    *   Check that there is a `.husky/_/husky.sh` file existing now
*   paste in the contents that were in the `pre-commit` file before back into it
*   try testing husky again the way we were talking about in the other PR (make a bad syntax change, try to add and commit it and see if `lint-staged` runs).

NatachaKey: image
oh thank you, finally  able to merge it

NatachaKey: @akosasante omg!!! thank youu!!! works now automatically after commiting
image

akosasante: Awesome!! So glad we got it working 🎉

PR #81, opened 2023-10-30
When a user pauses listening to an album send an update to the backend PS-181

DYA13:


.eslintrc.json
```diff
@@ -18,5 +18,9 @@
     "es2023": true,
     "browser": true,
     "jest/globals": true
+  },
+  "parserOptions": {
+    "ecmaVersion": 2023,
+    "sourceType": "module"
```
akosasante: I am curious why this was added?

src/views/listeningAlbum.ejs
akosasante, line 1: Looks good! Next, there should be a route that allows navigating to this page. Similar to the `orderNotifications` route here: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/socket_pause_listening/src/expressServer.js#L141-L144.](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/socket_pause_listening/src/expressServer.js#L141-L144.)  Use `res.render` to tell Express to render the HTML from an ejs file in the views folder

src/public/listeningAlbum.js
```diff
@@ -0,0 +1,15 @@
+import io from 'socket.io-client';
```
akosasante: This isn't working for me; I'm seeing a browser error about this import:



What we found worked previously for the orderNotifications page, was to import the socket-io client library in the HTML/EJS file:

`<script src="/socket.io/socket.io.js"></script>` (see: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/socket_pause_listening/src/views/orderNotifications.ejs#L7)](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/socket_pause_listening/src/views/orderNotifications.ejs#L7))

src/public/listeningAlbum.js
akosasante, line 1: For client-side javascript (in HTML documents, or in this case our server-rendered EJS/HTML doc), it's good practice to wrap the code in a DOM function that makes sure that the code will only run once the page and its scripts have finished loading. One way is to wrap it in an event listener: 

```js
document.addEventListener('DOMContentLoaded', function () {
 ... all your code here ...
 });
```

src/public/listeningAlbum.js
```diff
@@ -0,0 +1,15 @@
...
+  socket.emit('listening-to-album-play');
+});
+
+document.getElementById('pauseButton').addEventListener('click', () => {
+  socket.emit('listening-to-album-pause');
+});
+
+window.addEventListener('beforeunload', () => {
+  socket.emit('user-leaving-page');
+});
```
akosasante: :star\_struck:

src/live/index.js
```diff
@@ -25,6 +26,21 @@ const onConnect = (io, socket) => {
   socket.on('chat:album', async (data) => {
     handleAlbumChat(io, socket, data);
   });
+
+  socket.on('listening-to-album-play', (albumId) => {
+    albumController.startAlbumPlayback(albumId);
```
akosasante: Currently this function doesn't exist; what are you planning on having the controller do? So far in the ticket outlines, we will just be emitting a message back to the frontend so that other users can see what this user is listening to; or we can list all the users who are listening to this album right now. But if you have additional functionality to add, I'm all for it!

src/live/index.js
```diff
@@ -25,6 +26,21 @@ const onConnect = (io, socket) => {
...
+    console.log('Album is now playing');
+  });
+
+  socket.on('listening-to-album-pause', () => {
+    io.emit('album-paused', 'Album is now paused');
```
akosasante: Keep in mind that `emit` sends this message to _all_ sockets that have subscribed to the `album-paused` event. But right now the message we're emitting doesn't give those listeners much context about _which_ album was paused and by _whom_.

.eslintrc.json
DYA13: I had a problem that import of module was not possible and to resolve it I have found this solution.  Actually below in comments seems we clarified that I need to delete it as there is another solution.😊

src/views/listeningAlbum.ejs
DYA13: It is done here in this commit 561f8af3dc26d58ee4e223d4d03d58a0f791b90d

src/public/listeningAlbum.js
DYA13: I had doubts about it as well, on this reason I added in eslintrc.json the following  (it is your first comment): 
 "parserOptions": {
    "ecmaVersion": 2023,
    "sourceType": "module" 
I tried it as an option to resolve the problem 😄
I have added in ejs as per your comment , commit 4dec20b67555d1faef548824f0a49ef22d8b2568
In this case I can remove parserOptions from eslintrc.json  , right?

src/public/listeningAlbum.js
DYA13: wrapped in DOM function , commit 2a8be4a9ee8790be8590aded7c4a8c5f488beded

src/public/listeningAlbum.js
DYA13: ☺️

src/live/index.js
DYA13: Actually I was thinking about to suggest similar albums or tracks to users based on their listening history or preferences. But possibly it should a part of this ticket, I don't know.

src/live/index.js
DYA13: Yes,  you  are right, it is true, should I delete it ?

src/expressServer.js
```diff
@@ -142,6 +142,10 @@ app.use(express.static(__dirname + '/public'));
 app.get('/order-notifications', (req, res) => {
   res.render('orderNotifications'); // Render the EJS template
 });
+// Define a route for the listening page
+app.get('/listening', (req, res) => {
+  res.render('listeningAlbum');
+});
```
akosasante, line 148: :+1:

src/public/listeningAlbum.js
akosasante: :+1:

.eslintrc.json
akosasante: Yes I believe you should be able to remove this now :+1:

src/public/listeningAlbum.js
akosasante: Yes I believe you can remove it.

In order to tell ESLint, that `io()` is an available global variable in this script, we should add the following comment at the top of the page: `/* global io */` (see similar comment in `orderNotifications.js` where we have both `io` and the `Toastify` library being imported as global scripts)

src/live/index.js
akosasante: I think likely you'll end up combining the changes you have here with your other PR right? I think that one has a good state right now of setting up rooms and emmitting a more full set of info. As long as you end up merging those in as the final state it should be fine.

src/live/index.js
akosasante: Ah I see, that's really cool! Spotify even has an API for that :open\_mouth: : [https://developer.spotify.com/documentation/web-api/reference/get-recommendations](https://developer.spotify.com/documentation/web-api/reference/get-recommendations). I can look into it a bit more and write a ticket around that. 
  
For now, we can just comment out this bit though since it will cause errors as it is since there is no function called `startAlbumPlayback` being exported from that controller file.

.eslintrc.json
DYA13: removed commit b74dcfc8b9e800c1e4ae5dba610c6e36a20f4c20

src/public/listeningAlbum.js
DYA13: I have removed from eslintrc.json ParserOptions and added global variable at the top listeningAlbum.js but I have it as a eslint problem anyway

image

src/live/index.js
DYA13: I have commented out this block with startAlbumPlayback function at the moment

akosasante: sounds good 👍  You could also leave the event listener in and just comment out line 31; that way you coould still have a console.log line here to test that the connection between the frontend and backend is working for the play button.

src/public/listeningAlbum.js
akosasante: Yes we must remove this `import` line altogether: `import io from 'socket.io-client';`. It won't work because the js script won't be able to import without the parser options change; and even with the parser options change, that path to the socket client file is not valid.

It is not needed because we have already essentially 'imported' the socket script into the browser's global environment, in our HTML/EJS file here: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/6274fe20437cfb68eb63fcfe1d10642b5eee86d9/src/views/listeningAlbum.ejs#L7

The guidance from the docs recommend doing it this way: https://socket.io/docs/v4/client-installation/#installation

akosasante:
After fixing the import issue; you'll want to make sure you merge in the changes from `main` since the PR you merged will have some conflicts with the changes you have here. I'm thinking you'll want to mostly keep the changes from the other PR; with the addition of the  "user left the page" event from this PR.

src/public/listeningAlbum.js
DYA13: Thank you , it is removed  commit fd311ed43a88bf4c9fcdf64bfcf4c6b8ced38a99

src/live/index.js
DYA13: I have removed albumId as parameter due to eslint problem. As albumId is declared but never used. commit 7c301b1032f5bb7d211517d360af28dc600fe5bc

src/live/index.js
```diff
@@ -49,6 +50,22 @@ const onConnect = (io, socket) => {
...
+  socket.on('listening-to-album-play', () => {
+    // albumId should be added as a parameter   socket.on('listening-to-album-play', (albumId) => {
+    //   albumController.startAlbumPlayback(albumId);
+    console.log('Album is now playing');
+  });
+
+  socket.on('listening-to-album-pause', () => {
+    io.emit('album-paused', 'Album is now paused');
+    console.log('Album is now paused');
+  });
```
akosasante: After your latest merge/commit I'm seeing two `listening-to-album-pause` and `listening-to-album-play` handlers here. Please remove one (probably these two highlighted can be removed for now)

akosasante (approved):
After fixing the duplicate event handlers, you should be good to merge 🎉

src/live/index.js
DYA13: Removed duplicated handlers, commit 4a3c59816c9eb9ca69a9cfb1002af8b074554931

PR #82, opened 2023-10-30
Backend send frontend websockets live updates of when listening starts/stops PS-183

DYA13:


akosasante (commented):
This looks great! 

Depending on how we want to use these events in the frontend, it can be really nice to have these socket events as "rooms" because then the play/pause messages will only be sent to subscribers of the `pause`/`play` events who previously were listening to that album. 

Keep in mind that we can send a payload that isn't just a string, if we want to send some additional data. Eg, we could do: `socket.to(albumRoom).emit('pause', { user: userId, album: albumId, message: "User paused the album"})`

DYA13:
> This looks great! 
> 
> Depending on how we want to use these events in the frontend, it can be really nice to have these socket events as "rooms" because then the play/pause messages will only be sent to subscribers of the `pause`/`play` events who previously were listening to that album. 
> 
> Keep in mind that we can send a payload that isn't just a string, if we want to send some additional data. Eg, we could do: `socket.to(albumRoom).emit('pause', { user: userId, album: albumId, message: "User paused the album"})`
Yeah , it is a very nice idea! 👍
Could you please check my implementation of rooms for albums? commit 286d5ae0e1415abe7dde823aa10b5cc8a32b95e9

I suppose it will be necessary to make some changes in listeningAlbum.js, but the code is in the other branch not merged yet
PS: I have added with comment for PR 82 , commit a050002899807d3b74b2d61ec9bd622e9bd8eee0

akosasante (approved):
LGTM!

PR #83, opened 2023-11-03
An endpoint that returns album/song recommendations based on a user's play history PS-210

DYA13:
Akos, I've begun working on an endpoint, but I'm uncertain if I've taken the correct approach. Could you please review what I've done so far and provide guidance or suggestions? Thank you❤️


src/models/AlbumRecommendation.js
akosasante, line 1: I think this is fine for now; we might end up later deciding to store as: `userId: String, songId: [String]`. So that one Mongo document has _all_ the recommendations for a single user; instead of storing each recommendation as a separate row/document. 

Also, once we implement more of this feature, we'll have a better idea of this will recommend songs or albums or artists or maybe all three. But I'm fine with leaving it as songs for now 👍

akosasante (commented):
Looks good! Next steps would be refactoring to follow best practices for defining our routes (controller/routes folder, allow error handling middleware to catch errors, etc). 

Once that's done, this should be good to merge! I'll add a ticket later today for how we could go about generating those recommendations in the backend when the user has clicked play on a new song/album 😄

src/models/AlbumRecommendation.js
DYA13: I have done refactoring , I created controller and route files  , I hope I didn't mess up too much 😄

src/expressServer.js
```diff
@@ -82,6 +79,7 @@ app.use(
     stream: { write: (message) => logger.info(message.trim()) },
   })
 );
+app.use('/recommendations', recommendationRoutes);
```
akosasante:
```suggestion
app.use('api/v1/recommendations', recommendationRoutes);
```

And let's please move it to line 155, so that this line occurs _after_ all our middleware that we're using to keep our API endpoints secure, logged, and handled in a consistent manner.

src/expressServer.js
```diff
@@ -38,6 +40,22 @@ app.get('/', (req, res) => {
...
+      await recommendationController.generateAlbumRecommendations(userId);
+
+    res.json(recommendations);
+  } catch (error) {
+    console.error('Error while generating recommendations:', error);
+    res
+      .status(500)
+      .json({ error: 'An error occurred while generating recommendations.' });
+  }
+});
```
akosasante: This code should live in the router and controller. The express server should have minimal business logic for _how_ a request should be handled. This makes it easier for us to test separate routes in our unit tests and easier for someone new reading the code to reason about how our code is organized. 

Currently, the only cases where we keep the full request/response cycle here in the express server initialization file is  for things that are short like:
- health check route on line 32
- returning the simple html for our backend root page on line 38
- serving documentation on `/tests` route
- rendering views (though these honestly should be moved to their own router+controller eventually)

And for things that we explicitly _don't_ want to be affected by our middleware like:
- swagger routes on lines 96 and 102


Anything else that is an API endpoint that we expect a browser to call (whether the React frontend code or a server-rendered EJS page) should be written following our existing patterns:
- `app.use("/api/v1/<name of route>, <imported router module>)` in expressServer.js
- `router.route` and `router.<method>` calls in a `src/routers/<router name>` file
- `app.<method>` request handlers that handle reading `req` and writing to `res` in a `src/controllers/<controller name>` file..

src/routes/recommendationRoutes.js
```diff
@@ -0,0 +1,20 @@
...
+    const recommendations =
+      await recommendationController.generateAlbumRecommendations(userId);
+    res.json(recommendations);
+  } catch (error) {
+    console.error('Error while generating recommendations:', error);
+    res
+      .status(500)
+      .json({ error: 'An error occurred while generating recommendations.' });
+  }
+});
```
akosasante: Preferably, to follow our current conventions, the code starting with `async (req, res) => ...` should all live in the controller. The router file should just be the definition of "this route is handled by this controller function". So  something like: `router.get('/:userId', getAlbumRecsForUser)`

src/expressServer.js
DYA13: Corrected and moved on line 154, commit 080bc44e62efd44e56c1b970ed3fb373362d1c38

src/expressServer.js
DYA13: Thank you Akos❤️, actually I moved this code to routes/recommendationRoutes.js ,I have forgotten to remove it from here. 🙈

DYA13: commit 69773ada3796d96c7cbbaabc4b872af3e5893819

src/routes/recommendationRoutes.js
DYA13: I moved getAlbumRecsForUser function in controller ,commit c89f45fe910de2dad5613d88d052837c137e4b30

src/expressServer.js
```diff
@@ -17,6 +17,8 @@ const pathToSwaggerUi = require('swagger-ui-dist').absolutePath();
 const { readFileSync } = require('fs');
 const { join } = require('path');
 const expressStaticGzip = require('express-static-gzip');
+const recommendationRoutes = require('./routes/recommendationRoutes');
+const recommendationController = require('../controllers/recommendationController');
```
akosasante: can be removed per lint warning 👍

src/controllers/recommendationController.js
```diff
@@ -0,0 +1,23 @@
...
+  try {
+    const recommendations =
+      await recommendationController.generateAlbumRecommendations(userId);
+    res.json(recommendations);
+  } catch (error) {
+    console.error('Error while generating recommendations:', error);
+    res
+      .status(500)
+      .json({ error: 'An error occurred while generating recommendations.' });
+  }
```
akosasante: Keep in mind, now that we've defined this route in `expressServer` _under_ all our middleware, that means any errors raised by this code will be handled by: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/c6ca1551a1d1906a3d66bd33174b52f4c58bd98b/src/middleware/error-handler.js#L3-L2

Which means it is redundant to have the try/catch here (because the errors that occur here, will be caught by the middleware function and handled centrally in that function; instead of every controller code needing to be wrapped with try/catch). 

```suggestion
  const recommendations =
    await recommendationController.generateAlbumRecommendations(userId);
  return res.json({ recommendations: recommendations });
```

akosasante: And for following common convention in this app, we will also wrap our return value in an JSON object `{}` with the key `recommendations`.

src/controllers/recommendationController.js
```diff
@@ -0,0 +1,23 @@
+const AlbumRecommendation = require('../models/AlbumRecommendation');
+const recommendationController = require('../controllers/recommendationController');
+async function generateAlbumRecommendations(userId) {
+  return await AlbumRecommendation.find({ userId });
+}
+const getAlbumRecsForUser = async (req, res) => {
+  const { userId } = req.params;
+  try {
+    const recommendations =
+      await recommendationController.generateAlbumRecommendations(userId);
```
akosasante: At this point, having `generateAlbumRecommendations` as a separate function doesn't add too much and may add a bit of misdirection. Let's refactor so that the call to the model is inline here again instead of a calling a function to call a function 😄

src/expressServer.js
DYA13: I have removed line 21,  as route  line 20 is required on line 137 
`app.use('api/v1/recommendations', recommendationRoutes);`
commit a2c198d2a525d0bdd8623b71066d27ef1cef8e6d

src/controllers/recommendationController.js
DYA13: Refactoring done 😊

src/expressServer.js
```diff
@@ -133,6 +134,7 @@ app.use('/api/v1/orders', orderRouter /* #swagger.tags = ['Orders'] */);
 app.use('/api/v1/reviews', reviewRouter /* #swagger.tags = ['Reviews'] */);
 app.use('/api/v1/wishlist', wishlistRoutes /* #swagger.tags = ['Wishlist'] */);
 app.use('/api/v1/chat', chatRoutes);
+app.use('api/v1/recommendations', recommendationRoutes);
```
akosasante:
```suggestion
app.use('/api/v1/recommendations', recommendationRoutes);
```

without the leading "slash" this is currently defining a route at: `localhost:8000api/v1/recommendations/...` instead of `localhost:8000/api/v1/recommendations/...`

akosasante (approved):
One small fix and you should be good to merge. You should be able to successfully test this route using a `curl` command or using Postman. Once you're confident it works as expected, feel free to merge

DYA13:
Akos , I checked in Postman, it works 

image

PR #84, opened 2023-11-11
populate name

mvstoyan:


src/routes/chatRoutes.js
```diff
@@ -6,7 +6,8 @@ router.get('/:albumId', async (req, res) => {
   const { albumId } = req.params;
   const messages = await ChatMessage.find({ albumId })
     .sort({ timestamp: -1 }) // Sort messages in reverse chronological order
```
akosasante, line 8: I'm not sure if you actually want these in reverse chronological order. When you'res showing them in the frontend, this ends meaning the most recent messages are scrolled all the way at the top of the chatbox

Screenshot 2023-11-12 at 11 29 44 AM

akosasante (approved):
One pointer and then looks good 👍

PR #85, opened 2023-11-12
store imgs in mongo

NatachaKey:


src/models/User.js
```diff
@@ -67,7 +56,7 @@ const UserSchema = new mongoose.Schema(
       enum: ['admin', 'user'],
       default: 'user',
     },
-    profileImage: profileImageSchema,
+    profileImage: Buffer,
```
akosasante: In this case, since you're not storing the raw image but instead a base-64 encoded string that _represents_ the image, the type of this field is not `Buffer` but `String`:

(see this part of the docs: https://codethedreamteam4-2023.atlassian.net/wiki/spaces/PT4/pages/10059785/Store+images+in+MongoDB+directly#:~:text=Or%20we%20could%20even%20change%20profileImage%20from%20storing%20a%20Buffer%20type%20to%20instead%20storing%20a%20plain%20String%20type.%20Then%20we%20could%20make%20sure%20to%20convert%20the%20uploaded%20image%20to%20a%20string%20first%20before%20storing%20it%20in%20the%20db%2C%20or%20to%20help%20the%20frontend%20even%20more%2C%20store%20it%20as%20a%20data%20URL%3A)

src/controllers/userController.js
```diff
@@ -121,6 +122,57 @@ const updateUserPassword = async (req, res) => {
      #swagger.responses[401] = { description: 'Error. Invalid credentials.' }
   */
 };
+
+const uploadUserImage = async (req, res) => {
+  const userId = req.params.id;
+  fileUploadMiddleware({ limits: { fileSize: 10000000 }, abortOnLimit: true }),
+    async (req, res) => {
```
akosasante: The syntax here is not quite correct. You have a variable `const uploadUserImage` which is being set to a function `async (req, res) => {` which is then on line 128 calling a middleware, then somehow starting another `async (req, res)` function.

---

Generally, for calling middleware we've been putting it in our `router`s where we have the url path, followed by any middleware functions, finally followed by the request handler.

So in `userRoutes.js` we would have: `router.route('/:id/uploadUserImage').post(authenticateUser, fileUploadMiddlware(...put the options here...), uploadUserImage);`

akosasante: Then in here we can remove the `fileUploadMiddlware` and the second `async (req, res)` lines. I haven't tested your current code but I'm not sure that the code inside the second `async (req, res)` would be reached at all as it is.

NatachaKey: @akosasante thank youuu! I think I got it- noy able to see the foto on the frontend yet cause we don't display it, but able to see it in mongo: 
image
image

akosasante: Nice! I don't see the changes pushed here in GH yet but happy to take a look when they're ready.

Feel free to create a branch in the frontend to test showing the profile image with this approach; should be basically able to do this by passing in `profileImage` value to the `img.src` attribute (as shown here: https://codethedreamteam4-2023.atlassian.net/wiki/spaces/PT4/pages/10059785/Store+images+in+MongoDB+directly#Frontend:~:text=Then%2C%20the%20frontend%20will%20only%20need%20to%20do%20%3Cimg%20src%3D%7Buser.profileImage%7D%20alt%3D%22user%27s%20profile%20picutre%22%20/%3E%20with%20no%20additional%20decoding%20needed%20on%20their%20part.)

NatachaKey: @akosasante Akos, here is a commit with that changes- https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/85/commits/b6edb0b8e8fc30ecc1ee829ad516a910f676660d 
// will add frontend as well to see changes in browser

src/controllers/userController.js
```diff
@@ -121,6 +122,55 @@ const updateUserPassword = async (req, res) => {
...
+  if (!user) {
+    return res.status(400).send('no user found');
+  }
+
+  // Assuming that the request named the file sent to us, `profile`
+  // The contents of this variable are outlined here: https://www.npmjs.com/package/express-fileupload#:~:text=The%20req.files,the%20uploaded%20file
+  //?? CHECK IN POSTMAN WHICH PROPERTIES HAS REQ.FILES????
+  console.log(req.files);
+
+  const profile = req.files.foo;
```
akosasante: We should probably also do a check for if `profile` is null (if the request forgot to pass in a `foo` property or mistyped it or something) and return a 400 with appropriate error message in that case

src/routes/userRoutes.js
```diff
@@ -43,4 +45,13 @@ router
   .route('/:id')
   .delete(authenticateUser, authorizePermissions('admin'), deleteSingleUser);
 
+router.route('/:id/uploadUserImage').post(
```
akosasante: Since we're using `authenticateUser`, we are already requiring that the user is logged-in, and we can access that logged-in user's id with `req.user.userId`. 

I would suggest that it doesn't really make sense to also need the caller to pass in the user's id in the URL. It might also make it possible that a request passes in a different user's id in the URL than the requesting logged-in user. Maybe better to avoid the `:id` parameter altogether in this case

src/controllers/userController.js
```diff
@@ -6,6 +6,7 @@ const {
   attachCookiesToResponse,
   checkPermissions,
 } = require('../utils');
+const fileUploadMiddleware = require('express-fileupload');
```
akosasante: This can be removed here

src/controllers/userController.js
akosasante: Any particular reason to make the property `foo`, instead of something more descriptive about the use-case?

akosasante (approved):
Looks good! Left some last suggestions but overall it is well-implemented and correctly saves the profile image on the user 👍🏾

src/controllers/userController.js
NatachaKey: @akosasante will use **userImage** instead of **foo**// just thought after reading in docs it was obligatory to name it _foo_

src/routes/userRoutes.js
NatachaKey: @akosasante oh thank you , Akos!! got it! that absolutely makes sence

src/controllers/userController.js
akosasante: Ah sorry for the confusion; `foo` is a placeholder value often used in programming; but I should've have made that clearer in context! https://kb.iu.edu/d/aetq#:~:text=In%20the%20world%20of%20computer,descriptions%2C%20and%20other%20computer%20documentation.

akosasante (approved):
LGTM

src/controllers/userController.js
NatachaKey: Didn’t know that, thank you !

akosasante:
(for these image upload ones, I _think_ we decided to hold off on merging them until all 3 are done and y'all can compare and learn from each and choose which one to merge into main)

akosasante (changes requested):
(doesn't actually need any changes; just marking it as such to prevent accidental merge)

PR #86, opened 2023-11-13
[Imgur] - Backend should be able to store profile pictures durably PS-215

DYA13:


src/public/imgur/imgur.js
akosasante, line 1: Why is this in the `public` folder? That is usually for scripts or files that we need to load in our server-rendered EJS files.

akosasante (commented):
I'm not sure how to test your code; are you able to test this in Postman? Could you point me to how you're running this code to test it out?

The js file is in a folder that doesn't get loaded by our usual express file and I'm not sure what port it gets loaded on.

Please refactor this to be loaded as a route from our `expressServer` so that its handled in a similar way to our other routes.

src/public/imgur/imgur.js
DYA13: I have done refactoring , moved ejs and all other code regarding imgur in imgur.js.   Before refactoring I have checked in Postman and I had status code 200 for uploading file( to be exact as per this commit 01e70627b1448534f3bdac9d6f027637eb67365d), but after refactoring I have 404 , route does not exist. And I can't find where I am mistaken, probably router or import , could you please have a look? 
And another strange thing if I check from Postman I have POST request ,If I check from localhost I have GET request ,I also can't find where I need to edit
 `2023-11-14 14:09:55 [INFO]: POST /api/v1/profile/65522d3fa0e6823344217146/uploadDB 404 6.413 ms - 20
2023-11-14 14:10:37 [INFO]: GET /api/v1/profile/65522d3fa0e6823344217146/uploadDB 404 1.566 ms - 20`

Sorry that I have too many console.log , I was looking for errors before

src/expressServer.js
```diff
@@ -18,11 +18,9 @@ const { readFileSync } = require('fs');
 const { join } = require('path');
 const expressStaticGzip = require('express-static-gzip');
 const recommendationRoutes = require('./routes/recommendationRoutes');
+const imgur = require('./views/imgur/imgur');
 
 const app = express();
-// Express Async Errors must be used before any route is used,
-// it will catch any errors that are thrown in the async functions without needing a try/catch
-require('express-async-errors');
```
akosasante: Any reason for deleting the comment + moving this line? 

It's fine since it doesn't change any behaviour of the app but just curious if it was an intentional change

src/expressServer.js
```diff
@@ -18,11 +18,9 @@ const { readFileSync } = require('fs');
 const { join } = require('path');
 const expressStaticGzip = require('express-static-gzip');
 const recommendationRoutes = require('./routes/recommendationRoutes');
+const imgur = require('./views/imgur/imgur');
```
akosasante: Here this is importing a javascript file from your `views` folder, `imgur.js`. That javascript file seems to be being used to define Express routes. If that's the case, we should have that file in our `routes` folder for easier discoverability and to make it clear to other developers what the functionality of that file is. We can also name it `imgurRoutes` to follow the pattern of other files in that folder.

While I like the idea of co-locating all the code that has to do with `imgur` in to a shared folder; one thing that makes me lean towards following the existing approach of having just `.ejs` files in the `views` folder and just routes files in the `routes` folder is that the `imgur` upload capability could potentially in the future be used for other features apart from the user profile. Like what if we want admins to be able to upload custom album art or something like that.

src/expressServer.js
```diff
@@ -100,7 +98,7 @@ app.set('view engine', 'ejs');
 app.use(favicon(__dirname + '/public/favicon.ico'));
 // Cookie parser middleware with JWT secret
 app.use(cookieParser(process.env.JWT_SECRET));
-
+app.use(express.static('public'));
```
akosasante: On line 139 there is already a line that says `app.use(express.static(`_**`__dirname`**_ `+ '/public'));`. That means that Express will always look in this path, `<the path where you have this backend repo stored>/src/public`, to find any files matching the current URL, _before_ it looks to match the current URL against any defined routes that come after line 139.

So for example if your browser requests `localhost:8000/my_special_file.js` , Express will look for a file in `<path to repo>/src/public/my_special_file.js` and return that to the browser. And this applies to anything after the `localhost:8000` part, not just Javascript files. 

But here on line 101 you've added a separate call to tell Express to also look for files in the `public` folder _not_ inside `src`. So now it's looking in two different places for arbitrary resources, which I think will cause confusion in the future and make things more difficult to debug.

src/expressServer.js
```diff
@@ -164,7 +162,7 @@ app.get('/tests', (req, res) => {
 // Error handling middleware (**MUST** be defined _after_ all other routes and middleware)
 app.use(notFoundMiddleware); // Not found middleware to handle invalid routes
 app.use(errorHandlerMiddleware); // Error handler middleware
-
+app.use('/api/v1/profile', imgur);
```
akosasante: Express reads the code in this file, from top to bottom. It will try to match the current URL to each line. You can explore this in the logs by enabling Express's debugger: [https://expressjs.com/en/guide/debugging.html](https://expressjs.com/en/guide/debugging.html) 

```
DEBUG=express:router npm run dev
```

This will add additional logging to the terminal to show what Express is doing. So now if you were to go to `localhost:8000/not_a_real_url.jpeg` in your terminal, you'll see that the request is handled by a bunch of middleware defined early in this file like `helmet`, then it eventually logs `express:router notFound  : /my_special_file.js +14ms` which means that it went top to bottom in this file, and the first route or middleware it found to handle this request that returns a response was `notFoundMiddleware`.

On the other hand, if we go to a route we know exists like `localhost:8000/api/v1/albums`, we see in the Express debug logs that the request goes through a bunch of middleware again, and then ends up printing `express:router router /api/v1/albums : /api/v1/albums +0ms` which shows that this route is being handled by the imported `albumRouter` on line 118. Since the route in the `albumRouter` file calls a controller function that returns a response `res.status(...).json(...)` that means Express knows it can return that response to the browser and it doesn't need to go through any of the rest of the routes or middlewares defined in this file like the `notFoundMiddleware` or the `errorHandlerMiddleware`

---

In your case, you have your `api/v1/profile` router defined _after_ the `notFoundMiddleware` and the `errorHandlerMiddleware`. That means Express is going to get to the `notFoundMiddleware` _**first.**_ The not-found middleware matches _any_ route and will return a 404 response. Once a response is defined, Express does not try to call any routes defined later in the file. 

To fix this, you should have your `app.use` router defitinition _before_ the `notFoundMiddleware`. This is the reason why we've grouped all of our routes together from line 116 - 122 in this file. And even our one-off routes like our `/health` route or our routes for EJS files or for API docs, all happen earlier in this file _before_ the `notFoundMiddleware`. 

**conclusion:** You should put the `app.use` line under the `chatRoutes` line (or anywhere in the file _before_ `notFoundMiddleware` is defined.

src/views/imgur/imgur.js
akosasante, line 1: This is fine as-is; but one consideration I would make; is since the code in this route is is specifically written for uploading a profile image for a user; it might make more sense to define this under `api/v1/users/:userId/uploadProfile` under the existing user routes. Rather than having a separate route for just a single api endpoint.

src/views/imgur/imgur.js
```diff
@@ -0,0 +1,72 @@
...
+const imgurClient = new ImgurClient({
+  clientId: process.env.IMGUR_CLIENT_ID,
+  clientSecret: process.env.IMGUR_CLIENT_SECRET,
+  refreshToken: process.env.IMGUR_REFRESH_TOKEN,
+});
+
+router.post(
+  '/:userId/uploadDB',
+  fileUploadMiddleware({ limits: { fileSize: 10000000 }, abortOnLimit: true }),
+  async (req, res) => {
```
akosasante: This is fine for the initial work here; but for production I would expect this code to follow our current convention in this repo; where the code inside the async request handler function (line 16-69) is in a controller file, so that the controller can be more easily unit tested separate from the routes definition+middleware.

src/views/imgur/imgur.js
```diff
@@ -0,0 +1,72 @@
...
+    console.log('File MIME Type:', profile.mimetype);
+    console.log('File Size:', profile.size);
+
+    if (!profile || !/^image/.test(profile.mimetype)) {
+      return res.status(400).send('File is not an image.');
+    }
+    if (!req.files || !req.files.profile) {
+      console.log('No profile file in the request.');
+      return res.status(400).send('No profile file in the request.');
+    }
```
akosasante: Do we need to keep both this _and_ essentially the same check on line 41-43?

src/views/imgur/imgur.js
```diff
@@ -0,0 +1,72 @@
...
+      title: `${user._id}_profile_pic`,
+    });
+    if (imgurRes.success) {
+      user.profileImage = {
+        url: imgurRes.data.link,
+        altText: 'user profile picture',
+      };
+    }
+    res.status(200).send('Updated profile picture');
+    res.render('../views/imgur', { userId: req.params.userId });
```
akosasante: Express will not let you return two responses. You'll eventually see an error logged in the backend terminal like: `2023-11-19 12:29:49 [ERROR]: Cannot set headers after they are sent to the client - Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client` 

Here you are returning a response on line 67 with status of 200 and the text "Updated profile picture". And then you're _also_ trying to return an HTML response for the browser to render a page, defined by an EJS view. But that won't work because you've already told Express to respond to the browser with some text.

akosasante: `res.render` looks in the views folder automatically, so we do not need to pass a relative path into it (and in fact, by doing this, Express won't be looking in the right place for the views files). 

[https://expressjs.com/en/5x/api.html#res.render](https://expressjs.com/en/5x/api.html#res.render)

[https://www.geeksforgeeks.org/express-js-res-render-function/](https://www.geeksforgeeks.org/express-js-res-render-function/)

If you take a look in the repo for the other places where we call `res.render`, you'll notice, we give Express just the name of the EJS file. Eg: `res.render('orderNotifications')` will mean that Express will look for a file in the path `<backend repo>/src/views/orderNotifications.ejs`. 

So if we do want to indeed render an HTML page, then we should remove line 67, and update this line to be:

```suggestion
    res.status(200).send('Updated profile picture');
    res.render('imgur/imgur', { userId: req.params.userId });
```

(note that I had to do `imgur/imgur` because our EJS file is in a folder _inside_ `views/imgur`).

src/views/imgur/imgur.ejs
akosasante, line 1: This looks like a page where the user would go to upload their image. Am I understanding correctly? What URL should the user be typing to get to this page?

Right now you have this page being rendered as a _response_ to the API endpoint that handles uploading the profile picture, which I'm not sure makes sense. But how would I get here in the first place? There should be a `GET` route defined in the application so I can type in a URL that would load up this page.

src/expressServer.js
DYA13: No, it wasn't intentional change, corrected commit 9cb630ca70d7b497bcbc7be339804d0192b7a2cc

src/expressServer.js
DYA13: ImgurRoutes.js created and relocation of files done , commit 749d1e07510f8835388f0e1f8989f97383a7be79, commit 39723399eaf86ce4f2c099e07370da4d1e1cf18d

src/expressServer.js
DYA13: Removed separate call to public commit  38f8ea507d8892ec277bdd098227947c2d237b3c

src/expressServer.js
DYA13: Thank you very much , it is corrected , moved before the notFoundMiddleware is defined, commit d669580ef4c2171bb3476035eba87b46efdc642b

src/views/imgur/imgur.js
DYA13: edited , commit 94a664d059c981855d4d2c8e753744f63008b021

src/expressServer.js
```diff
@@ -1,5 +1,5 @@
 require('dotenv').config();
-
+require('express-async-errors');
```
akosasante: This line can be removed since we already require `express-async-errors` in this file on line 27

src/routes/imgurRoutes.js
```diff
@@ -0,0 +1,73 @@
...
+      title: `${user._id}_profile_pic`,
+    });
+    if (imgurRes.success) {
+      user.profileImage = {
+        url: imgurRes.data.link,
+        altText: 'user profile picture',
+      };
+    }
+    res.status(200).send('Updated profile picture');
+    res.render('../views/imgur', { userId: req.params.userId });
```
akosasante: This is still trying to return 2 responses; one plain text response and one HTML/EJS response; which won't work per previous comment

src/routes/imgurRoutes.js
```diff
@@ -0,0 +1,73 @@
...
+const User = require('../models/User');
+
+const router = express.Router();
+const imgurClient = new ImgurClient({
+  clientId: process.env.IMGUR_CLIENT_ID,
+  clientSecret: process.env.IMGUR_CLIENT_SECRET,
+  refreshToken: process.env.IMGUR_REFRESH_TOKEN,
+});
+
+userRoutes.post(
```
akosasante: If we're defining this on the `userRoutes` then let's put this file in the `userRoutes` file instead of spreading across two files?

src/routes/imgurRoutes.js
```diff
@@ -0,0 +1,73 @@
+const userRoutes = require('./userRoutes');
+const express = require('express');
+const fileUploadMiddleware = require('express-fileupload');
+const { ImgurClient } = require('imgur');
+const User = require('../models/User');
+
+const router = express.Router();
```
akosasante: Here you define a router. But in this file you're not adding any routes to this variable (instead you're defining `.post` on the `userRoutes` router). And then you export this `router` and tell Express to use it for anything starting with `/api/v1/profile` (in `expressServer.js`). But since you've not added any routes to this `router` variable, that means that there's nothing attached to `/api/v1/profile`

src/routes/imgurRoutes.js
akosasante: Since you've defined this route on `userRoutes`, that means that Express only knows to look for `/:userId/uploadProfile` when it's after `api/v1/users`. 

I think that's totally fine, but in that case, this doesn't need to be here in a separate file, we should instead be defining this alongside the rest of the `api/v1/users` routes in the `userRoutes.js` file

src/views/imgur/imgur.js
DYA13: Refactoring done , commit b7a89e1e90807bfff900463307f888d0f4ce4348

src/views/imgur/imgur.js
DYA13: I left only one check 👍
` if (!profile) {
    console.log('No profile file in the request.');
    return res.status(400).send('No profile file in the request.');
  }`

src/views/imgur/imgur.js
DYA13: I have deleted the plain text response , commit https://github.com/Code-the-Dream-School/dd-prac-team4-back/commit/ba4adec530fe34b6d22e428f759d95c8376d4fb3

DYA13: thank you, updated the path , commit 3df974401a87024e46bdebb7e8f7f6491e58f9b5

src/views/imgur/imgur.ejs
DYA13: I did some modifications, could you please check if it is correct? Or may be there is another way to do it?  commit da183096058dc03dda2b9fcc0ccc2ce01038ea26

src/expressServer.js
DYA13: Deleted 👍, commit 690de2ede2bcf1ff5d191ae45108c75b09ce9520

src/routes/imgurRoutes.js
DYA13: I have deleted the plain text response , commit ba4adec530fe34b6d22e428f759d95c8376d4fb3

src/routes/imgurRoutes.js
DYA13: imgur routes moved to UserRoutes, commit 0cb57b38cc269ee57c85dcee133abe92f8bfcf10

src/routes/imgurRoutes.js
DYA13: Akos , I have made some modifications but I am not sure if it correct. Could you check please?

src/expressServer.js
```diff
@@ -18,8 +17,9 @@ const { readFileSync } = require('fs');
 const { join } = require('path');
 const expressStaticGzip = require('express-static-gzip');
 const recommendationRoutes = require('./routes/recommendationRoutes');
-
+const userRouter = require('./routes/userRoutes');
```
akosasante: Not a big deal; but any reason this was moved from being in the "// IMPORT ROUTERS" section of the code under line 117 to up here? It generally makes sense to keep like things together.

src/expressServer.js
```diff
@@ -135,6 +134,7 @@ app.use('/api/v1/reviews', reviewRouter /* #swagger.tags = ['Reviews'] */);
 app.use('/api/v1/wishlist', wishlistRoutes /* #swagger.tags = ['Wishlist'] */);
 app.use('/api/v1/chat', chatRoutes);
 app.use('/api/v1/recommendations', recommendationRoutes);
+app.use('/api/v1/profile', userRouter);
```
akosasante: If we will use the `userRouter`, I think it's fine to define our endpoint where a user can upload their profile picture, on `api/v1/user` . So I don't think we need this separate `api/v1/profile` route in that case and this line can be deleted

src/routes/userRoutes.js
```diff
@@ -43,4 +44,13 @@ router
   .route('/:id')
   .delete(authenticateUser, authorizePermissions('admin'), deleteSingleUser);
 
+// Route for rendering the upload page
+router.get('/:userId/uploadProfile', imgurController.renderUploadPage);
```
akosasante: Looks good! Do we have a function called `renderUploadPage` being exported from that controller yet ? Currently this is erroring because Express can't find that.

src/views/imgur.ejs
```diff
@@ -0,0 +1,19 @@
...
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Image Upload</title>
+  </head>
+  <body>
+    <h1>Image Upload</h1>
+    <form
+      action="/api/v1/profile/<%= userId %>/uploadDB"
```
akosasante: This should point to what we have in `userRoutes`:

```suggestion
      action="/api/v1/user/<%= userId %>/uploadProfile"
```

src/controllers/imgurController.js
```diff
@@ -0,0 +1,63 @@
...
+
+  if (imgurRes.success) {
+    user.profileImage = {
+      url: imgurRes.data.link,
+      altText: 'user profile picture',
+    };
+    await user.save();
+  }
+
+  res.render('imgur', { userId: req.params.userId });
```
akosasante: Just verifying, that what you intend here is that if the request was successful, we have the prowser re-render the upload image page?

src/expressServer.js
DYA13: const recommendationRoutes = require('./routes/recommendationRoutes')  moved under // IMPORT ROUTERS section

src/expressServer.js
DYA13: Deleted 👍, commit 4384e65b8ff7353a719510708473a81665ed6585

src/routes/userRoutes.js
DYA13: No, we don't have this function 😊, we have uploadProfile in imgurcontroller

src/controllers/imgurController.js
DYA13: I was expecting that after a successful image upload, the page is re-rendered, displaying the updated user information or profile picture on the 'imgur' page

src/controllers/imgurController.js
```diff
@@ -0,0 +1,63 @@
...
+const User = require('../models/User');
+
+const imgurClient = new ImgurClient({
+  clientId: process.env.IMGUR_CLIENT_ID,
+  clientSecret: process.env.IMGUR_CLIENT_SECRET,
+  refreshToken: process.env.IMGUR_REFRESH_TOKEN,
+});
+
+async function uploadProfile(req, res) {
+  console.log('Received request for /api/v1/profile/:userId/uploadProfile');
```
akosasante:
```suggestion
  console.log('Received request for /api/v1/user/:userId/uploadProfile');
```

src/routes/userRoutes.js
akosasante: Are you planning on adding `renderUploadPage` in this PR or in a future one? Currently since this value doesn't exist, it will cause errors that cause the server not to be able to start up. If we will apply that in a future PR we can comment it out here for now.

This is the error I get when I run `npm run dev` and it points to this line:

```
Error: Route.get() requires a callback function but got a [object Undefined]
    at Route.<computed> [as get] (/Users/aasante/dev/practicum2023/backend/node_modules/express/lib/router/route.js:211:15)
    at Function.proto.<computed> [as get] (/Users/aasante/dev/practicum2023/backend/node_modules/express/lib/router/index.js:521:19)
    at Object.<anonymous> (/Users/aasante/dev/practicum2023/backend/src/routes/userRoutes.js:48:8)
    at Module._compile (node:internal/modules/cjs/loader:1103:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1157:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/Users/aasante/dev/practicum2023/backend/src/expressServer.js:122:20)
[nodemon] app crashed - waiting for file changes before starting...
```

akosasante: Personally, I think it would make sense to go ahead and create the matching controller function in this PR since you already have the EJS file and everything else ready.

DYA13: Agree, we can do it within this PR , I have added these lines of code, possibly needs correction, could you check please? commit c9734635973e12cf2dc6a76b51630d1700283fed

src/controllers/imgurController.js
```diff
@@ -0,0 +1,68 @@
...
+const User = require('../models/User');
+
+const imgurClient = new ImgurClient({
+  clientId: process.env.IMGUR_CLIENT_ID,
+  clientSecret: process.env.IMGUR_CLIENT_SECRET,
+  refreshToken: process.env.IMGUR_REFRESH_TOKEN,
+});
+
+const renderUploadPage = (req, res) => {
+  res.render('imgur');
```
akosasante: We need to pass the EJS page the userId here:

src/views/imgur.ejs
```diff
@@ -0,0 +1,19 @@
...
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Image Upload</title>
+  </head>
+  <body>
+    <h1>Image Upload</h1>
+    <form
+      action="/api/v1/user/<%= userId %>/uploadProfile"
```
akosasante: Sorry, actually looking at `expressServer.js` closer I see this should be "users" plural:

```suggestion
      action="/api/v1/users/<%= userId %>/uploadProfile"
```

src/controllers/imgurController.js
```diff
@@ -0,0 +1,68 @@
...
+
+  if (imgurRes.success) {
+    user.profileImage = {
+      url: imgurRes.data.link,
+      altText: 'user profile picture',
+    };
+    await user.save();
+  }
+
+  res.render('imgur', { userId: req.params.userId });
```
akosasante: It might be nice to pass in a variable like `message`, so that we can show when the page reloads some text that says "Image uploaded successfully" or something. Or you could even pass back the image link so that we can show the image on the page once successful :)

src/routes/userRoutes.js
akosasante: A small nitpick. So while this works (after making the fixes I've noted in this PR, I was able to navigate to http://localhost:8000/api/v1/users/64ef5e90086e1c19b888c72c/uploadProfile and see the EJS page), it's a little awkward to go to a `/api` URL to view a page. All our `/api` routes should be for retruning API/JSON data.

What we've done with the other EJS files for now, is just have a standalone route in `expressServer`. So just for this `GET` route, I would suggest moving it to area of the code: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/e8fb9a522bfdf8cdbb502c62e09e89dea6581dae/src/expressServer.js#L143-L150

src/controllers/imgurController.js
akosasante: Another idea, if you feel like trying your hand at it; is to return a JSON object here instead of re-rendering the page.

Then, in the `imgur.ejs` file you would add a `<script>` tag. 

Then in the `src/public` folder you would create a javascript file that the `script` tag should poin tto . This javascript file is where you would handle submitting the form (instead of letting the default DOM behaviour handle it like you have right now). 

[https://matiashernandez.dev/blog/post/handling-forms-with-vanilla-javascript](https://matiashernandez.dev/blog/post/handling-forms-with-vanilla-javascript)

The main advantage of that is that you can add custom logic, like showing a Toast/Alert bar when the request is successful; or dynamically showing the uploaded image along with a message. If you want to see an example of showing a toast you can look at the orderNotification page: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/7c9b523158abe7f8e952de14b3bbdca95d4584cf/src/public/orderNotifications.js#L20-L31](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/7c9b523158abe7f8e952de14b3bbdca95d4584cf/src/public/orderNotifications.js#L20-L31)

If you want to see an example of submitting a form via javascript, you can look at the admin album update page: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/7c9b523158abe7f8e952de14b3bbdca95d4584cf/src/public/admin.js#L7

src/views/imgur.ejs
DYA13: I have noticed as well in previous comment, corrected commit 2127735a8baeed895a8a5f80939c8e241567cf1f

src/controllers/imgurController.js
DYA13: Could you please check if I have done correctly? commit d15de3de677b6e581a48610a569ff1c22ea95889

src/routes/userRoutes.js
DYA13: I have moved the route in expressServer 👍 , could you please check if it is correct? commit 27422acf5ecac14ca9f9edfe89abcb65694e369f

src/controllers/imgurController.js
DYA13: Thank you very much for your suggestions 👍, as for the message and image link , implemented in this commit 676f52ad0aef21e29a81f7d3703441773c9ad82c

DYA13: > Another idea, if you feel like trying your hand at it; is to return a JSON object here instead of re-rendering the page.
> 
> Then, in the `imgur.ejs` file you would add a `<script>` tag. 
> 
> Then in the `src/public` folder you would create a javascript file that the `script` tag should poin tto . This javascript file is where you would handle submitting the form (instead of letting the default DOM behaviour handle it like you have right now). 
> 
> https://matiashernandez.dev/blog/post/handling-forms-with-vanilla-javascript
> 
> The main advantage of that is that you can add custom logic, like showing a Toast/Alert bar when the request is successful; or dynamically showing the uploaded image along with a message. If you want to see an example of showing a toast you can look at the orderNotification page:
> 
> https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/7c9b523158abe7f8e952de14b3bbdca95d4584cf/src/public/orderNotifications.js#L20-L31
> 
> If you want to see an example of submitting a form via javascript, you can look at the admin album update page:
> 
> https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/7c9b523158abe7f8e952de14b3bbdca95d4584cf/src/public/admin.js#L7

Akos could you please check my attempt to realize the suggestion regarding to  the implementation of  a JSON object  instead of re-rendering the page, adding custom logic , toastify?   commit 588889e6ebaa1a5e20188da1695e3e159ba2bc59

src/expressServer.js
akosasante, line 1: 👍🏾

src/views/imgur.ejs
```diff
@@ -0,0 +1,20 @@
...
+    <h1>Image Upload</h1>
+    <form
+      action="/api/v1/users/<%= userId %>/uploadProfile"
+      method="post"
+      enctype="multipart/form-data"
+    >
+      <input type="file" name="profile" accept="image/*" required />
+      <button type="submit">Upload Image</button>
+    </form>
+    <script src="uploadHandler.js"></script>
```
akosasante:
```suggestion
    <script src="/uploadHandler.js"></script>
```

The forward slash `/` tells the browser to try to request `uploadHandler.js` from the root URL (so in the local dev case, from `http://localhost:8000/uploadHandler.js`). In the server, Express will see that request and will see that there are no routes that match so it will then check the `public` folder for a file called `uploadHandler.js` and find it and return it to the browser.

Screenshot 2023-11-23 at 9 32 01 AM

But without the forward slash, the browser instead will instead just try to add `uploadHandler.js` to the current url pattern, so it will try to ask the server for: `http://localhost:8000/:userId/uploadHandler.js`. In the server, Express will look for a directory in the `public` folder with the name `:userId` and then an `uploadHandler.js` file inside that. But since we don't have that, this will not work of course.

Screenshot 2023-11-23 at 9 34 14 AM

src/public/uploadHandler.js
```diff
@@ -0,0 +1,33 @@
+/* global Toastify */
+document.addEventListener('DOMContentLoaded', function () {
+  const form = document.querySelector('form');
+  const userId = document.querySelector('#userId').value;
```
akosasante: There is no element in the `imgur.ejs` file with an id of `userId` so I don't think this line will find anything. 
`document.querySelector` can only be used to find elements in the DOM (in the HTML document).

If we want to get information from the current URL, we can use the browser function: `window.location`, specifically, we can use `window.location.pathname` to get the value of the URL after the base url: https://developer.mozilla.org/en-US/docs/Web/API/Location/pathname

So for `http://localhost:8000/64ef5e90086e1c19b888c72c/uploadProfile`, `window.location.pathname` will return `/64ef5e90086e1c19b888c72c/uploadProfile`. We can then use the string split function to split that string by the `"/"` character and then grab the value that has the id: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split

Give it a try and feel free to explore the `window.location` function in the browser dev tools to figure out how to get the user id that we need here.

src/controllers/imgurController.js
```diff
@@ -0,0 +1,83 @@
+const { ImgurClient } = require('imgur');
+const User = require('../models/User');
+
+const imgurClient = new ImgurClient({
+  clientId: process.env.IMGUR_CLIENT_ID,
+  clientSecret: process.env.IMGUR_CLIENT_SECRET,
+  refreshToken: process.env.IMGUR_REFRESH_TOKEN,
+});
+
+const renderUploadPage = (req, res, message, imageUrl) => {
```
akosasante: All controller functions can only take two values, `req`, and `res`. Otherwise, Express won't be able to handle it. If we need to pass in `message`, and `imageUrl` we should be passing it in via params, query parameters, or a request body.

```suggestion
const renderUploadPage = (req, res) => {
```

src/controllers/imgurController.js
```diff
@@ -0,0 +1,83 @@
...
+  const userId = req.params.userId;
+  res.render('imgur', { userId, message, imageUrl });
```
akosasante: When the user first loads the page by just navigating to `http://localhost:8000/:userId/uploadImage` they won't have a message or image url yet because they haven't yet tried to upload anything. So we can just set those variables to be empty:

```suggestion
  res.render('imgur', { userId, message: null, imageUrl: null });
```

src/controllers/imgurController.js
```diff
@@ -0,0 +1,83 @@
...
+      };
+      await user.save();
+
+      // Render with success message and image link
+      renderUploadPage(
+        req,
+        res,
+        'Image uploaded successfully!',
+        imgurRes.data.link
+      );
```
akosasante: This result is being handled by the `fetch` call that we're doing in `uploadHandler.js`. That function is expecting a JSON response (see on line 20 of `uploadHandler.js` where you're doing `const data = await response.json()`). But here you're returning an HTML/EJS page. So that's going to result in an error if you check the dev console, where it gets a value that starts with `!<doctype html` but it tries to handle it as a JSON.

Screenshot 2023-11-23 at 9 52 29 AM

In this approach we don't need to re-render the imgur page, since we are already on that page. We just want to get information via JSON (with our fetch AJAX request) and in our javascript file (`uploadHandler.js`) we want to decide ourselves how to handle it instead of letting the browser's default behaviour take over.

So this should return a json in the usual way any of our other controller handlers would.

src/views/imgur.ejs
akosasante: We also need to add `<script src="/toastify/toastify.js"></script>` in order to use the Toastify library. See how it was done in the `orderNotifications.ejs` file for example.

Otherwise we will see the following error in the dev console:
Screenshot 2023-11-23 at 9 57 39 AM

src/public/uploadHandler.js
```diff
@@ -0,0 +1,33 @@
...
+        '/api/v1/users/' + userId + '/uploadProfile',
+        {
+          method: 'POST',
+          body: formData,
+        }
+      );
+
+      if (response.ok) {
+        const data = await response.json();
+
```
akosasante, line 22: currently you're not doing anything with the `imageUrl` that should be returned in the response. 

You can use DOM functions to append an `img` element to the page when the response is successful, and set the `src` attribute of that element to the imageUrl that the server returned.

You may need to update the `app.use('helmet')` line in Express to tell the server that it's okay to show images from `imgur` (cross-site content):

```
app.use(helmet({
  contentSecurityPolicy: {
    directives: {
      "img-src": ["'self'", "https://i.imgur.com"]
    }
  }
}));
```

src/public/uploadHandler.js
DYA13: You are right, I was sure it won't work but I didn't know what can be the solution, thank you very much ❤️ 
I am not sure but assume that user ID will be the second element (index1) here
 function getUserIdFromURL() {
    const pathArray = window.location.pathname.split('/');
    return pathArray[1];
  }
commit 8e7d231c585c469a8c9bbae727cfd995ad912d8d

src/controllers/imgurController.js
DYA13: Akos , could you please check? commit 311fd8c5230019e23d6933c28811a08f81566a89

src/views/imgur.ejs
DYA13: Done , commit 5e52aef09b122b0ffe056489ec2c252ef7832e77

src/public/uploadHandler.js
DYA13: could you please check?
 commit  acaff4dbd58217f9489eabf637b86280f11d49ab

src/public/uploadHandler.js
```diff
@@ -0,0 +1,45 @@
+/* global Toastify */
+document.addEventListener('DOMContentLoaded', function () {
+  const form = document.querySelector('form');
+
+  form.addEventListener('submit', async function (event) {
+    event.preventDefault();
+
+    const userId = getUserIdFromURL();
```
akosasante, line 8: :star:  love the refactoring here into helper functions!

akosasante (approved):
Looks and works fantastically! Great job Yelena 🎉

akosasante (changes requested):
No further changes needed, but I am marking it this way so that we don't accidentally merge until all 3 approaches are complete and we can compare and decide which way we want to finalize and proceed with 👍🏾

akosasante:
Opting to go with #87 for now per convo in Slack -- though this may still be used as a reference. Great work @DYA13 !

PR #87, opened 2023-11-18
firebase connected

mvstoyan:


package.json
```diff
@@ -29,17 +30,20 @@
...
     "express-xss-sanitizer": "^1.1.6",
+    "firebase": "^10.6.0",
     "helmet": "^7.0.0",
     "http-status-codes": "^2.2.0",
     "jsonwebtoken": "^9.0.1",
     "logger": "^0.0.1",
     "mongo-sanitize": "^1.1.0",
     "mongoose": "^7.4.1",
     "morgan": "^1.10.0",
+    "multer": "^1.4.5-lts.1",
```
akosasante: My understanding is that `multer` is another library for handling file uploads ("multi-part uploads). I'm not sure that we need to have _both_ express-fileupload _and_ multer. Did you find that you had separate needs for both libraries?

src/expressServer.js
akosasante, line 1: Hm, maybe we don't need to import/export it from this file; if nothing in this file is using `firebaseApp` and currently the one place you require the variable, (`userController.js`), you're just directly importing it from the `firebase.js` file anyway

src/firebase.js
akosasante, line 1: Could you share the env values you're using to test this with me in Slack please?

src/routes/userRoutes.js
```diff
@@ -43,4 +44,7 @@ router
   .route('/:id')
   .delete(authenticateUser, authorizePermissions('admin'), deleteSingleUser);
 
+router.route('/updateProfileImage').patch(authenticateUser, updateProfileImage);
```
akosasante: You must also use the `express-file-upload` middleware here for this route, in order for Express to be able to handle receiving the file via a `form-data` HTTP request, and in order for it to be put into the `req.files` variable.

akosasante:
```javascript
const fileUploadMiddleware = require('express-fileupload');
```

```suggestion
router.route('/updateProfileImage').patch(authenticateUser, fileUploadMiddleware({ limits: { fileSize: 10000000 }, abortOnLimit: true }), updateProfileImage);
```

src/controllers/userController.js
```diff
@@ -168,6 +177,20 @@ const getCurrentUserWithPurchasedAlbums = async (req, res) => {
   */
 };
 
+const updateProfileImage = async (req, res) => {
+  console.log('Updating profile image.');
+  const firebaseStorage = getStorage(firebaseApp);
+  const user = req.user;
+  const imageRef = ref(firebaseStorage, `images/${user._id}_profile_pic`);
```
akosasante: When `authenticateUser` middleware initializes the `req.user` variable, it puts the id in a `userId` property rather than `_id` : [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/9f81dd657a81ecb96b06946e251aaae8cf26152f/src/middleware/authentication.js#L16](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/9f81dd657a81ecb96b06946e251aaae8cf26152f/src/middleware/authentication.js#L16)

```suggestion
  const imageRef = ref(firebaseStorage, `images/${user.userId}_profile_pic`);
```

src/controllers/userController.js
```diff
@@ -168,6 +177,20 @@ const getCurrentUserWithPurchasedAlbums = async (req, res) => {
...
+  const { profile } = req.files;
+  await uploadBytes(imageRef, profile.data, {
+    contentType: profile.mimetype,
+  });
+  const imageUrl = await getDownloadURL(imageRef);
```
akosasante: Unless we want the client/browser to have to make a second call to the backend to update the user object with this URL, we should go ahead here and save this profile image link on the User object for them

src/firebase.js
```diff
@@ -0,0 +1,17 @@
+const firebase = require('firebase/app');
+require('firebase/auth');
+
+const firebaseConfig = {
+  apiKey: process.env.FIREBASE_API_KEY,
+  authDomain: process.env.FIREBASE_AUTH_DOMAIN,
+  projectId: process.env.FIREBASE_PROJECT_ID,
+  storageBucket: process.env.FIREBASE_STORAGE_BUCKET,
+  messagingSenderId: process.env.FIREBASE_MESSAGING_SENDER_ID,
```
akosasante: I don't think we're using anything that requires this messaging sender id value?

akosasante (commented):
Apologies, I just realized there were some leftover references to 'imgur' in the firebase doc. I've updated it now to fix that: https://codethedreamteam4-2023.atlassian.net/wiki/spaces/PT4/pages/10551297/Store+images+in+an+external+service+-+Firebase

src/firebase.js
mvstoyan: sent in Slack

akosasante (commented):
After testing with your firebase api credentials; I think you may need to adjust the security rules on your firebase storage console to allow writes from any requesters:



Later on, for security reasons we can add Firebase [AppCheck](https://firebase.google.com/docs/app-check/web/recaptcha-provider?hl=en&authuser=2&_gl=1*167rah9*_ga*MTE2MzM0ODYzNi4xNjk4MDg2MDk3*_ga_CW55HF8NVT*MTcwMDQ0MTgwMi41LjEuMTcwMDQ0MjYyOC4xNi4wLjA.) to ensure that bad actors can't randomly also upload files directly into our firebase data store

package.json
mvstoyan: I'm looking Firebase and thought I might need to control photo sizes. I found express-fileupload doesn't do that, so I installed multer. I removed multer

src/firebase.js
mvstoyan: Removed, I just copied the whole document from project settings.

mvstoyan:
> After testing with your firebase api credentials; I think you may need to adjust the security rules on your firebase storage console to allow writes from any requesters:
> 
> 
> 
> Later on, for security reasons we can add Firebase [AppCheck](https://firebase.google.com/docs/app-check/web/recaptcha-provider?hl=en&authuser=2&_gl=1*167rah9*_ga*MTE2MzM0ODYzNi4xNjk4MDg2MDk3*_ga_CW55HF8NVT*MTcwMDQ0MTgwMi41LjEuMTcwMDQ0MjYyOC4xNi4wLjA.) to ensure that bad actors can't randomly also upload files directly into our firebase data store

I made a reCAPTCHA key, I'll try to do it.

src/controllers/userController.js
```diff
@@ -168,6 +177,31 @@ const getCurrentUserWithPurchasedAlbums = async (req, res) => {
...
+  const profile = req.files.profile;
+  if (!profile || !profile.data || !profile.mimetype) {
+    throw new CustomError.BadRequestError('File is not an image.');
+  }
+  const imageRef = ref(firebaseStorage, `images/${userId}_profile_pic`);
+  await uploadBytes(imageRef, profile.data, {
+    contentType: profile.mimetype,
+  });
+  const imageUrl = await getDownloadURL(imageRef);
+  const updatedUser = await User.findByIdAndUpdate(
```
akosasante: We'll get a lint warning here about `updatedUser` variable not being used. We can either make the function call without storing the result in a variable; or we can prefix the variable with an underscore, which lets ESLint know that we are intentionally not using this variable later in the function (based on this ESLint configuration: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/7c9b523158abe7f8e952de14b3bbdca95d4584cf/.eslintrc.json#L8](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/7c9b523158abe7f8e952de14b3bbdca95d4584cf/.eslintrc.json#L8)

```suggestion
    await User.findByIdAndUpdate(
```

or  
 

```suggestion
  const _updatedUser = await User.findByIdAndUpdate(
```

src/firebase.js
```diff
@@ -0,0 +1,17 @@
...
+const firebaseConfig = {
+  apiKey: process.env.FIREBASE_API_KEY,
+  authDomain: process.env.FIREBASE_AUTH_DOMAIN,
+  projectId: process.env.FIREBASE_PROJECT_ID,
+  storageBucket: process.env.FIREBASE_STORAGE_BUCKET,
+  appId: process.env.FIREBASE_APP_ID,
+};
+
+const firebaseApp = initializeApp(firebaseConfig);
+const firebaseStorage = getStorage();
```
akosasante: It's up to you, but one thing we could do is call 

```suggestion
const firebaseStorage = getStorage(firebaseApp);
```

here and then export it; and then in the controller, we can just use `firebaseStorage` without having to do the step of "get storage" for each request. But I'm not sure if that really optimizes much (that is, I'm not sure if the library is making a full HTTP call to Firebase to get this information or what exactly); so I'm fine leaving the call in the controller as-is otherwise.

src/routes/userRoutes.js
```diff
@@ -43,4 +45,13 @@ router
   .route('/:id')
   .delete(authenticateUser, authorizePermissions('admin'), deleteSingleUser);
 
+router.route('/:id/updateProfileImage').post(
```
akosasante: I think you had this route as only `'/updateProfileImage'` before rather than prefixed with the user id. I think that makes sense because we're getting the logged-in user from `authenticateUser`, so there's no reason to make the client _also_ include the user's id in the URL path. Plus it opens up the possibility of having to handle if the caller accidentally (or maliciously) puts an invalid id there or another user's id in the URL.

akosasante: And I don't see the code in the controller using it (`req.params.id`) anyway so it should be safe to just remove)

package.json
akosasante: Ah I see. Yeah let's remove for now but we can adjust later if needed. I think Firebase also has its own extensions that we could potentially use on their side for things like images.

akosasante (commented):
Nice! I was able to upload an image with your routes and see it in my user profile with your front end branch 👍🏾

Question, are you using the RECAPTCHA\_KEY env value anywhere yet for the AppCheck part?

mvstoyan:
> Nice! I was able to upload an image with your routes and see it in my user profile with your front end branch 👍🏾
> 
> Question, are you using the RECAPTCHA_KEY env value anywhere yet for the AppCheck part?

I was trying to do tickets 165/208. I've just returned to reCAPTCHA Enterprise. I need to understand how to integrate it into the code. I have only generated the key that was sent earlier in Slack so far.

akosasante:
Gotcha, sounds good.   
Once you get to this, it should be able to follow the steps outlined in these docs for setting up AppCheck when initializing the Firebase app: [https://firebase.google.com/docs/app-check/web/recaptcha-provider](https://firebase.google.com/docs/app-check/web/recaptcha-provider)

src/routes/userRoutes.js
mvstoyan: Yes, the previous commit was without :id, I added it later. Now I've checked, and it works fine without :id, so I removed it. Thank you!

mvstoyan:
> Gotcha, sounds good.  Once you get to this, it should be able to follow the steps outlined in these docs for setting up AppCheck when initializing the Firebase app: https://firebase.google.com/docs/app-check/web/recaptcha-provider

I generated a key for reCAPTCHA Enterprise instead of  reCAPTCHA v3. https://firebase.google.com/docs/app-check/web/recaptcha-enterprise-provider. I'm having trouble understanding the logic and where to start. I assume that it makes sense to connect App Check on the client side(our React frontend). I'm keep looking...

akosasante:
> I generated a key for reCAPTCHA Enterprise instead of reCAPTCHA v3. [firebase.google.com/docs/app-check/web/recaptcha-enterprise-provider](https://firebase.google.com/docs/app-check/web/recaptcha-enterprise-provider). 

Aha yes sorry I missed that those were separate docs. The ones you linked should work.

> I'm having trouble understanding the logic and where to start. I assume that it makes sense to connect App Check on the client side(our React frontend). I'm keep looking...

It's possible I'm missing something since I have not used this part of Firebase before, but I think we should be adding App Check in the backend, since that's where we are making calls to Firebase from in our case. If we were making the Firebase calls on the client-side, _then_ we would add the App Check credentials there.

From the docs, you've already completed the first part of [step 1](https://firebase.google.com/docs/app-check/web/recaptcha-enterprise-provider#project-setup) (`1. Set up your Firebase project`) since I know we have a Firebase project :white\_check\_mark: 

Then it asks us to create a reCAPTCHA Enterprise key which I believe you're already completed, with a type of "Website" (rather than ios or android app), here: [https://console.cloud.google.com/security/recaptcha?project=\_.](https://console.cloud.google.com/security/recaptcha?project=_.)

My understanding is that for our use-case we can skip the parts shown in the reCAPTCHA console about adding a script to the frontend.

You've already completed [Step 2](https://firebase.google.com/docs/app-check/web/recaptcha-enterprise-provider) since you have a `firebase.js` file where you do things like `initializeApp` :white\_check\_mark: 

Then in [Step 3](https://firebase.google.com/docs/app-check/web/recaptcha-enterprise-provider#initialize) is where we need to add `firebase/app-check` code to our `firebase.js` per the example shown in that section.

That should be all that's needed to get it set up and activated for now.

mvstoyan:
> > I generated a key for reCAPTCHA Enterprise instead of reCAPTCHA v3. [firebase.google.com/docs/app-check/web/recaptcha-enterprise-provider](https://firebase.google.com/docs/app-check/web/recaptcha-enterprise-provider).
> 
> Aha yes sorry I missed that those were separate docs. The ones you linked should work.
> 
> > I'm having trouble understanding the logic and where to start. I assume that it makes sense to connect App Check on the client side(our React frontend). I'm keep looking...
> 
> It's possible I'm missing something since I have not used this part of Firebase before, but I think we should be adding App Check in the backend, since that's where we are making calls to Firebase from in our case. If we were making the Firebase calls on the client-side, _then_ we would add the App Check credentials there.
> 
> From the docs, you've already completed the first part of [step 1](https://firebase.google.com/docs/app-check/web/recaptcha-enterprise-provider#project-setup) (`1. Set up your Firebase project`) since I know we have a Firebase project ✅
> 
> Then it asks us to create a reCAPTCHA Enterprise key which I believe you're already completed, with a type of "Website" (rather than ios or android app), here: https://console.cloud.google.com/security/recaptcha?project=_.
> 
> My understanding is that for our use-case we can skip the parts shown in the reCAPTCHA console about adding a script to the frontend.
> 
> You've already completed [Step 2](https://firebase.google.com/docs/app-check/web/recaptcha-enterprise-provider) since you have a `firebase.js` file where you do things like `initializeApp` ✅
> 
> Then in [Step 3](https://firebase.google.com/docs/app-check/web/recaptcha-enterprise-provider#initialize) is where we need to add `firebase/app-check` code to our `firebase.js` per the example shown in that section.
> 
> That should be all that's needed to get it set up and activated for now.

@akosasante Sorry, I'm spending more time on this; the topic is new to me but very interesting. At the moment, I understand the situation like this: 

Firebase App Check adds security by ensuring that requests to the backend come from our app. Tokens and Firebase Admin SDK are used for this: https://firebase.google.com/docs/admin/setup?authuser=0&hl=en , right?

Firebase App Check can work and be configured without using reCAPTCHA v3 or reCAPTCHA Enterprise, especially if the current focus is on backend security only for Firebase Storage. It's not necessary to connect reCAPTCHA v3 or reCAPTCHA Enterprise right now. We can add them later on the backend and use them on the frontend where needed for protection against bots, like selecting images with bicycles or answering "I'm not a robot," etc.  
app check


The documentation for integrating reCAPTCHA Enterprise led me through many steps, creating additional keys—Session token, action token, score-based, and connecting to Google Cloud. I didn't understand where to stop to make the code work. I don't know the correct course of action in this situation and what exactly will be the solution for our case.

I can also connect you to my Firebase, or if we decide to stick with Firebase instead of MongoDB and Imgur, we can make another akk for all of us, not my personal, and we can integrate reCAPTCHA Enterprise into our project together with the team. For example, during registration, login, album purchase, etc., if it's part of the free features and doesn't require connecting to Google Cloud, as otherwise, we'll have a 3-month trial version and then it's paid. I might be a bit lost, right?

akosasante:
> > It's possible I'm missing something since I have not used this part of Firebase before, but I think we should be adding App Check in the backend, since that's where we are making calls to Firebase from in our case. If we were making the Firebase calls on the client-side, _then_ we would add the App Check credentials there.  
> > From the docs, you've already completed the first part of [step 1](https://firebase.google.com/docs/app-check/web/recaptcha-enterprise-provider#project-setup) (`1. Set up your Firebase project`) since I know we have a Firebase project ✅  
> > Then it asks us to create a reCAPTCHA Enterprise key which I believe you're already completed, with a type of "Website" (rather than ios or android app), here: [console.cloud.google.com/security/recaptcha?project=](https://console.cloud.google.com/security/recaptcha?project=)\_.  
> > My understanding is that for our use-case we can skip the parts shown in the reCAPTCHA console about adding a script to the frontend.  
> > You've already completed [Step 2](https://firebase.google.com/docs/app-check/web/recaptcha-enterprise-provider) since you have a `firebase.js` file where you do things like `initializeApp` ✅  
> > Then in [Step 3](https://firebase.google.com/docs/app-check/web/recaptcha-enterprise-provider#initialize) is where we need to add `firebase/app-check` code to our `firebase.js` per the example shown in that section.  
> > That should be all that's needed to get it set up and activated for now.
> 
> @akosasante Sorry, I'm spending more time on this; the topic is new to me but very interesting. At the moment, I understand the situation like this:
> 
> Firebase App Check adds security by ensuring that requests to the backend come from our app. Tokens and Firebase Admin SDK are used for this: [firebase.google.com/docs/admin/setup?authuser=0&hl=en](https://firebase.google.com/docs/admin/setup?authuser=0&hl=en) , right?
> 
> Firebase App Check can work and be configured without using reCAPTCHA v3 or reCAPTCHA Enterprise, especially if the current focus is on backend security only for Firebase Storage. It's not necessary to connect reCAPTCHA v3 or reCAPTCHA Enterprise right now. We can add them later on the backend and use them on the frontend where needed for protection against bots, like selecting images with bicycles or answering "I'm not a robot," etc. 
> 
> app check
> 
> The documentation for integrating reCAPTCHA Enterprise led me through many steps, creating additional keys—Session token, action token, score-based, and connecting to Google Cloud. I didn't understand where to stop to make the code work. I don't know the correct course of action in this situation and what exactly will be the solution for our case.
> 
> I can also connect you to my Firebase, or if we decide to stick with Firebase instead of MongoDB and Imgur, we can make another akk for all of us, not my personal, and we can integrate reCAPTCHA Enterprise into our project together with the team. For example, during registration, login, album purchase, etc., if it's part of the free features and doesn't require connecting to Google Cloud, as otherwise, we'll have a 3-month trial version and then it's paid. I might be a bit lost, right?

Hey @mvstoyan. My sincere apologies for the long wait time between reviews here :sweat: .

  
So I've taken some time to dig in deeper to Firebase and I'm pretty sure I was on the wrong track before. My initial understanding from the documentation was that `reCAPTCHA` could be used in both the frontend _or_ the backend to secure our requests to the Firebase/Google servers. My goal was to allow our server to upload files to the Firebase Storage, but ensure that we're securing it so that if some bad actor out there finds out the URL, it should not allow them to upload images to our Storage bucket.

However, I have come to understand that `AppCheck` and `reCAPTCHA` aren't the appropriate approach here after all. You are correct that `reCAPTCHA` generally protects against bots or bad actors by making them perform a check like the one you mentioned about selecting images with bicycles. As well, both `reCAPTCHA v3` and `reCAPTCHA Enterprise` have the option to do a [score-based check](https://cloud.google.com/recaptcha-enterprise/docs/instrument-web-pages); with that one, the user is _not_ presented with any sort of usual CAPTCHA challenge. Instead, the script runs in the background tracking the user's way of interacting with the website, what IP the request is coming from, and additional information in order to assign the request a score. If the score is high enough, it determines that the user is a legitimate human and sends a token in the request to our backend, that our backend would then forward to Firebase, which would allow the request to occur successfully. 

---

That was just some background to verify your understanding of `reCAPTCHA`. Now below is what I have tested and verified as the way we would need to set up the firebase code to work in a secure manner:

1.  Disable AppCheck and delete the `reCAPTCHA` keys you've created, we will not need them for this approach (apologies for the wasted work; hopefully was still some interesting learnings :sweat\_smile: )
2.  In the Firebase Storage rules tab, set the rules to the following to disallow reads/writes from regular users:

```
rules_version = '2';

// For all buckets, and all paths in those buckets, we are setting the rules to NOT allow reads or writes
// This means the only way remaining that can access this storage bucket is a logged in admin or 
// an admin Google Service Account
service firebase.storage {
  match /b/{bucket}/o {
    match /{allPaths=**} {
      allow read, write: if false
    }
  }
}
```



3\. In the Google Console, go to the Service Account page (under `IAM and admin` section: [https://console.cloud.google.com/iam-admin/serviceaccounts?)](https://console.cloud.google.com/iam-admin/serviceaccounts?)) 

4\. Click "Create Service Account". What we're doing here is creating a 'service account' which Google Cloud (which runs Firebase) uses as basically a "machine user" who will be given permissions to access some aspects of our Firebase project

5\. Give the service account a clear name and description



6\. Give the service account access to the project with the `"Cloud Storage for Firebase Service Agent"` role. This is a pre-created role from Google Cloud that tells them "this service account is allowed to make requests to the Firebase Storage API. But is not allowed to make any changes to the application settings or access any other Firebase resources". These are permissions we can always change later if we end up needing access to more Firebase resources from our backend.



We can skip setting any user permissions for this service account.

7\. After creating the service account, click on it and go to the "Keys" tab. Create a new Key for this service account. This should download a JSON file to your computer.

8\. In our `.env` file for the backend, we can remove the other `FIREBASE_*` env values you added before (**except we should still keep the `FIREBASE_STORAGE_BUCKET` value)**. And now we just need one, but we need to make sure we format it correctly since it is a JSON object rather than a plain string. Copy the contents of the JSON file that we downloaded in step 7, and paste it into your `.env` file, wrapped by backticks (\` \`):

```
FIREBASE_SERVICE_ACCOUNT_KEY=`{
                                "type": "service_account",
                                "project_id": "practicum-test-7dc21",
                                "private_key_id": "lsfkjsdlkfjklsdjflksdjfkljsdlkfj",
                                "private_key": "this will be a super long string that I am hiding here to not show my real private key",
                                "client_email": "firebase-adminsdk-7kk19@practicum-test-7dc21.iam.gserviceaccount.com",
                                "client_id": "108930927378831812321312312",
                                "auth_uri": "https://accounts.google.com/o/oauth2/auth",
                                "token_uri": "https://oauth2.googleapis.com/token",
                                "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
                                "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/firebase-adminsdk-7kk19%40practicum-test-7dc21.iam.gserviceaccount.com",
                                "universe_domain": "googleapis.com"
                              }`
```

9\. Remove the `@firebase/storage` and `firebase` libraries (`npm rm -S @firebase/storage firebase`)

10\. Install the FIrebase admin SDK (`npm i -S firebase-admin`)

11\. In `src/firebase.js`: we can change the code as follows to use the admin SDK:

```javascript
const admin = require('firebase-admin');
const serviceAccount = JSON.parse(process.env.FIREBASE_SERVICE_ACCOUNT_KEY);

const firebaseConfig = {
  storageBucket: process.env.FIREBASE_STORAGE_BUCKET,
  credential: admin.credential.cert(serviceAccount),
};

const firebaseApp = admin.initializeApp(firebaseConfig);

const firebaseBucket = firebaseApp.storage().bucket();

console.log('Firebase connected successfully!');

module.exports = { firebaseBucket };
```

12\. In `src/controllers/userController.js` we can change the code as follows to use the admin SDK:

```javascript
// We can remove the imports from `firebase/storage` and instead now we only 
// need to use the bucket from the admin SDK
const { firebaseBucket } = require('../firebase');

//

const updateProfileImage = async (req, res) => {
	console.log('Updating profile image.');
	const userId = req.user.userId;
	if (!req.files || Object.keys(req.files).length === 0) {
		throw new CustomError.BadRequestError('No files were uploaded.');
	}
	const profile = req.files.profile;
	if (!profile || !profile.data || !profile.mimetype) {
		throw new CustomError.BadRequestError('File is not an image.');
	}

	const file = firebaseBucket.file(`images/${userId}_profile_pic`);
	await file.save(profile.data, { contentType: profile.mimetype });

	const imageUrl = file.publicUrl();
	await User.findByIdAndUpdate(
		userId,
		{ profileImage: { url: imageUrl, altText: 'user profile picture' } },
		{ new: true },
	);
	console.log('Profile image updated successfully!');
	res.status(StatusCodes.OK).json({ imageUrl });
};
```

Let me know if you run into any issues!

akosasante:
Some resources I used:

[https://firebase.google.com/docs/admin/setup](https://firebase.google.com/docs/admin/setup)

https://firebase.blog/posts/2019/03/firebase-security-rules-admin-sdk-tips/

[https://medium.com/litslink/firebase-admin-sdk-basics-in-examples-ee7e009a1116](https://medium.com/litslink/firebase-admin-sdk-basics-in-examples-ee7e009a1116)

[https://firebase.google.com/docs/reference/admin/node/firebase-admin.storage](https://firebase.google.com/docs/reference/admin/node/firebase-admin.storage)

https://www.sentinelstand.com/article/guide-to-firebase-storage-download-urls-tokens

mvstoyan:
> Some resources I used:
> 
> https://firebase.google.com/docs/admin/setup
> 
> https://firebase.blog/posts/2019/03/firebase-security-rules-admin-sdk-tips/
> 
> https://medium.com/litslink/firebase-admin-sdk-basics-in-examples-ee7e009a1116
> 
> https://firebase.google.com/docs/reference/admin/node/firebase-admin.storage
> 
> https://www.sentinelstand.com/article/guide-to-firebase-storage-download-urls-tokens

Thank you very much! I'm happy that I could explore all three paths: Firebase Admin SDK, reCAPTCHA, and App Check. I followed the advice, but I'm still learning about Firebase Admin SDK. Right now, the photos are uploaded, but they aren't showing up due to a 403 error. I'm still investigating.

akosasante (approved):
lgtm!

akosasante:
@mvstoyan  -- we'll need to add the environment variables needed by the tests (eg: `FIREBASE_SERVICE_ACCOUNT_KEY`) to the Github Actions secrets -- so that these tests pass in Github Actions/CI: https://github.com/Code-the-Dream-School/dd-prac-team4-back/actions/runs/8888914602/job/24406451245?pr=87

mvstoyan:
@akosasante I don't understand why the CI aren't passing. I've checked the keys, and they match. I copied them directly from the JSON file. I've fixed all the tests, and they're all passing now. But the CI still isn't passing.

akosasante:
> @akosasante I don't understand why the CI aren't passing. I've checked the keys, and they match. I copied them directly from the JSON file. I've fixed all the tests, and they're all passing now. But the CI still isn't passing.

@mvstoyan  Ah, I think I forgot/didn't realize that y'all did not have permission to add secrets to CI. Keep in mind that the `.env` file on your local computer is _local only_. We do not commit it to Github to avoid having secret information available publicly to anyone who browses the repo (or any bad actors who have bots scraping all public Github repos). For production, we have to add the secrets/environment variables directly in Render.io. The same idea applies here for Github actions. We have to add the values in the repo settings; and then set the environment variables in our Github Actions yaml file.

I have made all you ladies admins of this repo now to allow you to practice doing this sort of repo configuration.

Here is a gif showing how to add a secret to this Github repo. There is also a tab called "Variables" which is for values that we want to use that aren't actually secret and we don't mind them being visible in the github actions output.

adding a secret


Then we need to add the environment variable to the Github Actions yaml file here in the step where we run the test command: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/.github/workflows/main.yml#L37-L40](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/.github/workflows/main.yml#L37-L40). For variables we can refer to them as `${{ vars.NAME_OF_VARIBALE_IN_GITHUB_CONFIG }}` and for anything in the secrets tab we refer to it as `${{ secrets.NAME_OF_SECRET_IN_GITHUB_CONFIG }}`

Please add all the environment variables that are needed in the tests to this yaml file and add them as secrets or variables to the repo configuration. Let me know if you run into any trouble.

PR #88, opened 2023-11-22
Page update password on frontend PS-163

DYA13:


src/mailing/sender.js
```diff
@@ -76,11 +76,18 @@ async function sendWelcomeEmail(recipient, user) {
 
 async function sendForgotPasswordEmail(toEmail, resetToken) {
   try {
+    const frontendBaseUrl =
+      process.env.NODE_ENV === 'production'
+        ? 'https://beatbazaar.onrender.com'
+        : 'http://localhost:3000';
```
akosasante: It would be best to put this into an environment variable, so that if we have future cases where we need this value we don't need to rewrite this ternary in other places.

akosasante (approved):
One comment but otherwise looks good

akosasante:
While we're in here, can you fix this line please: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/controllers/authController.js#L200](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/controllers/authController.js#L200)

This is causing an error when that line is called because the correct variable exported by the status codes library is: `StatusCodes.UNAUTHORIZED`

src/mailing/sender.js
DYA13: thank you , done , commit 54da004c79142c7939b127249727ec46cb324794

src/mailing/sender.js
```diff
@@ -76,11 +76,18 @@ async function sendWelcomeEmail(recipient, user) {
 
 async function sendForgotPasswordEmail(toEmail, resetToken) {
   try {
+    const frontendBaseUrl =
+      process.env.NODE_ENV === 'production'
+        ? process.env.FRONTEND_BASE_URL_PROD
+        : process.env.FRONTEND_BASE_URL_DEV;
```
akosasante: We can even just make this:

```suggestion
    const frontendBaseUrl = process.env.FRONTEND_BASE_URL
```

In dev mode, we have a `.env` file where we set this value. We would usually have this value set to the localhost one. But if for some reason a dev really wanted to test against the prod url they would just change the value in the .env file, without havign to fully change the `NODE_ENV` to production.

In production, we set this value in [https://render.com/](https://render.com/) dashboard (make sure to do that before merging). There's no need to have separate env values for prod and dev, because the places where we set these values are already so separate.

akosasante (approved):
One note to make the env value checking simpler but otherwise LGTM

akosasante (approved):
LGTM!

PR #90, opened 2023-11-29
When a new entry is added to the user's listening history, we should generate new recommendations PS-213

DYA13:


DYA13:
Akos I have done initial code , could you please guide me further what to do? 
I would like to ask as wel SPOTIFY_CLIENT_SECRET should be here instead of SPOTIFY_ACCESS_TOKEN , right? 
recommendationUtility.js
    const response = await axios.get(spotifyApiEndpoint, {
      params: {
        seed_artists: artistIds.join(','),
      },
      headers: {
        Authorization: 'SPOTIFY_ACCESS_TOKEN',
      },
    });
as for  spotifyApiEndpoint, should I generate it on site?
Probably files would need refactoring

akosasante (commented):
Just a quick note on ordering of tickets. I see in Jira that I marked `PS-213` as being blocked by `PS-212`. So that means we need work from PS-212 first. And it looks like you've assigned yourself that ticket ("When a user starts playing an album, add it to the listening history")

But `PS-212` is marked as being blocked by `PS-182` ("Backend should store play history"), which is assigned to Natalia. Just a reminder to make sure y'all have talked and are not re-doing same work by implementing the schema for `ListeningHistory.js`. 

If the idea is that in your PR here you're just writing up a barebones version of that schema, to be filled out further by Natalia's work then that's totally fine :+1: . I just wanted to call it out to make sure everyone is on the same page :)

src/utils/recommendationUtility.js
```diff
@@ -0,0 +1,40 @@
+// recommendationUtility.js
+const axios = require('axios'); // for making HTTP requests
+const AlbumRecommendation = require('../models/AlbumRecommendation');
+const ListeningHistory = require('../models/ListeningHistory');
```
akosasante: This will run into some issues because the version of Node.js that we're using doesn't handle circular dependencies very well.

This is a circular dependency because we have `ListeningHistory.js`, which is saying requires `recommendationUtility.js`. But then `recommendationUtility.js` is saying that it needs `ListeningHistory.js`. So one of those won't end up being loaded and will just be null.

You can verify this by trying to load these modules manually in the node REPPL console:

```
# Type the following into your terminal while in the root of this repo
# this will load the node REPL with our `.env` variables and include extra logging for any warnings Node wants to give us
node -r dotenv/config --trace-warnings

# the following is typed inside the node REPL
# you don't need to type the ">" character itself
# lines starting with "//" are comments to explain what I'm doing 

> // connect to the db:
> const connectDb = require('./src/db/connect')
> const db = await connectDb(process.env.MONGO_URL)

> // load the listeningHistory module:
> let ListeningHistory = require('./src/models/ListeningHistory')

> // try to save a ListeningHistory document:
> let x = await ListeningHistory.create({userId: "abc123"})

# This returned the following error and warnings from node
# You can see that the warnings mention "circular dependency"

Error generating recommendations: TypeError: ListeningHistory.find is not a function
    at generateRecommendations (/Users/aasante/dev/practicum2023/backend/src/utils/recommendationUtility.js:9:53)
    at model.<anonymous> (/Users/aasante/dev/practicum2023/backend/src/models/ListeningHistory.js:13:11)
    at callMiddlewareFunction (/Users/aasante/dev/practicum2023/backend/node_modules/kareem/index.js:628:27)
    at next (/Users/aasante/dev/practicum2023/backend/node_modules/kareem/index.js:263:9)
    at Kareem.execPost (/Users/aasante/dev/practicum2023/backend/node_modules/kareem/index.js:296:3)
    at _cb (/Users/aasante/dev/practicum2023/backend/node_modules/kareem/index.js:420:15)
    at /Users/aasante/dev/practicum2023/backend/node_modules/mongoose/lib/model.js:447:5
    at /Users/aasante/dev/practicum2023/backend/node_modules/mongoose/lib/model.js:312:14
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
(node:56436) Warning: Accessing non-existent property 'find' of module exports inside circular dependency
    at emitCircularRequireWarning (node:internal/modules/cjs/loader:707:11)
    at Object.get (node:internal/modules/cjs/loader:721:5)
    at generateRecommendations (/Users/aasante/dev/practicum2023/backend/src/utils/recommendationUtility.js:9:53)
    at model.<anonymous> (/Users/aasante/dev/practicum2023/backend/src/models/ListeningHistory.js:13:11)
    at callMiddlewareFunction (/Users/aasante/dev/practicum2023/backend/node_modules/kareem/index.js:628:27)
    at next (/Users/aasante/dev/practicum2023/backend/node_modules/kareem/index.js:263:9)
    at Kareem.execPost (/Users/aasante/dev/practicum2023/backend/node_modules/kareem/index.js:296:3)
    at _cb (/Users/aasante/dev/practicum2023/backend/node_modules/kareem/index.js:420:15)
    at /Users/aasante/dev/practicum2023/backend/node_modules/mongoose/lib/model.js:447:5
    at /Users/aasante/dev/practicum2023/backend/node_modules/mongoose/lib/model.js:312:14

```

akosasante: One way we can avoid this here is doing the fetching of `ListeningHistory` documents inside the post-save middleware in `ListeningHistory.js` and then just pass that list of documents to the `generateRecommendations` function instead of asking it to do the fetching itself.

akosasante: Inside the post-save middleware; we can access the model (to do `ListeningHistory.find` ) with `this.constructor` (see the `Review.js` file for an example).

src/utils/recommendationUtility.js
```diff
@@ -0,0 +1,40 @@
...
+
+async function generateRecommendations(userId) {
+  try {
+    // Fetch the top 5 listened to artists from the user's listening history
+    const listeningHistory = await ListeningHistory.find({ userId })
+      .sort({ playCount: -1 })
```
akosasante: Based on what's outlined in `PS-182` we will be entering a separate row/document in the `ListeningHistory` schema for each time the user listens to an album: [https://codethedreamteam4-2023.atlassian.net/browse/PS-182](https://codethedreamteam4-2023.atlassian.net/browse/PS-182)

So there may not be an overall `playCount` attribute in the documents (unless we feel like that will be a better approach to go forward with). Instead, we'll have documents like:

```json
[
  {
    "user": "abc",
    "artist": "Shakira",
    "album": "Laundry Service",
    "timeListened": "2023-01-12"
  },
  {
    "user": "abc",
    "artist": "Drake",
    "album": "To all the Dogs",
    "timeListened": "2023-01-12"
  },
  {
    "user": "xyz",
    "artist": "Drake",
    "album": "To all the Dogs",
    "timeListened": "2023-01-11"
  },
  {
    "user": "abc",
    "artist": "Shakira",
    "album": "Laundry Service",
    "timeListened": "2023-01-18"
  },
  {
    "user": "abc",
    "artist": "Shakira",
    "album": "She Wolf",
    "timeListened": "2023-01-18"
  }
]
```

We want to take that list of documents and (1) filter them to just get this particular user's listens (which you've done with \`.find({ userId})\` :white\_check\_mark: , and (2) use Mongoose `aggregate` function to group them so that we can get a count of times they've listened to each artist. For example in the above example, user `abc` has listened to Shakira 3 times (even though there's two different albums they've listened to, and one album they listened to twice). user `abc` also listened to Drake 1 times. And user `xyz` listened to Drake 1 time.

So user `abc`'s top artist would be Shakira; or if we were returning up to 5 artists, then we'd return both Shakira and Drake. 

https://dev.to/petrussola/mongoose-and-how-to-group-by-count-5gcm
https://mongoosejs.com/docs/api/aggregate.html#Aggregate.prototype.sortByCount()

src/utils/recommendationUtility.js
```diff
@@ -0,0 +1,40 @@
...
+    // Make a call to Spotify's "Get Recommendations" endpoint
+    const spotifyApiEndpoint = 'https://api.spotify.com/v1/recommendations';
+    const response = await axios.get(spotifyApiEndpoint, {
+      params: {
+        seed_artists: artistIds.join(','),
+      },
+      headers: {
+        Authorization: 'SPOTIFY_ACCESS_TOKEN',
+      },
+    });
```
akosasante: We have in this backend repo a library installed that allows us to access the Spotify Web API endpoints in a more structured way than needing to make our own HTTP/axios calls: [https://github.com/thelinmichael/spotify-web-api-node](https://github.com/thelinmichael/spotify-web-api-node)

You can see it in use in the admin script `fetchAlbumSpotify.js`: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/321237cb8427951d64b61e7c73e1a6062f131ff3/fetchAlbumSpotify.js#L2](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/321237cb8427951d64b61e7c73e1a6062f131ff3/fetchAlbumSpotify.js#L2)

To use that one, we can continue use the `SPOTIFY_CLIENT_ID` and `SPOTIFY_CLIENT_SECRET` that we already have in the `.env` file.  Then we make a call to the Spotify OAuth endpoint to get an access token for the session (see the lines in that `fetchAlbumSpotify.js` file where we call `clientCredentialsGrant` and then `setAccessToken`)

src/utils/recommendationUtility.js
DYA13: Hello Akos, I have implemented some changes in 
commit f84662a8097de12c4e8987582553b5feda3ae22f 
Not sure if it works

src/utils/recommendationUtility.js
DYA13: I have implemented the proposed changes , thank you 👍

DYA13: Could you please check if it works with my changes ? 🫣

src/models/ListeningHistory.js
akosasante, line 1: A little bit of bad planning on my part sorry! It looks like @NatachaKey 's ticket PS-182 also led to her creating a schema for storing recently listened to albums: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/91/files](https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/91/files)

Let's merge those changes in (`git merge origin/backendStoresPlayHistory`) and then include the `post-save` callback you've done here but instead in the `RecentlyListened` schema

src/models/ListeningHistory.js
```diff
@@ -0,0 +1,34 @@
...
+const listeningHistorySchema = new Schema({
+  userId: String,
+});
+
+listeningHistorySchema.post('save', async function (doc, next) {
+  try {
+    // Access the model to fetch ListeningHistory documents
+    const ListeningHistory = this.constructor;
+
+    // Fetch the top 5 listened to artists from the user's listening history
```
akosasante, line 14:
```suggestion
    // Fetch the 5 most recent listened to albums from the user's listening history
```

src/utils/recommendationUtility.js
```diff
@@ -0,0 +1,40 @@
...
+    // Check if listeningHistory is empty
+    if (listeningHistory.length === 0) {
+      console.warn('No listening history found for user:', userId);
+      return;
+    }
+
+    // Extract artistIds from listeningHistory
+    const artistIds = listeningHistory.map((entry) => entry.artistId);
+
+    // Make a call to Spotify's "Get Recommendations" endpoint
```
akosasante, line 21: Before we can make any calls to Spotify's api we need to authenticate our session. We can follow the same pattern we've done over in the `fetchAlbumSpotify` file:

https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/321237cb8427951d64b61e7c73e1a6062f131ff3/fetchAlbumSpotify.js#L21-L23

src/utils/recommendationUtility.js
```diff
@@ -0,0 +1,40 @@
...
+async function generateRecommendations(userId, listeningHistory) {
+  try {
+    // Check if listeningHistory is empty
+    if (listeningHistory.length === 0) {
+      console.warn('No listening history found for user:', userId);
+      return;
+    }
+
+    // Extract artistIds from listeningHistory
+    const artistIds = listeningHistory.map((entry) => entry.artistId);
```
akosasante, line 19: Currently, we aren't storing the spotify artist id in our application. So we will need to first make a call to Spotify to get the artist id. 

There is a `Get Album` by id endpoint: [https://developer.spotify.com/documentation/web-api/reference/get-an-album](https://developer.spotify.com/documentation/web-api/reference/get-an-album)

Our spotify-web-api library that we're using can access it with the `getAlbum` function: [https://github.com/thelinmichael/spotify-web-api-node#more-examples](https://github.com/thelinmichael/spotify-web-api-node#more-examples)

The album id we need is the _spotify album id_ and not our own Mongo object id. Right now, the only place we're storing that is in the `spotifyUrl`, we may want to change that later but for now, we can follow a similar pattern as the frontend where we use regex to extract that id from the url field: [https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/40ffb979b0229c5f2485eaecc87e31af55fd3fa4/src/components/album/AlbumPreview.js#L19-L29](https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/40ffb979b0229c5f2485eaecc87e31af55fd3fa4/src/components/album/AlbumPreview.js#L19-L29)

An album can have multiple artists, for now just grab the first artist id in the returned list.

src/utils/recommendationUtility.js
```diff
@@ -0,0 +1,40 @@
...
+      seed_artists: artistIds.join(','),
+    });
+
+    // Store the returned recommended tracks/songs in the AlbumRecommendation schema
+    const recommendations = response.body.tracks.map((track) => ({
+      userId,
+      songId: track.id,
+    }));
+
+    await AlbumRecommendation.insertMany(recommendations);
```
akosasante, line 32: We will probably want to make `AlbumRecommendations` have a `uniqueIndex` so that we don't insert the same recommended song multiple times. And in that case we'll want to add `ordered: false` so that it will insert all unique entries and return errors for any that were not unique: https://mongoosejs.com/docs/api/model.html#Model.insertMany()

